PR #20, opened 2021-06-10
punctuation bug fix

sharkbait2:
without punctuations it's harder to read, so I fixed it!


fridis:
I agree on the importance of punctuation for readability, but it is also somewhat important that the compiler accepts the code.

sharkbait2:
Ooh yeah, It does more complaining than compiling.

PR #41, opened 2021-08-11
Add -acemode switch for generating a mode file for the ace editor.

miclill:


src/dev/flang/tools/AceMode.java
```diff
@@ -0,0 +1,250 @@
...
+ *
+ * Source of class AceMode
+ *
+ *---------------------------------------------------------------------*/
+
+package dev.flang.tools;
+
+import dev.flang.parser.Lexer;
+import dev.flang.util.ANY;
+import java.util.stream.*;
```
fridis, line 31: please do not use * in imports, in particular if all you need is
```suggestion
import java.util.stream.Stream;
```

fridis, line 31: I keep imports sorted alphabetically and packages separated by blank lines, i.e., 
```suggestion
import dev.flang.parser.Lexer;

import dev.flang.util.ANY;

import java.util.stream.Stream;
```
even though in this case this does not make a bit difference since there are only single classes imported from each package.

fridis (changes requested):

src/dev/flang/tools/AceMode.java
michaellilltokiwa: That's definitely better. I replaced the line in question by:
import java.util.stream.Collectors;
import java.util.stream.Stream;

michaellilltokiwa: Done in #88d9dff

PR #55, opened 2021-08-19
Add precedence definition for the fraction operator

michaellilltokiwa:


michaellilltokiwa:
@fridis I'm not confident of the test. Feedback is welcome.

PR #73, opened 2021-09-16
Fix/windows tests

michaellilltokiwa:
This PR should fix the tests on windows.

left linux, right windows:

image

### DIFF
--- a/tests/_cur_dir.sh
+++ b/tests/_cur_dir.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+# @returns:
+#   - unix: /my/current/path
+#   - windows: C:\\\my\\\current\\\path
+
+CURDIR="$PWD"
+if [ -x "$(command -v cygpath)" ]
+then
+    CURDIR=$(cygpath -w $PWD | sed 's/\\/\\\\/g')
+fi
+
+echo $CURDIR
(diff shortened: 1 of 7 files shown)


tests/_cur_dir.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
fridis, line 1: please add the Fuzion header like in other scripts

fridis (changes requested):
looks good.

tests/_cur_dir.sh
michaellilltokiwa: @fridis done

tests/_cur_dir.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of fz command, the main Fuzion tools entry point
```
fridis, line 20: this needs an update

tests/_cur_dir.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+#  Author: Fridtjof Siebert (siebert@tokiwa.software)
```
fridis, line 22: and this should be you.

fridis (changes requested):

PR #76, opened 2021-09-17
fix some spelling mistakes

michaellilltokiwa:


src/dev/flang/ast/ValueType.java
```diff
@@ -60,7 +60,7 @@ private ValueType()
 
 
   /**
-   * true iff this is the return type of a constructor feature, i.e., "this" is
```
fridis, line 63: This is not a typo, see [https://en.wikipedia.org/wiki/If_and_only_if](if and only if)

fridis (commented):
good, just one little problem found.

PR #163, opened 2022-02-08
shellscripts: aplly suggested fixes of shellcheck

michaellilltokiwa:
shellcheck -f diff [files] | git apply


fridis:
Can you add a .phony target to the Makefile that runs shellcheck to show if there are any new issues? This could then become a task to check during nightly builds, but without adding a dependency on shellcheck for normal builds.

michaellilltokiwa:
@fridis yes, good idea!

PR #189, opened 2022-02-18
stdllib: string, implement feature isBlank

michaellilltokiwa:
- returns true if string is empty or contains whitespace only


lib/string.fz
```diff
@@ -38,6 +38,9 @@ string ref : hasHash<string>, ordered<string>, strings is
   # is this string empty?
   isEmpty => !utf8.asStream.hasNext
 
+  # returns true if string is empty or contains whitespace only
+  isBlank => utf8 ∀ (u) -> isAsciiWhiteSpace u
```
fridis, line 42: nice. Only one minor suggestion: the parentheses around '(u)' are not needed and should be removed.

PR #210, opened 2022-03-08
Lib/fuzion.std.nano time

michaellilltokiwa:


lib/fuzion/std.fz
```diff
@@ -35,3 +35,6 @@ fuzion.std is
   panic(msg string) void is
     say "*** panic: $msg"
     exit 1
+
+  # no guaranty can be given for precision nor resolution
```
fridis, line 39: guarantee

lib/bench.fz
```diff
@@ -0,0 +1,52 @@
...
+  warm_up
+
+  iterations := for iter := u64.zero, iter + u64s.one
+  while fuzion.std.nano_time - start < nano_seconds
+  do
+    f.call
+  else
+    iter
+
+  (iterations.as_f64 / (nano_seconds.as_f64 * 1E-9))
```
fridis, line 46: Remove `iterations := ` at the beginning and put `iter.as_f64 / (nano_seconds.as_f64 * 1E-9)` in the `else` part.

lib/bench.fz
```diff
@@ -0,0 +1,52 @@
...
+  start := fuzion.std.nano_time
+
+  warm_up =>
+    do
+      f.call
+    until fuzion.std.nano_time - start > warm_up_nano_seconds
+
+  warm_up
+
+  iterations := for iter := u64.zero, iter + u64s.one
```
fridis, line 39: why don't you just write
```
  iter := u64 0, iter + 1
```
The `zero`/`one` constants are basically for generic code in `integer` that should also work for user-defined numbers-like things like a matrix.

lib/bench.fz
```diff
@@ -0,0 +1,52 @@
...
+# benchmark f for nano_seconds
+# returns: iterations per second
+#
+bench<R>(f Function<R>, nano_seconds u64, warm_up_nano_seconds u64) f64 is
+  start := fuzion.std.nano_time
+
+  warm_up =>
+    do
+      f.call
+    until fuzion.std.nano_time - start > warm_up_nano_seconds
```
fridis, line 35: It would be better to put the warm-up and the iteration-loop into one single loop and just not increment `iter` during warm-up, such that filling the instruction caches and branch target prediction is part of the warm-up.

lib/bench.fz
```diff
@@ -0,0 +1,52 @@
...
+#  Source code of Fuzion standard library feature bench
+#
+#  Author: Michael Lill (michael.lill@tokiwa.software)
+#
+# -----------------------------------------------------------------------
+
+# benchmark f for nano_seconds
+# returns: iterations per second
+#
+bench<R>(f Function<R>, nano_seconds u64, warm_up_nano_seconds u64) f64 is
```
fridis, line 29: Fine for now, but eventually the `universe` is getting overloaded with stuff, we will have to define some structure at some point.

PR #211, opened 2022-03-10
Various minor fixes

michaellilltokiwa:


src/dev/flang/be/c/CExpr.java
```diff
@@ -88,7 +88,7 @@ static CExpr dummy(String text)
    *
    * @return the resulting expression
    */
-  static CExpr ident(String name)
+  static CIdent ident(String name)
```
fridis, line 91: I do not agree with this. The idea ist that this creates an Expression, which is also documented. The result is of type CExpr, but might be some subtype of that, e.g., we might add comments or debugging code to every single expression produced.

fridis (commented):
Would be better to have several pull requests and not to mix unrelated things into a larger patch. In particular, semantic changes and comments or restructuring should be separate, different semantic changes should definitely be separate patches.

PR #232, opened 2022-04-05
lib/be: implement atan2 and isNaN

michaellilltokiwa:


tests/floating_point_math/floating_point_math.fz
```diff
@@ -41,6 +41,14 @@ floating_point_math is
     zero := a.zero
     one := a.one
     two := one + one
+    three := two + one
+    four := two + two
```
fridis, line 45: What are these for? 
For integers, I define two and ten since they are needed for conversion to / from string. For floats, this should not be needed.

PR #244, opened 2022-04-13
lib/Sequence: flavour of first which takes a default value

michaellilltokiwa:
example:
ex =>
  say (lists.empty<string>.first "Hello")
  say (["Hello"].first "World")

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -80,6 +80,13 @@ Sequence<T> ref is
     => asList.forceHead
 
 
+  # get the head of this list or default
+  #
+  first(default T)
+    =>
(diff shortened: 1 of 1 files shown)


lib/Sequence.fz
```diff
@@ -80,6 +80,13 @@ Sequence<T> ref is
     => asList.forceHead
 
 
+  # get the head of this list or default
+  #
+  first(default T)
+    =>
+      if isEmpty then default else asList.forceHead
```
fridis, line 87:
```
if isEmpty then default else first
```
would be slightly better. 

Can you also add `last(T)`?

PR #245, opened 2022-04-14
Lib/float sequence providing min/max/avg/std_dev ...

michaellilltokiwa:


lib/numericSequence.fz
```diff
@@ -0,0 +1,83 @@
...
+numericSequence<N: numeric<N>>(from Sequence<N>) : Sequence<N> is
+
+  # create a list from this Sequence.
+  #
+  redef asList => from.asList
+
+
+  # the arithmetic mean of the sequence
+  # https://en.wikipedia.org/wiki/Arithmetic_mean
+  average option<N> is
```
fridis, line 37: I try to separate the comment from the feature by a blank comment line
```
  #
```
even though I am not fully consistent with this. Try to do this as well.

fridis (approved):

PR #246, opened 2022-04-14
lib/Sequence: add nth for getting the nth element in a sequence

michaellilltokiwa:


lib/Sequence.fz
```diff
@@ -285,6 +285,18 @@ Sequence<T> ref is
   zip<U,V>(b Sequence<U>, f (T,U)->V) => asList.zip0 b.asList f
 
 
+  # the nth element in the sequence,
+  # starting at 1
```
fridis, line 289: I would like to be consistent in starting to count elements at `0`. Then this feature would behave similar to `array.index [ ]`.

michaellilltokiwa: @fridis yes, good point

PR #255, opened 2022-04-21
lib/float: add round, ties to away

michaellilltokiwa:


lib/float.fz
```diff
@@ -46,6 +46,25 @@ float(redef F float<F>.type) : numeric<F>, floats<F> is
   redef infix **!(other F) bool is true
 
 
+  as_i64 i64 is abstract
+
+  from_i64(val i64) F is abstract
```
fridis, line 51: some comments would be nice, in a separate patch.

PR #265, opened 2022-04-26
lib/float: add fract[ion], floor, ceil[ing], changed return type of round.

michaellilltokiwa:


lib/f32.fz
```diff
@@ -70,6 +70,27 @@ f32(val f32) : float<f32>, f32s is
     castTo_u32.hash
 
 
+  # the fraction of the floating point number
+  redef fract f32 is
```
fridis, line 74: redef is not needed if the original feature is abstract. but it also does not hurt.

PR #273, opened 2022-05-02
lib/list: remove unnecessary match

michaellilltokiwa:


fridis:
I think I did this intentionally to optimize for the case that t is empty. Without this, doing something like
```
for 
  l := [a], l.concat nil
  i in 1..10000
else
  say l
```
will wrap `[a]` into 10000 lazy Cons-cells.

michaellilltokiwa:
@fridis ah okay. I did not see this. :+1:

PR #290, opened 2022-05-10
lib/write: changed to write utf8 byte arrays

michaellilltokiwa:
With this change flush is unused.  We may consider removing it.


fridis:
this is in conflict with a change I made meanwhile. Can you resolve it?

michaellilltokiwa:
@fridis resolved

PR #311, opened 2022-06-07
util: SourceFile.codePointInLine erroneously returned byte-Index

michaellilltokiwa:
@fridis: This fix may have some performance implications since `codePointInLine()` is called in the Lexer quite often. 
Put differently, I'm open to suggestions to fix this bug in a better way.


fridis:
> @fridis: This fix may have some performance implications since `codePointInLine()` is called in the Lexer quite often. Put differently, I'm open to suggestions to fix this bug in a better way.

I am not too worried about this.

PR #312, opened 2022-06-08
util: fix ArrayIndexOutOfBoundsException for empty source files

michaellilltokiwa:
The PR yesterday introduced a minor bug that hit me when running the tests of the language server. Turns out `line` can take the special value 0. Then lineStartPos() crashes with an out of index exception.

### DIFF
--- a/src/dev/flang/util/SourceFile.java
+++ b/src/dev/flang/util/SourceFile.java
@@ -807,10 +807,13 @@ public class SourceFile extends ANY
    */
   public int codePointInLine(int pos, int line)
   {
+    if(PRECONDITIONS)
+      require(line > 0);
     int c = 1;
-    for (int i = lineStartPos(line); i < pos; i = i + sizeFromCpAndSize(decodeCodePointAndSize(i))){
-      c++;
-    }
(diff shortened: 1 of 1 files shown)


src/dev/flang/util/SourceFile.java
```diff
@@ -807,10 +807,13 @@ public int lineEndPos(int line)
    */
   public int codePointInLine(int pos, int line)
   {
+    if(PRECONDITIONS)
```
fridis, line 810: please add a space between `if` and `(`. 

Also, the form I use for preconditions is
```
if (PRECONDITIONS) require
  (line >= 0);
```
this is to make this look more similar to Eiffel's `requires` or Fuzion's `pre` clause:
```
  pre
     line >= 0
```
please fix in separate PR.

src/dev/flang/util/SourceFile.java
```diff
@@ -821,7 +824,12 @@ public int codePointInLine(int pos, int line)
    */
   public int codePointInLine(int pos)
   {
-    return codePointInLine(pos, lineNum(pos));
+    int line = lineNum(pos);
+    if(line == 0)
```
fridis, line 828: Add space: `if (` in separate PL.

PR #313, opened 2022-06-09
lib/thread: relax return type of code passed to spawn

michaellilltokiwa:
I think this change might make thread spawning more convenient.
@fridis Is there any downside to allowing the code passed to a thread returning not just unit but anything?

### DIFF
--- a/lib/concur/thread.fz
+++ b/lib/concur/thread.fz
@@ -37,7 +37,7 @@ is
 
   # spawn a new thread
   #
-  spawn(code ()->unit) =>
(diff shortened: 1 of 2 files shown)


fridis:
Not sure that this does what you want, making the result `Object` would make in impossible to provide a function that results in `unit` (or `i32`, `string` or anything that is not `Object`). `spawn`should eventually have a type parameter like this:
```
spawn(R type, code ()->R) Future R is ...
```

michaellilltokiwa:
@fridis Even better.

PR #314, opened 2022-06-09
lib/map: simplify code for getting keys/values of map

michaellilltokiwa:


fridis:
good. I want to remove `stream` completely!

PR #345, opened 2022-07-20
pre/post-conditions: show where the failure originates from

michaellilltokiwa:


src/dev/flang/util/ANY.java
```diff
@@ -55,6 +55,16 @@
   /*-----------------------------  methods  -----------------------------*/
 
 
+  private static String Origin(){
```
fridis, line 58: Can you add a comment? Please rename using lower case `origin`.

fridis (commented):
Good idea!

Just the usual nitpicking...:-)

PR #346, opened 2022-07-20
readFile implementation

WaelYoussfi:
readFile intrinsic reads a byte in a specified position from a file and returns it

### DIFF
--- a/lib/fuzion/std/fileio.fz
+++ b/lib/fuzion/std/fileio.fz
@@ -30,24 +30,40 @@ public fuzion.std.fileio is
 
   # read a file into an array u8.
   # 
-  # NYI : an equivalent to a Java MappedByteBuffer would be a better solution
-  #
   public readFile(
                   # the (relative or absolute) file name, using platform specific path separators  
                   path string) =>
-    file := array getFileSize(path).as_i32 i->(u8 0)
+    file := array getFileSize(path).as_i32 i->(u8 0) # allocating an array the size of the file
     arr := path.utf8.asArray # transforms the path to a utf8 byte array
-    readFile arr.internalArray.data arr.length file.internalArray.data
+    readFile arr.internalArray.data arr.length file.internalArray.data file.length
     file
 
-  public readFile(path Object, pathLength i32, fileArray Object) unit is intrinsic
+  # intrinsic that fills an array u8 with the file bytes.
+  # 
+  private readFile(
+                   # the internal array data representing the file path in bytes
+                   path Object,
+                   # the length of the internal array representing the path
+                   pathLength i32,
+                   # the internal array data representing the container for the bytes to be read from the file
+                   fileArray Object,
+                   # the length of the array that represents the file bytes
+                   fileArrayLength i32) i64 is intrinsic
 
   # retrieve the file size in bytes and returns -1 in case of an error
   #
-  public getFileSize(path string) =>
+  public getFileSize(
+                     # the (relative or absolute) file name, using platform specific path separators
+                     path string) =>
     arr := path.utf8.asArray
     getFileSize arr.internalArray.data arr.length
 
-  public getFileSize(path Object, pathLength i32) i64 is intrinsic
+  # intrinsic that returns the file size in bytes or -1 in case of an error
+  #
+  private getFileSize(
+                      # the internal array data representing the file path in bytes
+                      path Object,
+                      # the length of the internal array representing the path
+                      pathLength i32) i64 is intrinsic
 
-  public writeFile(path string, content array<u8>) is intrinsic # NYI
+  private writeFile(path string, content array<u8>) is intrinsic # NYI
--- a/lib/fuzion/std/fileio.fz
+++ b/lib/fuzion/std/fileio.fz
@@ -1,9 +1,53 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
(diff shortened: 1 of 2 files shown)


lib/fuzion/std/fileio.fz
```diff
@@ -0,0 +1,9 @@
+public fuzion.std.fileio is
```
fridis, line 1: Please add our standard header as in `out.fz` consisting of the GPLv3 header,  the Tokiwa banner, and a comment giving in the first line the qualified feature name folllowed by `--` and short description followed an optional longer description in the following paragraphs.

lib/fuzion/std/fileio.fz
```diff
@@ -0,0 +1,9 @@
...
+
+    public readFile(path string, pos i32) =>
```
fridis, line 3: please add a comment before every feature explaining its purpose, if needed, the arguments. Comments for arguments should go before the arguments, i.e., 

```
  # read a file starting at given position into an array u8.
  # 
  # NYI: currently reads a single u8 byte only. 
  #
  public readFile(
                  # the (relative or absulute) file name, using platform specific path separators
                  path string,

                  # the byte position to read from 
                  pos i32) =>
```
In general, please use ì64` (or `u64`?) for file positions and lengths, will avoid trouble in the future! I am thinking about making `i64` the default integer type anyways.

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -178,6 +179,20 @@ public static Callable call(Interpreter interpreter, Clazz innerClazz)
               return Value.EMPTY_VALUE;
             };
         });
+    put("fuzion.std.fileio.readFile", (interpreter, innerClazz)-> args ->
+        {
+          byte[] content;
+          byte[] pathBytes = (byte[])args.get(1).arrayData()._array;
+          int pos = (int)args.get(3).i32Value(); //position of the byte in file
+          Path path = Path.of(new String(pathBytes, StandardCharsets.UTF_8)); //path to the file
+          try {
```
fridis, line 188: Always place `{` in a new line as in 
```
  try
    {
      doSomething(); 
    }
  catch (...)
```

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -178,6 +179,20 @@ public static Callable call(Interpreter interpreter, Clazz innerClazz)
...
+            content = Files.readAllBytes(path);
+            return new u8Value(content[pos]);
+          } catch (IOException e) {
+            System.err.println(e);
+            return new u8Value(-1);
```
fridis, line 193: this looks like trouble. For now, just `throw new Error("some useful message "+e)` and add a comment starting `NYI` explaining that we still need proper error handling here.

fridis (commented):
Good first PR!

My review is a lot of nitpicking trying to keep a coherent style in the sources, no major problems.

fridis:
One more point: Please add a check for flag `ENABLE_UNSAFE_INTRINSICS` to the intrinsics you add. This flag is disabled by default on fuzion running on flang.dev to make it harder to take over the webserver by typing in malicious code.  You can enable these intrinsics via `fz -unsafeIntrinsics=on` or adding `-Dfuzion.enableUnsafeIntrinsics=true` to the `java` command when running `dev.flang.tools.Fuzion`.

lib/fuzion/std/fileio.fz
```diff
@@ -0,0 +1,9 @@
...
+        arr := path.utf8.asArray #transform the string to a utf8 byte array
+        readFile arr.internalArray.data arr.length pos
+
+    public readFile(path Object, length i32, position i32) u8 is intrinsic
+
+    public writeFile(path string, content array<u8>) is intrinsic
```
fridis, line 9: please always and a text file with a LF.

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -178,6 +181,37 @@ public static Callable call(Interpreter interpreter, Clazz innerClazz)
...
+          var byteArr = (byte[])args.get(3).arrayData()._array;
+          try
+            {
+              File file = new File(new String(pathBytes , StandardCharsets.UTF_8));
+              FileInputStream fs = new FileInputStream(file);
+              fs.read(byteArr);
+              fs.close();
+              return Value.EMPTY_VALUE;
+            } 
+            catch (Exception e)
```
fridis, line 196: the catch block should be indented 2 spaces less.

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -178,6 +181,37 @@ public static Callable call(Interpreter interpreter, Clazz innerClazz)
...
+    put("fuzion.std.fileio.getFileSize", (interpreter, innerClazz) -> args ->
+        {
+          byte[] pathBytes = (byte[])args.get(1).arrayData()._array;
+          Path path = Path.of(new String(pathBytes, StandardCharsets.UTF_8));
+          try
+            {
+              long fileLength = Files.size(path);
+              return new i64Value(fileLength);
+            } 
+            catch (Exception e)
```
fridis, line 210: the catch block should be indented 2 spaces less.

lib/fuzion/std/fileio.fz
```diff
@@ -0,0 +1,53 @@
...
+
+# fuzion.std.fileio -- fileio presents two simple methods:
+# readFile to retrieve data from a file
+# WriteFile to insert data into a file 
+#
+public fuzion.std.fileio is
+
+  # read a file into an array u8.
+  # 
+  # NYI : an equivalent to a Java MappedByteBuffer would be a better solution
```
fridis, line 33: this NYI comment is not need, MappedByteBuffer is a different solution that is better for some uses, worse for others.

lib/fuzion/std/fileio.fz
```diff
@@ -0,0 +1,53 @@
...
+  #
+  public readFile(
+                  # the (relative or absolute) file name, using platform specific path separators  
+                  path string) =>
+    file := array getFileSize(path).as_i32 i->(u8 0)
+    arr := path.utf8.asArray # transforms the path to a utf8 byte array
+    readFile arr.internalArray.data arr.length file.internalArray.data
+    file
+
+  public readFile(path Object, pathLength i32, fileArray Object) unit is intrinsic
```
fridis, line 43: please add comment. 
intrinsics should not be public. Eventually, nothing in fuzion.std or fuzion.sys should be exported, only the corresponding effects io.file.read or similar should be.

fridis (commented):
Please add the check for ENABLED_UNSAFE_INTRINSICS. This is important since otherwise, we would permit access to all files on our flang.dev webserver to anyone who has access to the internat.

PR #347, opened 2022-07-20
lib: redefined asString of mutable value to return asString of the cu…

michaellilltokiwa:
…rrent value

### DIFF
--- a/lib/mutate.fz
+++ b/lib/mutate.fz
@@ -104,8 +104,9 @@ is
       mut get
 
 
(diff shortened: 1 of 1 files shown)


lib/mutate.fz
```diff
@@ -104,6 +104,10 @@ is
       mut get
 
 
+    redef asString => get.asString
```
fridis, line 107: nice. 
Just add a small comment and remove one of the three empty lines :-)

PR #351, opened 2022-07-22
writeFile feature implementation

WaelYoussfi:


src/dev/flang/be/c/Intrinsics.java
```diff
@@ -93,6 +93,12 @@ public class Intrinsics extends ANY
                                 outOrErr(in)
                               ));
         });
+    IntrinsicCode noFileIo = (c,cl,outer,in) ->
+      CStmnt.seq(CExpr.fprintfstderr("*** C backend support does not support fileio (yet).\n"),
```
fridis, line 97: English sounds strange. Maybe remove the first `support`.

lib/fuzion/std/fileio.fz
```diff
@@ -66,4 +66,26 @@ public fuzion.std.fileio is
                       # the length of the internal array representing the path
                       pathLength i32) i64 is intrinsic
 
-  private writeFile(path string, content array<u8>) is intrinsic # NYI
+  # creates a new file using an array of u8 as content and a string as the file path
+  # it overrides the file, if it exists, or creates a new one if it does not exist
```
fridis, line 70: overrides -> overwrites?

fridis (commented):
just nitpicking in the English text.

lib/fuzion/std/fileio.fz
WaelYoussfi: typo because i used to use @Override in Java

PR #352, opened 2022-07-25
lib: option, add flat_map feature

michaellilltokiwa:


lib/option.fz
```diff
@@ -90,6 +90,13 @@ is
                 | nil => nil
 
 
+  # flat_map this option using f, i.e., map nil to nil and any value v to f v
+  #
+  flat_map (B type, f T -> option B) option B is
+    option.this ? v T => f v
+                | nil => nil
+
+
```
fridis, line 99: Please implement this using `bind`. We have three names for this then: `bind`, `infix >>=` and `flat_map`. We should eventually settle to only one or two, but I am not sure which one..,

fridis (commented):
Can this maybe be implemented in monad.fz using `infix >>=`? Then we would have `flat_map` as an alias for `bind` for all monads!

lib/option.fz
michaellilltokiwa: @fridis I'm gonna close this and add to my TODO the research of the most appropriate name. To me personally flatmap is more familiar than bind but that is probably just my history of languages java(script),c#,...

PR #357, opened 2022-07-27
changing to snake case notation and starting the io.file implementation

WaelYoussfi:


lib/read_file.fz
```diff
@@ -0,0 +1,29 @@
...
+#  Source code of Fuzion standard library feature read_file
+#
+#
+# -----------------------------------------------------------------------
+
+# read_file -- shortcut for io.file.read
+#
+# A handy shortcut for io.file.read, returns the content of the file in the path as an array of u8.
+#
+read_file(path string) => io.file.read path # NYI
```
fridis, line 29: I do not think this is necessary, `read_file "a"` saves only 3 characers compared to `io.file.read "a"`, and then it should be possible to inherit from `io.file` such that one could write `read "a"`.

lib/io/file.fz
```diff
@@ -0,0 +1,30 @@
...
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature file
+#
```
fridis, line 21: You can put yourself as the author here, I think it is good to have a reference to the original author in a file in addition to the git history.

lib/io/file.fz
```diff
@@ -0,0 +1,30 @@
...
+#
+# -----------------------------------------------------------------------
+
+# file -- effect related to fileio
```
fridis, line 25: `io.file` should probably not be the effect, but just a unit type that contains effects related to files. The actual effects should be features like `io.file.read`, `io.file.write`, `io.file.delete`, `io.file.rename`, `io.file.file_size`, etc.

fridis (commented):
For future PRs, it would be good to not put independent things in the same PR, i.e., renaming features and adding new functionality should be separate PRs. I myself still have to learn what is the bet workflow in git to do this, but it makes reviews much easier if you can pull as much as possible early while nitpicking on some minor things.

PR #361, opened 2022-07-28
lib/stream: add comment on take feature

michaellilltokiwa:


lib/stream.fz
```diff
@@ -53,6 +53,8 @@ stream(redef T type) ref : Sequence<T> is
       nil
 
 
+  # take n items from stream, less if stream has less than n items
```
fridis, line 56: `less` -> `fewer` :-)

PR #366, opened 2022-08-03
implementing intrinsics that create temp directories and files

WaelYoussfi:


fridis (commented):
I might be wrong but I fear this intrinsic might require quite a lot of code in the C backend. Do you have a plan how to implement this in C? 
I prefer the intrinsics be relatively simple in for all targets and implement more in fuzion itself.

PR #368, opened 2022-08-03
implementation of the feature exists to verify the existence of a file/dir

WaelYoussfi:


lib/fuzion/std/fileio.fz
```diff
@@ -83,6 +83,7 @@ public fuzion.std.fileio is
 
   # intrinsic to write bytes (internal array data) in a file using the file path
   #
+
```
fridis, line 86: remove this blank line.

PR #369, opened 2022-08-04
lib/lists: generate a list via supplier lamda

michaellilltokiwa:


fridis (commented):
This looks convenient. However, it requires supplier to be non-pure (depend on some effect), and it adds `mutate` effect used in `generate`.  I need to see examples how this is used, for the examples I imagine myself I think I would prefer avoiding this impurity.

PR #374, opened 2022-08-05
implementation of is_dir a feature to check whether the path is a directory or not

WaelYoussfi:


fridis (commented):
I think the documentation should be more specific here, in particular, how should links and sym links be treated: is a link to a directory considered a directory or not? How is this in Java? Rust? Python?

How does this compare to adding an intrinsic `mode` that in the C backend would do a `stat` and return  the mode of the file? What would this look like on Windows?

Here, you could do some API design work that look at options, compares alternatives and rates them (implementation complexity in backends, portability, complexity of Fuzion code, etc.).  Lots of possibilities that can be explored!

WaelYoussfi:
#623

PR #381, opened 2022-08-10
Effects/write

WaelYoussfi:


lib/io/file/write.fz
```diff
@@ -0,0 +1,65 @@
...
+        fwh.write path content
+        replace
+
+# short-hand for accessing write effect in current environment
+#
+write =>
+  writes.installDefault
+  write.env
+
+File_Write_Handler ref is
```
fridis, line 49: please add a comment for this feature.

PR #382, opened 2022-08-10
air: fix an issue that resulted in a cycle when creating clazzes.

michaellilltokiwa:
I'm sure there is a better way to do this! :-)
So I don't expect this PR to be merged but to be a starting point to fix stream.asString.
Without the fix the compiler would run into an infinite loop, creating clazzes like:
```
ref list<u8>.ref asStream.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.map<string>.ref @726549.fromStream.ref @730058.head
```
`ref @726549` is the arg f given to map.

### DIFF
--- a/lib/stream.fz
+++ b/lib/stream.fz
@@ -198,9 +198,8 @@ stream(redef T type) ref : Sequence<T> is
   # create a string from the elements of this stream
   #
   redef asString ref string is
-    # NYI: check if this works: map<string>(x-> x.asString).fold strings.concat ", "
-    #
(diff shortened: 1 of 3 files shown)


fridis:
Cool, the outer-relationship is one of the most complex and tricky parts of the fuzion compiler!

PR #385, opened 2022-08-11
lib: cache feature, enabling some kind of global state

michaellilltokiwa:
```
ex is

  say (cache string (() -> say "creating result1"; "the result"))
  say (cache string (() -> say "creating result2"; "the result"))

# output:
#
# creating result1
# the result
# the result
```


fridis:
Good, this is what I imagined this API should look like. Would be good to add an example to the documentation that defines a new type for the cached value (not caching a 'string' since that will quickly result in conflicts).

PR #400, opened 2022-08-30
Lib/set of  has hash elements

michaellilltokiwa:


fridis:
Note that `hasEquals`, `hasHash` and `ordered` will all change significantly, operations in these features like `hasHash.hash` will eventually move to `hasHash.type.hash`.

PR #403, opened 2022-08-31
adding error handling to the fileio.write feature

WaelYoussfi:
I am not sure whether I need to make the return type as outcome or just bool

### DIFF
--- a/lib/fuzion/std/fileio.fz
+++ b/lib/fuzion/std/fileio.fz
@@ -72,19 +72,18 @@ public fileio is
   # creates a new file using an array of u8 as content and a string as the file path
   # it overwrites the file, if it exists, or creates a new one if it does not exist
   # if the director(y/ies) in the path d(oes/o) not exist, or an error happens during the writing operation an error will be returned
-  # in case of success TRUE will be returned
+  # in case of success a unit type will be the outcome
   #
   public write(
                # the (relative or absolute) file name, using platform specific path separators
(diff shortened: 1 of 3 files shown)


WaelYoussfi:
> I am not sure whether I need to make the return type as outcome or just bool

<outcome unit> made more sense than returning true in case of success and error in case of failure

michaellilltokiwa:
what about `option error`?

michaellilltokiwa:
@WaelYoussfi `outcome unit` and `option error` functionally are the same thing. i think it comes down to taste. which conveys better what you want to return. maybe outcome is better since you use this in other places as well so this would be more consistent.

fridis:
IMHO, `outcome unit` is the right type here, since the result is `unit` in case of success. `option error` would have the disadvantage that we get an asymmetry with i/o operation that return a value different than `unit`. 
We might (later) decide to introduce a new unit type called something like `success` and use `outcome success` instead, which would make the `match`-clause more readable.

WaelYoussfi:
@michaellilltokiwa thanks for the clarification.
@fridis I think that `success` will be very helpful in case I actually want to return something other than `unit`.

michaellilltokiwa:
@fridis @WaelYoussfi To me `success` seems to be a helpful unit-type feature. It's more clear than `outcome unit` and I assume this outweighs the downside of having yet another unit-type feature.

PR #413, opened 2022-09-05
lib: implement big int, integers of arbitrary size

michaellilltokiwa:


michaellilltokiwa:
test takes about 100ms for ~891 calculations on my machine with c backend.

fridis:
Nice work. Can't wait to see the corresponding idioms being implemented!
Next steps, IMHO (not urgent, but the way big ints are implemented in, e.g., Java): 

- rename `nat` as `uint`
- change `nat` to use a sequence of `u64` instead of a sequence of `bool`
- maybe replace the sign in `int` by a `i64`, such that `int` is a sequence of `i64`, `u64`, `u64`, `u64`,...

PR #414, opened 2022-09-06
tests: remove FUZION_JAVA_STACK_SIZE, not needed anymore

michaellilltokiwa:
I think these are not needed anymore?
Tests are green without these on my machine.


fridis:
I still get a stack overflow for rosettacode_primes, the others seem to be find by now.

PR #416, opened 2022-09-06
io.file.delete will be the effect that wraps the delete operations

WaelYoussfi:


lib/io/file/delete.fz
```diff
@@ -0,0 +1,65 @@
...
+#  Author: Wael Youssfi (wael.youssfi@tokiwa.software)
+#
+# -----------------------------------------------------------------------
+
+# delete -- effect wrapping the file delete operation
+#
+private delete(
+    dop Delete_Operation,
+    redef r effectMode.val,
+    _ unit
```
fridis, line 31: This `unit` argument field does not seem to be needed, this is only used in some cases to distinguish a private constructor from a public one with the same number of arguments.

PR #419, opened 2022-09-07
lib: string remove asStream from feature `*` and `=`

michaellilltokiwa:


lib/string.fz
```diff
@@ -58,39 +58,16 @@ string ref : hasHash string, ordered string, strings is
...
-        redef next u8 is
-          res := bytes.next
-          if !bytes.hasNext
-            set i := i + 1
-            if i < n
-              set bytes := string.this.utf8.asStream
-          res
+    redef utf8 Sequence u8 is
+      utf8 := string.this.utf8
+      utf8.cycle.take (utf8.count * n)
```
fridis, line 63: nice. Maybe we should add a feature `list.cycle(n u32)` or `list.repeat(n u32)`, when this would be even simpler.

PR #421, opened 2022-09-07
lib: string, remove asString in integer parsing

michaellilltokiwa:


fridis:
good. I love to see `stream` disappear from our code!

PR #430, opened 2022-09-13
lib/string: fix string concat

michaellilltokiwa:
@fridis I think this should also prevent quadratic performance when concatenating strings since in the end two lists are lazily being concatenated. This fix does not really explain why the previous code for concatenation failed. I suspect there is an issue in marray.add but I did not investigate thoroughly yet.


fridis:
Much nicer, let's see if there is any performance degradation.

PR #437, opened 2022-09-14
lib: add as `as_int` to i8...u64

michaellilltokiwa:


lib/u64.fz
```diff
@@ -133,6 +133,7 @@ u64(val u64) : wrappingInteger u64, hasInterval u64, u64s is
 #      analysis: result.as_u64 == thiz
     is
       u128 0 thiz
+  as_int  => int as_i64
```
fridis, line 136: This will not work for values larger than 9223372036854775807.

fridis (commented):
one potential bug and some nit-picking...

PR #439, opened 2022-09-15
lib: uint add missing init feature; move exponentation feature

michaellilltokiwa:


lib/int.fz
```diff
@@ -108,18 +108,9 @@ is
...
-      is_even := other.n.bits.last = false
-      if is_even
-        tmp := thiz**(other / int 2)
-        tmp * tmp
-      else
-        thiz * (thiz**(other-one))
-
+    match s
+      plus => int plus (n ** other.n)
+      minus => int minus (n ** other.n)
```
fridis, line 113: This seems wrong: -1 ** 2 should be +1.

lib/int.fz
```diff
@@ -177,24 +168,12 @@ ints is
...
     s sign := if val < 0 then minus else plus
-    n u32 := if val < 0 then (-val).as_u32 else val.as_u32
-    (res,_) :=
-      (0..31)
-        .reduce (lists.empty bool, n) (r, _ ->
-          (l, rest) := r
-          ([(rest & 1) = 1] ++ l, rest>>1)
-        )
-    int s (uint res)
+    n u64 := if val < 0 then (-val).as_u64 else val.as_u64
```
fridis, line 171: `-val` will overflow for `val = i64.min`

fridis (commented):
I think I found too corner cases that are handled incorrectly...

lib/int.fz
michaellilltokiwa: the test also fails but I didn't notice... :-(

sam@debian ~/n/fuzion (lib_uint_init_and_exponentation)1$ fz-c tests/int/int_test.fz |grep FAIL
FAILED: int: m -5 ** n 0
FAILED: int: m -5 ** n 2
FAILED: int: m -5 ** n 4
FAILED: int: m -4 ** n 0
FAILED: int: m -4 ** n 2
FAILED: int: m -4 ** n 4
FAILED: int: m -3 ** n 0
FAILED: int: m -3 ** n 2
FAILED: int: m -3 ** n 4
FAILED: int: m -2 ** n 0
FAILED: int: m -2 ** n 2
FAILED: int: m -2 ** n 4
FAILED: int: m -1 ** n 0
FAILED: int: m -1 ** n 2
FAILED: int: m -1 ** n 4

lib/int.fz
michaellilltokiwa: @fridis good catch. :-) not sure what is the best way to fix this? add a pre-condition that disallows this one negative i64 or handle this single case explicitly?

PR #445, opened 2022-09-21
lib: sequence.reduce, allow abort of reduction.

michaellilltokiwa:
example:
```
ex is
  a := (1..).reduce 0 (r, t -> if r+t > 4 then abort r+t else r+t)
  say a
```


michaellilltokiwa:
Currently this PR leads to pre- and postcodition failures when building the base.fum.
If I disable pre/postconditions base.fum builds fine and the example runs.

fridis:
I first was surprised (since I did not see that you added a feature `abort`), then I was worried to pollute the name space by adding `abort`, but now I think I like this. 

Can you add an issue for the pre/postcondition errors you see?

Can you use `A | B` instead of `choice A B` in your code?

PR #470, opened 2022-10-20
lib/float: fix ceiling function

### LINKED ISSUE #469: f64: ceiling function broken?
Quoting the definition of the ceiling function from Wikipedia:
> the *ceiling function* maps x to the least integer greater than or equal to x, denoted ceil(x) or ⌈x⌉.
This is the generally accepted mathematical definition of the ceiling function.

With Fuzion, the following happens:
```
ex is
  a f64 := 1.0
  say a
  say a.ceil
```
prints
```
1.0
2.0
```

In contrast, in Rust this works as expected:
```rust
fn main() {
    let x = f64::from(1);
    println!("{} {}", x, x.ceil())
}
```
prints
```
1 1
```

maxteufel:
The previous implementation returned the least integer greater than x. The commonly accepted definition is to return the least integer greater than or equal to x.

Fixes #469.

### DIFF
--- a/lib/float.fz
+++ b/lib/float.fz
@@ -78,7 +78,7 @@ float(redef F (float F).type) : numeric F, floats F is
 
   # ceiling: the smallest integer greater or equal to this
   ceil F is
-    if fract < zero
+    if fract <= zero
       thiz - fract
     else
       thiz - fract + one


michaellilltokiwa:
Perfect.
Maybe it would be good to add some tests for ceil and floor? @maxteufel What do you think?

michaellilltokiwa:
I wrote both so I should know, but I don't because I think this separation does not really make sense?
I think we can merge those two.

michaellilltokiwa (approved):

PR #482, opened 2022-10-24
GitHub Actions: Add Linux workflow

maxteufel:
This should allow test runs using GitHub Actions. This workflow is triggered manually for the time being.


.github/workflows/linux.yml
```diff
@@ -0,0 +1,23 @@
+on:
```
fridis, line 1: Please add the GPLv3 and Tokiwa-header and include a short description on what this is.

PR #483, opened 2022-10-25
be/interpreter: implement fuzion.std.args.* intrinsics

### LINKED ISSUE #468: Java backend: pass arbitrary arguments to envir.args
If we compile a Fuzion program using the C backend, and run the resulting executable, we can then
```bash
./fuzionProgram arg1 arg2 arg3...
```
and access the arguments using the `envir.args` effect.

This does not work with the Java backend, because there is no way of passing arguments to the program.
```bash
fz fuzionProgram.fz arg1 arg2 arg3...
```
returns an error.

If we implement this, we should allow
```bash
fz fuzionProgram.fz -- arg1 arg2 arg3...
```
Note I have used `--` in the example to signify the end of command options, this is customary in other UNIX programs, see <https://unix.stackexchange.com/a/11382> for example.

maxteufel:
Make the `envir.args` effect work with the interpreter backend. Arguments can be passed as follows:

```
fz feature.fz -- arg1 arg2 arg3...
```

Fixes #468.

### DIFF
--- a/src/dev/flang/tools/Fuzion.java
+++ b/src/dev/flang/tools/Fuzion.java
@@ -518,8 +518,7 @@ class Fuzion extends Tool
     var std = STANDARD_OPTIONS(xtra);
     var stdRun = "[-debug[=<n>]] [-safety=(on|off)] [-unsafeIntrinsics=(on|off)] ";
     var stdBe = "[-modules={<m>,..}] [-moduleDirs={<path>,..}] [-sourceDirs={<path>,..}] " +
-      (xtra ? "[-XdumpModules={<name>,..}] " : "") +
-      "(<main> | <srcfile>.fz | -) ";
+      (xtra ? "[-XdumpModules={<name>,..}] " : "");
     if (_backend == Backend.undefined)
       {
         var aba = new StringBuilder();
@@ -549,7 +548,9 @@ class Fuzion extends Tool
         return "Usage: " + _cmd + " " + ba + " " + bu +
                            (b.runsCode() ? stdRun : "") +
                            stdBe + std +
-                           (b.takesApplicationArgs() ? "[-- <list of arbitrary arguments for envir.args effect>] " : "");
+                           (b.takesApplicationArgs() ? "[--] " : "") +
+                           "(<main> | <srcfile>.fz | -) " +
+                           (b.takesApplicationArgs() ? "[<list of arbitrary arguments for envir.args effect>] " : "");
       }
   }
 
@@ -723,6 +724,11 @@ class Fuzion extends Tool
             if (a.equals("-"))
               {
                 _readStdin = true;
+
+                if (_backend.takesApplicationArgs() || _backend == Backend.undefined)
+                  {
+                    argsLeft = new ArrayList<String>();
+                  }
               }
             else if ((_backend.takesApplicationArgs() || _backend == Backend.undefined) && a.equals("--"))
               {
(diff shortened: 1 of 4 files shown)


michaellilltokiwa:
@maxteufel  This is exactly how I imagined it.

src/dev/flang/tools/Fuzion.java
```diff
@@ -690,8 +692,16 @@ private Runnable parseArgsAceMode(String[] args)
    */
   private Runnable parseArgsForBackend(String[] args)
   {
+    var argsLeft = args;
+
     for (var a : args)
       {
+        argsLeft = Arrays.copyOfRange(argsLeft, 1, argsLeft.length);
```
fridis, line 699: This causes quadratic performance: for n arguments, you are copying n, n-1, n-2, etc. to a new array argsLeft. 
Quadratic performance must be avoided, this will very likely cause performance problems later.

src/dev/flang/tools/Fuzion.java
```diff
@@ -690,8 +692,16 @@ private Runnable parseArgsAceMode(String[] args)
...
+        if (a.equals("--"))
+          {
+            /* stop argument parsing */
+            break;
```
fridis, line 703: do not use `break`, this causes surprising control flow. 

Instead: Make `argsLeft` an `ArrayList<String>` that is set to `null` outside the for-loop. Set it to `new ArrayList<>()` if `"--"` is encountered. Before the `if`, add another 
```
  if (argsLeft != null) 
    {
      argsLeft.add(a);
    }
  else`if (a.equals("--"))
```

src/dev/flang/tools/Fuzion.java
```diff
@@ -690,8 +692,16 @@ private Runnable parseArgsAceMode(String[] args)
    */
   private Runnable parseArgsForBackend(String[] args)
   {
+    var argsLeft = args;
+
     for (var a : args)
       {
+        argsLeft = Arrays.copyOfRange(argsLeft, 1, argsLeft.length);
+        if (a.equals("--"))
```
fridis, line 700: This should probably be 
```
  if (_backend.runsCode() && a.equals("--"))
```
since backends like `-c` should not allow to pass arguments.

fridis (commented):
What is missing is an addition to the output on `-help` for this. I know the code and the resulting output of `-help` is a bit messy, but try to find your way through this...:-)

src/dev/flang/tools/Fuzion.java
fridis: If this is used for some other purpose by some backend, that backend should add support for `--`, we could (then, not now), e.g., replace the predicate `runsCode()` by `usesDashDash()` or similar. Backends that do not have this should refuse `--`.

fridis: Since this is the last argument anyway, the undefined backend can have `takesApplictionArgs()==true`. In a later PR, this might be cleaned up, e.g., by replacing the undefined backend by something like the default interpreter backend. The whole argument parsing code is quite messy...

src/dev/flang/tools/Fuzion.java
```diff
@@ -744,7 +781,23 @@ else if (_main != null)
...
+            if (mainOrStdin.equals("-"))
+              {
+                _readStdin = true;
+              }
+            else
+              {
+                _main = mainOrStdin;
+              }
+          }
+        else
```
fridis, line 797: I am a bit unhappy to have this logic duplicated here, but I also do not see an easy solution to avoid this, so leave it for now.

PR #488, opened 2022-10-25
lib/wrappingInteger: implement bitwise NOT

### LINKED ISSUE #475: unsigned integers: bitwise NOT
Perhaps I missed it, but it seems there is no bitwise NOT for unsigned integers. It's easy to add
```
not (n u64) u64 is
  n ^ u64s.max
```
but maybe something like the `~` prefix in C can be added.

maxteufel:
Fixes #475.

### DIFF
--- a/lib/i128.fz
+++ b/lib/i128.fz
@@ -49,6 +49,9 @@ i128s : wrappingIntegers i128 is
   redef min => i128 -0x_8000_0000_0000_0000 0
   redef max => i128  0x_7fff_ffff_ffff_ffff 0x_ffff_ffff_ffff_ffff
 
+  # returns the number in whose bit representation all bits are ones
+  redef all_bits_set => i128 -1 0x_ffff_ffff_ffff_ffff
+
 
 # i128 -- returns value of unit type i128s
 #
--- a/lib/integer.fz
+++ b/lib/integer.fz
@@ -49,6 +49,16 @@ integer(I (integer I).type) : numeric I is
   infix |  (other I) I is abstract
   infix ^  (other I) I is abstract
 
+  # bitwise NOT
+  prefix ~ I
+    pre
+      is_bounded
+  is abstract
+
+  # bitwise NOT (Unicode alias)
+  prefix ¬ I is
+    ~thiz
+
   # shift operations
   infix >> (other I) I
     pre
@@ -59,6 +69,11 @@ integer(I (integer I).type) : numeric I is
     pre
       safety: other.sign >= 0
   is abstract
+
+  # check if this type of integer is bounded
+  #
+  # returns false unless redefined by a specific implementation of integer
+  is_bounded => false
 
   # greatest common divisor of this and b
   #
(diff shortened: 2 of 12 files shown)


michaellilltokiwa:
Does this work for signed integers as well?
Maybe we should move `max`/`min` to integer and also introduce a new class `unsigned_integer` which inherits integer and also is inherited by the `uXX`?

michaellilltokiwa:
Good point. We would need to change the type of min/max to `option T` then. I'm still not sure if it makes sense to move it though...

michaellilltokiwa:
What is the reasoning behind `unsigned_integer` inheriting `wrappingInteger`?

Also if you inherit from `unsigned_integer` and this inherits from `wrappingInteger` there is no need to inherit from `wrappingInteger` again.

Example:
This is the same
```
u8(val u8) : unsigned_integer u8, hasInterval u8, u8s is
```
as this, I think:
```
u8(val u8) : unsigned_integer u8, wrappingInteger u8, hasInterval u8, u8s is
```

fridis:
I prefer not to add any additional features like `unsigned_integer` that extend the inheritance hierarchy even more. 
Making the result of `max` an `option T` sounds like the right thing to do, but would make its use very cumbersome. What I prefer would be a boolean feature `isBounded` or similar and use this as a precondition of `min` and `max`. To implement `prefix ~`, you should not use `max`, since this is wrong for signed integers, but add new feature `all_bits_set` (or a better / shorter name if you find one:-).

lib/i128.fz
```diff
@@ -49,6 +49,8 @@ i128s : wrappingIntegers i128 is
   redef min => i128 -0x_8000_0000_0000_0000 0
   redef max => i128  0x_7fff_ffff_ffff_ffff 0x_ffff_ffff_ffff_ffff
 
+  redef all_bits_set => i128 -0 1
```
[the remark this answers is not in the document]
michaellilltokiwa: I would think it needs to be `i128 i64.all_bits_set u64.all_bits_set`?

lib/integer.fz
```diff
@@ -60,6 +64,8 @@ integer(I (integer I).type) : numeric I is
       safety: other.sign >= 0
   is abstract
 
+  isBounded => false
```
michaellilltokiwa, line 67: I think you should already use the "new" rules for underscores/case for identifiers. So `is_bounded` instead of `isBounded`. See here:
https://flang.dev/design/identifiers

lib/wrappingInteger.fz
```diff
@@ -188,3 +197,5 @@ wrappingIntegers(W (wrappingInteger W).type) : numerics W is
 
   min W is abstract
   max W is abstract
+
+  all_bits_set W is abstract
```
fridis, line 201: please add a comment describing what this returns. Also, copy this comment to all implementations of this. The comment will end up in the API documentation on https://flang.dev/docs/, which looks a bit sad without any documentation. 
Same as with snake_case: We need comments for existing features as well, but in separate PRs.

lib/integer.fz
fridis: agreed. New names should be according to the rules. You might also rename existing names, but in a separate PR.

PR #493, opened 2022-10-26
scripts: use /usr/bin/env bash; set -euo pipefail

michaellilltokiwa:
> Running a command through /usr/bin/env has the benefit of looking for whatever the default version of the program is in your current environment.
> 
> This way, you don't have to look for it in a specific place on the system, as those paths may be in different locations on different systems. As long as it's in your path, it will find it.
> 
> One downside is that you will be unable to pass more than one argument (e.g. you will be unable to write /usr/bin/env awk -f) if you wish to support Linux, as [POSIX is vague](https://unix.stackexchange.com/a/399702/6252) on how the line is to be interpreted, and Linux interprets everything after the first space to denote a single argument. You can use /usr/bin/env -S on some versions of env to get around this, but then the script will become even less portable and break on fairly recent systems (e.g. even Ubuntu 16.04 if not later).
> 
> Another downside is that since you aren't calling an explicit executable, it's got the potential for mistakes, and on multiuser systems security problems (if someone managed to get their executable called bash in your path, for example).

https://stackoverflow.com/a/16365367

### DIFF
--- a/bin/ebnf.sh
+++ b/bin/ebnf.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This file is part of the Fuzion language implementation.
 #
--- a/bin/fz
+++ b/bin/fz
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This file is part of the Fuzion language implementation.
 #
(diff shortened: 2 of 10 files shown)


michaellilltokiwa:
@maxteufel In some it was already present but it seems I missed the record_* and check_* scripts :-(

Issue #158, opened 2022-02-07
base library: support for file I/O


fridis:

michaellilltokiwa:
@fridis Maybe we can find a generic I/O-abstraction that is not just specific to files but also sockets, pipes etc.?

PR #502, opened 2022-10-26
be/c: add ref to #158 in NYI-comments

maxteufel:


src/dev/flang/be/c/Intrinsics.java
```diff
@@ -123,12 +123,12 @@ public class Intrinsics extends ANY
             );
         }
         );
-    put("fuzion.std.fileio.get_file_size", noFileIo); // NYI
-    put("fuzion.std.fileio.write"        , noFileIo); // NYI
-    put("fuzion.std.fileio.exists"       , noFileIo); // NYI
-    put("fuzion.std.fileio.delete"       , noFileIo); // NYI
-    put("fuzion.std.fileio.move"         , noFileIo); // NYI
-    put("fuzion.std.fileio.create_dir"   , noFileIo); // NYI
+    put("fuzion.std.fileio.get_file_size", noFileIo); // NYI: #158
```
fridis, line 126: would be better if it had a bit of text...

Issue #167, opened 2022-02-09
Language improvement: Add type 'like outer.this' or similar.


fridis:
The lack of `like outer.this` resulted in errors like #166 and in somewhat ugly code for 'numeric.thiz' using a generic parameter. 
Having type `like outer.this` would allow to declare
```
outer is
  me like outer.this is
    outer.this
```
which would mean that the actual type the caller sees is the type of the target, i.e., if we have
```
  a outer is
  b ref : outer is

  x := a.me  # value type 'a'
  y := b.me  # ref type 'b'
```
this could even be used for argument types as in

```
outer is
  f(minime like outer.this) is
    ...

  a outer is
  b ref : outer is
  c : b is

  x := a
  x.f a         # ok, type of minime is value type 'a' 
  x.f outer    # error, type of minime is value type 'a', which is incompatible to 'outer'
  y := b
  y.f b         # ok, type of minime is ref type 'b'
  y.f c         # ok, type of minime is ref type 'b', value type 'c' can be boxed to 'ref c' which is compatible
  y.f outer    # error, type of minime is ref type 'b', which is incompatible to value type 'outer'
```

fridis:
This support was added in PR #717 through #1209.

PR #503, opened 2022-10-26
lib: add ref to #167 in NYI-comments

maxteufel:


lib/partiallyOrdered.fz
```diff
@@ -47,7 +47,7 @@ is
   # get value of type P.
   #
   # NYI: Once Eiffel-style 'like this' works, we can use 'like this' instead of
-  # T and no longer need this feature.
+  # T and no longer need this feature. See #167.
```
fridis, line 50: for consistency, add this directly after the `NYI:`.

PR #512, opened 2022-10-27
be/int: fix assign to nested choices

michaellilltokiwa:
c backend still fails these tests.


fridis (commented):
Good that your found this and good that you understand the code well enough to fix this for the interpreter. 
I suggest a different approach that should fix all backends automatically:
- first, create an issue with an example
- you can then add your test for the issue with a skip-entry
- the solution should be part for the front end, when instances of `Tag` are added to the AST: We might need several nested `Tag` instances.
- Then, the interperter and C backend code should stay the same
- finally, it would be great to have tests for more corner cases:
  - overlap, e.g., a feature `X : choice A, B` of choice types `A : choice i32, boo`l and `B : choice bool, i32, string` with overlapping types `bool` and `i32`. This should probably result in errors, need to check that error messages are useful. 
  - the case of nested choices where all entries are ref types (quite sure that you will find some existing problems there)

michaellilltokiwa:
superseded by #535

PR #515, opened 2022-10-27
file `exists` intrinsic in c backend

WaelYoussfi:

PR #516, opened 2022-10-27
Fileio/delete

WaelYoussfi:
#515 needs to be merged for the `delete` intrinsic to work in C be

### DIFF
--- a/lib/fuzion/std/fileio.fz
+++ b/lib/fuzion/std/fileio.fz
@@ -137,16 +137,10 @@ public fileio is
                 # the (relative or absolute) file name, using platform specific path separators
                 path string) outcome unit is
     arr := path.utf8.asArray
-    match exists path
-      b bool =>
-        if b = TRUE
-          if delete arr.internalArray.data unit
-            unit
-          else
-            error "an error occurred while performing the delete operation on the following file/dir: \"$path\""
-        else
-          error "file/dir \"$path\" does not exist"
-      err error => err
+    if delete arr.internalArray.data unit
+      unit
+    else
+      error "an error occurred while performing the delete operation on the following file/dir: \"$path\""
 
   # intrinsic that deletes the file/dir represented by the path returning TRUE in case of success
   # and FALSE in case of failure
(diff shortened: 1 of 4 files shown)


lib/fuzion/std/fileio.fz
```diff
@@ -137,19 +137,25 @@ public fileio is
                 # the (relative or absolute) file name, using platform specific path separators
                 path string) outcome unit is
     arr := path.utf8.asArray
-    if delete arr.internalArray.data arr.length
-      unit
-    else
-      error "an error occurred while performing the delete operation on the following file/dir: \"$path\""
+    match exists path
+      b bool =>
+        if b = TRUE
```
fridis, line 142: `if b = TRUE` should just be `if b`

lib/fuzion/std/fileio.fz
```diff
@@ -137,19 +137,25 @@ public fileio is
...
+          if delete arr.internalArray.data unit
+            unit
+          else
```
fridis, line 145: instead of 
```
if f a b c
  unit
else
  error
```
you should better write
```
if ! f a b c
  error
```
or
```
if !(f a b c)
  error
```

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -273,7 +273,7 @@ public static Callable call(Interpreter interpreter, Clazz innerClazz)
           Path path = Path.of(new String(pathBytes, StandardCharsets.UTF_8));
           try
             {
-              boolean b = Files.deleteIfExists(path);
+              boolean b = Files.deleteIfExists(path); // NYI : not sure whether to keep it or use Files.delete instead because the existence check already takes place in Fuzion
```
fridis, line 276: Even if there is an existence test before, another process might have created the file at any time.  So the question is what is the behaviour of `Files.delete` if it does not exist. If this would throw an `IOException`, this would not be good since then we could see sporadic exceptions for a file that is created by a concurrent process, `deleteIfExists` might be better.

PR #528, opened 2022-10-31
lib/choice: use modern syntax in documentation

maxteufel:


lib/choice.fz
```diff
@@ -83,8 +83,8 @@
 #
 # When assigning values to temp, we need to wrap the values accordingly:
 #
-#   set temp := centigrade 37
-#   set temp := fahrenheit 99
+#   temp := centigrade 37
+#   temp := fahrenheit 99
```
fridis, line 87: The assignment to `temp` here declares new variables with the same name, so they are of type `centrigrade` or `fahrenheit`, not a choice type. 
Can you change this example to declare `temp` as an argument, e.g., 

```
hasFever(tmp centigrade | fahrenheit) bool is ...
```
and then do 
```
  hasFever (centrigrad 37)
  hasFever (fahrenheit 99)  
```

lib/choice.fz
```diff
@@ -79,12 +79,12 @@
 # Now we define the choice type using the wrapped i32 types, which are
 # distinct:
 #
-#   temp centigrade | fahrenheit := ...
+#   hasFever(temp centigrade | fahrenheit) bool is ...
 #
-# When assigning values to temp, we need to wrap the values accordingly:
+# When giving values to this feature, we need to wrap them accordingly:
```
michaellilltokiwa, line 84: @maxteufel What do you think about this:
~~When giving values to this feature~~
When passing arguments to this feature

PR #530, opened 2022-10-31
lib: use modern syntax in various places

maxteufel:


lib/numOption.fz
```diff
@@ -139,7 +139,7 @@ is
 #
 # insted of
```
fridis, line 140: `instead`

lib/void.fz
```diff
@@ -76,7 +76,7 @@ void(absurd void) is absurd
 
 # NYI: possible alternative definitions base on choice types
 #
-# void : choice <> is
+# void : choice () is
 #
-# void (absurd choice<>) is absurd
+# void (absurd choice ()) is absurd
```
fridis, line 81: I think this should be `choice`, and not `choice ()` since the latter is a choice of the empty tuple.

PR #533, opened 2022-11-01
dfa: return false if effect does not exist

michaellilltokiwa:


fridis:
When did you run into a problem with this? 
The `_bool` results means either `true` or `false` since we do not know which one is correct. This is correct for effects that install a default effect explicitly, but it would be nicer if, as a result of DFA, code would be added to install these defaults automatically, then `exists` could always be `true` for effects with a default, and be `false` for those without a default implementation that are not installed explicitly by the application code.

michaellilltokiwa:
Ah okay. Makes sense. I did not understand it then.

PR #539, opened 2022-11-02
`move` intrinsic in C backend

WaelYoussfi:


WaelYoussfi:
> It took me quite a while to understand that the dummy parameter (in this and the subsequent pull requests) is necessary to make feature overloading work (also in part because GitHub does not show the relevant code before). Perhaps this can be mentioned in the comments?

@maxteufel you are right. I should mention that in the comment just before the parameter, which I did in a previous PR  #516, but I thought it would be redundant to mention it in every occurrence. However, I think I will be adding the comments, nevertheless, to avoid any confusions.

fridis:
@WaelYoussfi, can you resolve the conflicts?

WaelYoussfi:
> @WaelYoussfi, can you resolve the conflicts?

Resolved

PR #540, opened 2022-11-02
Fileio/stat `get_file_size` intrinsic in C be

WaelYoussfi:
For symlinks C `stat` resolves the link and returns the stats of the actual file/dir

PR #541, opened 2022-11-02
`create_dir` intrinsic in the c be

WaelYoussfi:
#540 needs to be merged first


fridis:
This also suffers from #570, but I pull it for now.

Issue #536, opened 2022-11-02
`if` statement does not work as expected when a branch returning `unit` type is not explicitly provided


WaelYoussfi:
The following code should reproduce the problem:
```
ex =>
  log(a bool) outcome unit is
    if !a
       error "a is false"

  say (log (envir.args.count < 2))
```

When `a` is a constant, both the C backend and the interpreter return an error at compile time but in the aforementioned case it results in a segmentation fault in C be and NullPointerException in the interpreter be.

michaellilltokiwa:
I can confirm the error in the c-backend but in interpreter the example works for me:
```
--error: a is false--
```

PR #544, opened 2022-11-02
ast/If: change if semantics to always return result. unit being the default.

michaellilltokiwa:
I think `unit` is correct here since further down in the code:
```java
private AbstractType typeFromIfOrElse()
  {
    AbstractType result;
    if (hasUntakenElseBranch())
      {
        result = Types.resolved.t_unit;
      }
...
```


fridis:
The original comment is wrong and your correction matches the behaviour. But reviewing the code of `typeFromIfOrElse`

```
    if (hasUntakenElseBranch())
      {
        result = Types.resolved.t_unit;
      }
    else
      {
        result = Types.resolved.t_void;
        Iterator<Expr> it = branches();
        while (it.hasNext())
          {
            var t = it.next().typeForFeatureResultTypeInferencing();
            result = result == null || t == null ? null : result.union(t);
          }
      }
```
I think this should be
```
    if (hasUntakenElseBranch())
      {
        result = Types.resolved.t_unit;
      }
    else
      {
        result = Types.resolved.t_void;
      }
    Iterator<Expr> it = branches();
    while (it.hasNext())
      {
        var t = it.next().typeForFeatureResultTypeInferencing();
        result = result == null || t == null ? null : result.union(t);
      }
```
Can you try if the tests run fine with this change? Is this related to #536?

michaellilltokiwa:
@fridis I reworked the PR and changed the semantics of `if` to always return a result (default=`unit`) even when there is no explicit else branch.

src/dev/flang/ast/If.java
```diff
@@ -407,20 +375,42 @@ public void propagateExpectedType(Resolution res, AbstractFeature outer)
...
-      elseBlock != null ? elseBlock :
-      elseIf    != null ? elseIf    : new Block(pos(), new List<>());
+    if (PRECONDITIONS) require
+      (elseBlock != null || elseIf != null);
+
+    if (elseBlock != null)
+      {
+        return elseBlock;
+      }
+    return elseIf;
```
fridis, line 413: No a fan of `return` in nested blocks, use 
```
return 
  elseBlock != null ? elseBlock 
                    : elseIf;
```
instead (even if `? :` for many developers needs some time to get used to...).

fridis (commented):
Very nice patch, love to see `If.java` getting 25 lines shorter!
Just minor style nit-picking...

PR #545, opened 2022-11-03
ast/If: issue 536, default else branch returns `unit`

### LINKED ISSUE #536: `if` statement does not work as expected when a branch returning `unit` type is not explicitly provided

michaellilltokiwa:
fixes #536

### DIFF
--- a/src/dev/flang/ast/If.java
+++ b/src/dev/flang/ast/If.java
@@ -356,7 +356,7 @@ public class If extends ExprWithPos
     // special case, default else-branch returning `unit`:
     // there is no explicit else-branch but we assign if to a field which
     // we can assign `unit` to. Thus we generate a default else branch.
-    if(hasUntakenElseBranch() && r.resultType().isAssignableFrom(Types.resolved.t_unit))
+    if (hasUntakenElseBranch() && r.resultType().isAssignableFrom(Types.resolved.t_unit))
       {
         var unit = new Call(pos(), null, "unit").resolveTypes(res, outer);
         elseBlock = new Block(pos(), new List<Stmnt>(unit))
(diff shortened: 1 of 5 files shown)


src/dev/flang/ast/If.java
```diff
@@ -354,6 +353,15 @@ If assignToField(Resolution res, AbstractFeature outer, Feature r)
       {
         elseIf = elseIf.assignToField(res, outer, r);
       }
+    // special case, default else-branch returning `unit`:
+    // there is no explicit else-branch but we assign if to a field which
+    // we can assign `unit` to. Thus we generate a default else branch.
+    if(hasUntakenElseBranch() && r.resultType().isAssignableFrom(Types.resolved.t_unit))
```
fridis, line 359: please add space after `if`

src/dev/flang/fuir/analysis/Escape.java
```diff
@@ -97,7 +97,8 @@ public Escape(FUIR fuir)
   public boolean doesCurEscape(int cl)
   {
     if (PRECONDITIONS) require
-      (_fuir.clazzKind(cl) == IR.FeatureKind.Routine);
+      ( _fuir.clazzKind(cl) == IR.FeatureKind.Routine
+      ||_fuir.clazzKind(cl) == IR.FeatureKind.Intrinsic);
```
fridis, line 101: please add space after `||`

fridis (commented):
Looks very good, just very minor formatting nitpicking in my comments. 
Can you update #544 such that I can pull in both PRs at once.

PR #548, opened 2022-11-03
fzjava: do not generate duplicate features

### LINKED ISSUE #463: fzjava create duplicate features for Java packages
fzjava currently does not recognize Java packages that already have corresponding Fuzion features, these features have to be deleted manually. 
A solution would be to have fzjava process module files that the new module file depends on and find which package features are redundant.

maxteufel:
My current working state of a fix for issue #463. It works already, hence the 'rm's have been removed from the Makefile, except for one:
```bash
rm -f $(BUILD_DIR)/modules/java.desktop/Java/java/awt_pkg.fz
```
It seems that this file does not even get generated without this change. Perhaps it's just wrong copy and paste from java.datatransfer?

Fixes #463.

### DIFF
--- a/src/dev/flang/fe/FrontEnd.java
+++ b/src/dev/flang/fe/FrontEnd.java
@@ -112,7 +112,7 @@ public class FrontEnd extends ANY
    * The library modules loaded so far.  Maps the module name, e.g. "base" to
    * the corresponding LibraryModule instance.
    */
-  private TreeMap<String, LibraryModule> _modules = new TreeMap<>();
+  public TreeMap<String, LibraryModule> _modules = new TreeMap<>();
 
 
   /**
--- a/src/dev/flang/fe/FrontEnd.java
+++ b/src/dev/flang/fe/FrontEnd.java
@@ -34,6 +34,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 
+import java.util.Collection;
 import java.util.EnumSet;
 import java.util.TreeMap;
 
@@ -112,7 +113,7 @@ public class FrontEnd extends ANY
    * The library modules loaded so far.  Maps the module name, e.g. "base" to
    * the corresponding LibraryModule instance.
    */
-  public TreeMap<String, LibraryModule> _modules = new TreeMap<>();
+  private TreeMap<String, LibraryModule> _modules = new TreeMap<>();
 
 
   /**
@@ -319,6 +320,15 @@ public class FrontEnd extends ANY
       {
         m.loadInnerFeatures(f);
       }
+  }
+
+
+  /**
+   * Return the collection of loaded modules.
+   */
+  public Collection<LibraryModule> getModules()
+  {
+    return _modules.values();
   }
 
 }
--- a/src/dev/flang/tools/fzjava/FZJava.java
+++ b/src/dev/flang/tools/fzjava/FZJava.java
@@ -31,6 +31,7 @@ import dev.flang.ast.AbstractFeature;
 import dev.flang.fe.FrontEnd;
 import dev.flang.fe.FrontEndOptions;
 
+import dev.flang.tools.FuzionHome;
 import dev.flang.tools.Tool;
 
 import dev.flang.util.Errors;
@@ -213,7 +214,7 @@ public class FZJava extends Tool
       {
         List<String> emptyList = new List<>();
         var feOptions = new FrontEndOptions(/* verbose */ 0,
-                                            /* fuzionHome */ _options._fuzionHome,
+                                            /* fuzionHome */ (new FuzionHome())._fuzionHome,
                                             /* loadBaseLib */ true,
                                             /* eraseInternalNamesInLib */ true,
                                             /* modules */ _options._loadModules,
--- a/src/dev/flang/tools/fzjava/FZJava.java
+++ b/src/dev/flang/tools/fzjava/FZJava.java
@@ -232,13 +232,7 @@ public class FZJava extends Tool
                                             true);
         _fe = new FrontEnd(feOptions);
 
-        for (var m : _fe._modules.values())
-          {
-            if (m.name() != "base")
-              {
-                recurseDeclaredFeatures(m, _fe._universe);
-              }
-          }
+        recurseDeclaredFeatures(_fe, _fe._universe);
 
 
         for (var m : _options._modules)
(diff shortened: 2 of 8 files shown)


fridis:
> The build here fails because with PRECONDITIONS enabled, FrontEndOptions says you can't have readStdin = false and main = null. This is obviously sensible when using `fz`, but not here...

Then find a reasonable way to relax this precondition such that it still gets checked for `fz` but does not fail for `fzjava`.

src/dev/flang/fe/FrontEndOptions.java
```diff
@@ -133,7 +133,8 @@ public FrontEndOptions(int verbose,
                          List<String> sourceDirs,
                          boolean readStdin,
                          String main,
-                         boolean loadSources)
+                         boolean loadSources,
+                         boolean preconditionForNoInput)
```
fridis, line 137: No, this was not what I meant!

Try not to add complexity, but try to understand why a condition is there and refine it.
In FrontEnd, I see that readStdin/main are needed iff loadSources is true, so a solution might be to add !loadSources || in front of the precondition.

Alternatively, you could remove the precondition here and add it somewhere, e.g., to the constructor of FrontEnd, but I prefer to check this as early as possible.

Thinking about it more: why is readStdin true at all? I do not think that fzjava should ever read from stdin. How do you run into this?

src/dev/flang/tools/fzjava/FZJava.java
```diff
@@ -185,6 +216,33 @@ void execute()
...
+                                            true,
+                                            true,
+                                            _options._loadModules,
+                                            _options._moduleDirs,
+                                            emptyList,
+                                            0,
+                                            true,
+                                            true,
+                                            emptyList,
+                                            false,
```
fridis, line 231: `readStdin` is false, so the precondition should not fail.

michaellilltokiwa:
@maxteufel maybe I'm totally wrong but don't you need to re instantiate frontend every time you process a module?
```
       _fe = new FrontEnd(feOptions);

        for (var m : _fe._modules.values())
          {
            if (m.name() != "base")
              {
                recurseDeclaredFeatures(m, _fe._universe);
              }
          }


        for (var m : _options._modules)
          {
            if (!m.endsWith(".jmod"))
              {
                m = m + ".jmod";
              }
            processModule(m);
          }
```

michaellilltokiwa:
@maxteufel I thought after processing a module there might be a new fum-file not yet loaded by FrontEnd?

michaellilltokiwa:
I see. I think you need to go one at a time, yes...

fridis:
> @fridis Currently fzjava supports processing multiple java modules in one run. This is not used by the Makefile, however if it were used, we could not make use of the duplicate detection (see comments above). What is your opinion on making fzjava take just one argument? If wanted, I can make a separate pull request for that.

I do not remember why I allowed multiple modules, so one argument is IMHO fine. Change this in a separate PR.

src/dev/flang/fe/FrontEnd.java
```diff
@@ -112,7 +112,7 @@ public FeatureName featureName()
    * The library modules loaded so far.  Maps the module name, e.g. "base" to
    * the corresponding LibraryModule instance.
    */
-  private TreeMap<String, LibraryModule> _modules = new TreeMap<>();
+  public TreeMap<String, LibraryModule> _modules = new TreeMap<>();
```
fridis, line 115: If you make this `public`, you should at least make it `final`. 

Better would be to keep this `private` and add a method `modules()` that returns `_modules.values()`.

src/dev/flang/tools/fzjava/FZJava.java
```diff
@@ -86,6 +92,21 @@ public class FZJava extends Tool
...
+  private FrontEnd _fe;
+
+
+  /**
+   * The features that have already been defined in the loaded library modules.
+   * Maps the qualified name of the feature to the qualified name of the feature.
+   * This is a map rather than a set because Java's Sets do not provide efficient
+   * methods for checking whether an element is in a set.
+   */
+  TreeMap<String, String> _existingFeatures = new TreeMap<String, String>();
```
fridis, line 107: This is not true, `TreeSet.contains(x)` is just as efficient has `TreeMap.get(x) != null`.

src/dev/flang/tools/fzjava/FZJava.java
```diff
@@ -185,6 +214,26 @@ void execute()
   {
     if (createDestDir())
       {
+        List<String> emptyList = new List<String>();
```
fridis, line 217: You can write
```
List<String> emptyList = new List<>(); 
```
or
```
var emptyList = new List<String>(); 
```

src/dev/flang/tools/fzjava/FZJava.java
fridis: Different issue here: Would be good to add comments like `/* readStdin */ false` to make this more readable.

src/dev/flang/tools/fzjava/FZJava.java
```diff
@@ -426,6 +502,11 @@ void createOuter(String jfn)
     if (!_pkgs.contains(pkg))
       {
         _pkgs.add(pkg);
+        if (_existingFeatures.get(pkg.replace("/", ".")) != null)
+          {
+            // do not generate duplicate features
+            return;
+          }
         FeatureWriter.write(this, pkg, "_pkg", FeatureWriter.mangle(pkg.replace("/",".")) + " is\n");
```
fridis, line 510: Please do not use `return` (or `break`) in nested blocks, this results in non-intuitive control flow. Better do
```
        if (_existingFeatures.get(pkg.replace("/", ".")) == null)
          {
            FeatureWriter.write(this, pkg, "_pkg", FeatureWriter.mangle(pkg.replace("/",".")) + " is\n");
          }
```

src/dev/flang/tools/fzjava/FZJavaOptions.java
```diff
@@ -78,6 +96,19 @@ public class FZJavaOptions extends ANY
...
+   * _fuzionHome.
+   */
+  private String _fuzionHomeProperty = System.getProperty(FUZION_HOME_PROPERTY);
+
+
+  /**
+   * Home directory of the Fuzion installation.
+   */
+  Path _fuzionHome = _fuzionHomeProperty != null ? Path.of(_fuzionHomeProperty) : null;
+
```
fridis, line 111: This code and the declaration of FUZION_HOME_PROPERTY is duplicated from `Fuzion.java`. Would be better to extract this into a new clazz `FuzionHome` and use that both here and in `Fuzion.java`.

fridis (commented):
Very nice work! Some style related comments from me, as usally. 
One suggestion I have: There are typically very few Java packages (may tens), while there are easily tens of thousands of features in the modules. So instead of building the complete set of `_existingFeatures`, it will likely be much faster to check only the packages on demand. Should maybe be done in a separate PR.

PR #552, opened 2022-11-03
replace old style generics by new style in toString methods

michaellilltokiwa:
It's not perfect yet but it's a start.


src/dev/flang/ast/Call.java
```diff
@@ -959,8 +959,8 @@ public String toString()
             ? ""
             : target.toString() + ".")
       + (name != null ? name : _calledFeature.featureName().baseName())
-      + (_generics.isEmpty() ? "" : "<" + _generics + ">")
-      + (_actuals.isEmpty() ? "" : "(" + _actuals +")")
+      + new List(_generics, _actuals)
+        .toString(" ", " ", "", "(", ")", x -> x.toString().contains(" "))
```
michaellilltokiwa, line 963: @maxteufel @WaelYoussfi @fridis: I need a way to determine when to add braces here. Right now I use `arg.toString().contains(" ")` Any better ideas?

fridis: There is no easy solution. One solution may be to add a boolean flag `mayContainWhiteSpace` to `toString` and at any location that would add white space (e.g., joining the arguments with a call), put the result in `(` / `)` if that flag is set.

fridis (approved):

PR #558, opened 2022-11-04
add some useful profiling, debugging scripts/recipes

michaellilltokiwa:


michaellilltokiwa:
@maxteufel good point.

Issue #562, opened 2022-11-04
allow nested choices to have none disjoint elements


michaellilltokiwa:
```
ex =>
  another_choice : choice (choice nil string) nil is
```
```
sam@debian ~/n/fuzion (main)$ fz-c ~/playground/test.fz 

/home/sam/playground/test.fz:2:3: error 1: Generics arguments to choice type must be disjoint types
  another_choice : choice (choice nil string) nil is
--^
The following types have overlapping values:
'choice<nil, string>'
'nil'

one error
```
We should loosen this restriction.

PR #561, opened 2022-11-04
fix #556, when choice elements of outer and inner choice are not disjoint

michaellilltokiwa:
There might be other occurrences in the code where the use of `isAssignableFrom` is not sufficient.

PR #535, opened 2022-11-01
assignment to nested choice types 

michaellilltokiwa:
fixes issues #492,  #213

PR #563, opened 2022-11-04
fix issue #562: allow nested choices to have none disjoint elements

michaellilltokiwa:
depends on PR #535 and #561


michaellilltokiwa:
Also it might be unexpected that nil gets assigned to the inner choice.

michaellilltokiwa:
fixes #562

PR #565, opened 2022-11-04
`exist` effect provide feature `exists` to test files/directories existence

WaelYoussfi:
still wondering whether `exist` should be an effect by its own or a part of another effect.


fridis:
> still wondering whether `exist` should be an effect by its own or a part of another effect.

maybe this could be an operation in the `stat` effect instead? 

Another thing: This should be called `exists`, not `exist`. The plural variant could be either `exists_s` or `exists_type` since all the `*s` unit types will eventually be moved to the corresponding type features called `*.type`.

WaelYoussfi:
#582 will replace this.

Issue #559, opened 2022-11-04
Effects are not being properly installed when inherited


WaelYoussfi:
the following code should reproduce the issue:
```
test : io =>
  out.print "a"
```
here is the error message:
```
error 1: *** effect for io.out not present in current environment
    available are [test.out]
```

michaellilltokiwa:
#448 is same.

michaellilltokiwa:
@fridis @maxteufel 
I was revisiting this hoping the change in handling of outer types somehow solves this... When compiling to C the error currently is:
```
issue559.c:8785:34: error: no member named 'fzEnvInstalled_1476' in 'struct fzThrd_effectsEnvironment'; did you mean 'fzEnvInstalled_1477'?
  if (fzThrd_effectsEnvironment->fzEnvInstalled_1476)
                                 ^~~~~~~~~~~~~~~~~~~
                                 fzEnvInstalled_1477
issue559.c:8723:8: note: 'fzEnvInstalled_1477' declared here
  bool fzEnvInstalled_1477;
       ^
issue559.c:8907:37: error: no member named 'fzEnvInstalled_1476' in 'struct fzThrd_effectsEnvironment'; did you mean 'fzEnvInstalled_1477'?
    if (!fzThrd_effectsEnvironment->fzEnvInstalled_1476)
                                    ^~~~~~~~~~~~~~~~~~~
                                    fzEnvInstalled_1477
issue559.c:8723:8: note: 'fzEnvInstalled_1477' declared here
  bool fzEnvInstalled_1477;
```
I think we need to require features such as 
```
# short-hand for accessing out effect in current environment
out =>
  outs.installDefault
  io.out.env
````
that return an instance from an env to be marked as fixed? Or can we express somehow that `.env` always returns a `fixed` feature? Or do we want to allow inheriting features returning an effect and thus need to change the code figuring out which effects are being used? In the above case being `test.out`

michaellilltokiwa:
I believe this is now fixed.

PR #566, opened 2022-11-05
fuir/be: fix proposal for installation and usage of effects

michaellilltokiwa:


michaellilltokiwa:
should fix #559

fridis:
I have to think about this for a while. The problem here is that `ex559` inherits from the outer feature `io` of the effect `io.out`, so the installed effect is `ex559.out`, while what is accessed later is `io.out` which is not found since it is a different clazz.

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -911,12 +911,13 @@ static Callable effect(Interpreter interpreter, Clazz innerClazz)
         String in = innerClazz.feature().qualifiedName();   // == _fuir.clazzIntrinsicName(cl);
         switch (in)
           {
-          case "effect.replace": check(FuzionThread.current()._effects.get(cl) != null); FuzionThread.current()._effects.put(cl, m   );   break;
-          case "effect.default": if (FuzionThread.current()._effects.get(cl) == null) {  FuzionThread.current()._effects.put(cl, m   ); } break;
+          case "effect.replace": check(FuzionThread.current()._effects.get(cl.clazzOfType()) != null); FuzionThread.current()._effects.put(cl.clazzOfType(), m   );   break;
```
fridis, line 914: Instead of calling `cl.clazzOfType()` repeatedly, can you initailize `var et = innerClazz._outer.effectType()` and then use `et` throughout instead of `cl`?

fridis (commented):
Came to the conclusion that this should be changed as suggested. But please rename `clazzOfType`as `effectType`.

michaellilltokiwa:
fails with:
```
error 1: NYI: in issue559.out.print no targets for Call to issue559.ref Can_Print.print target boxed(issue559.ref defaultPrintToOut):issue559.defaultPrintToOut
```

michaellilltokiwa:
> Came to the conclusion that this should be changed as suggested. But please rename clazzOfTypeas effectType.

@fridis Can you please revisit this and tell me if your opinion on this is still the same?

michaellilltokiwa:
Will revisit this problem later

PR #567, opened 2022-11-05
lib: io.stdin, add features byte, bytes, codepoint, codepoints

michaellilltokiwa:


lib/io/stdin.fz
```diff
@@ -35,14 +35,46 @@ private stdin(
...
+
+  # try reading a codepoint from stdin
+  codepoint =>
+    read_codepoint(read_bytes list u8) choice error end_of_file codepoint is
+      match byte
+        byte u8 =>
+          bytes := read_bytes ++ [byte]
+          match (strings.fromBytes bytes).codepointsAndErrors.first
+            c codepoint => c
+            e error => if bytes.count >= 4 then e else read_codepoint bytes
```
fridis, line 61: instead of trying 4 times and checking `count >= 4`, it would IMHO be better to add a feature `codepoints.num_utf8_bytes(first_byte u8)` that returns `1` for 0..0x7f, `2`for `0b110xxxxx`, `3` for `0b1110xxxx`, and `4` for `0b11110xxx`. 
(In a separate patch,) You could even add `u8.leadingZeroes` and implement `num_utf8_bytes(first_byte u8) => (~first_byte).leadingZeroes+1`.

lib/io/stdin.fz
```diff
@@ -35,14 +35,46 @@ private stdin(
...
+
+    read_codepoint (lists.empty u8)
+
+
+  # a - possibly infinite - list of codepoints read from stdin
+  # Caution!: this list is not immutable since it is constructed lazily by reading from stdin!
+  # you may iterate only once.
+  # Additionally there may be read more from stdin than what is consumed
+  # via this list.
+  # NYI: How can we lessen these imperfections?
```
fridis, line 73: One way around this would be to memoize the result, i.e, add a field
```
codepoints_result := mut (option (list codepoint) nil)
```
and then check this in `codepoints`, return it if it is non-`nil` and read the next codepoint if it is. In case a non-empty list is returned, the new list must itself contain the `head` codepoint and use memoization for `tail`.

michaellilltokiwa: @fridis I think there might be other subtle issues though.
Lets say i do something like `codepoints.take 0` I would expect it to read zero `codepoints`. But actually it would read one since it has to construct a cons cell.

lib/codepoint.fz
```diff
@@ -108,3 +108,21 @@ codepoints is
...
+  num_utf8_bytes (first_byte u8) outcome i32 is
+    c := first_byte.as_u32 << 24
+    if utf8EncodedInOneByte.contains c
+      1
+    else if utf8EncodedInTwoBytes.contains c
+      2
+    else if utf8EncodedInThreeBytes.contains c
+      3
+    else if utf8EncodedInFourBytes.contains c
+      3
```
fridis, line 126: `4`

lib/io/stdin.fz
fridis: `take 0` just returns `nil`.

fridis:
This should eventually be moved to another feature to be reusable for input from other sources than `stdin`.

PR #586, opened 2022-11-08
`read` effect : adding short-hand to create effects

WaelYoussfi:


lib/io/file/read.fz
```diff
@@ -40,6 +40,17 @@ private read(
...
 
+# short-hand for installing and creating effects
+#
+read_custom(f () -> unit) => read_custom default_read_operation f
+
+# short-hand for installing and creating effects
+#
+read_custom(ro Read_Operation, f () -> unit) =>
+  s := read ro effectMode.new unit
+  s.use0 f # NYI : FUZION/lib/effect.fz:98:7: error 1: Precondition does not hold
```
fridis, line 51: this comment makes me worried...

WaelYoussfi: that is the error I get whenever I try installing an effect to run with a specific code using `effectMode.new` instead of the old `effectMode.inst`.
This even happens when I tried it with `io.stdin`.
I do not understand the reason behind it that is why I marked it as NYI.

PR #587, opened 2022-11-08
Issues/176 add regression test

### LINKED ISSUE #176: backend/C: C compilation error when using inner field and outer is boxed
This occurs in severial idioms when using the C backend:
```
 > cat idiom8ex.5.fz
ex8 is
  x r := v

  v : r is
    e(i i32) is
    o e | nil := e 0

  r ref is
 > ./build/bin/fz -c idiom8ex.5.fz
ex8.c:683:21: error: assigning to 'fzT_ex8__v__1e' (aka 'struct fzT_ex8__v__1e') from incompatible type 'fzT_ex8_Rv__1e' (aka 'struct fzT_ex8_Rv__1e')
  fzM_1.fzChoice.v0 = fzM_0;
                    ^ ~~~~~
1 error generated.

error 1: C backend: C compiler failed
C compiler call 'clang -O3 -o ex8 ex8.c' failed with exit code '1'

one error.
```
The problem is that the C backend declares two distinct C types for `v.e` and `ref v.e`, which are incompatible.

michaellilltokiwa:
closes #176


tests/reg_issue176_using_inner_field_of_boxed_outer/issue176.fz
```diff
@@ -0,0 +1,38 @@
...
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test issue176
+#
+#  Author: Fridtjof Siebert (siebert@tokiwa.software)
```
fridis, line 22: This is not from me :-)

fridis (commented):
good to have many regression tests!
I prefer if this is a simple.mk test, our you should add code that prints `PASS`/`FAIL` and dies `exit 1` in case of an error.

tests/reg_issue176_using_inner_field_of_boxed_outer/issue176.fz
michaellilltokiwa: I added this header since you created this issue. ;-)

PR #590, opened 2022-11-08
Effects/move  shorthand to install and create effect

WaelYoussfi:


lib/io/file/move.fz
```diff
@@ -39,6 +39,17 @@ private move(
         replace
         tmp
 
+# short-hand for creating and installing effects
+#
+move_custom(f () -> unit) => move_custom default_move_operation f
```
michaellilltokiwa, line 44: is it not possible to also call this `move`?

WaelYoussfi: it is, but i wanted to keep some consistency in naming.

PR #593, opened 2022-11-09
fz: allow passing C compiler and flags to C backend

### LINKED ISSUE #584: add command line option to allow different c compiler
see also #538

maxteufel:
Fixes #584.

### DIFF
--- a/src/dev/flang/be/c/C.java
+++ b/src/dev/flang/be/c/C.java
@@ -538,10 +538,14 @@ public class C extends ANY
     Errors.showAndExit();
 
     var cCompiler = _options._cCompiler != null ? _options._cCompiler : "clang";
-    var command = new List<String>(cCompiler, "-O3");
+    var command = new List<String>(cCompiler);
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
I stumbled over the word `flag`. Personally I would have used args. So I searched for the definitions.
Opinion 1:
> There are no consistent definitions of the terms "option", "argument", and "flag", and there is no central authority within the software development world that could enforce their usage. 

Opinion 2:
> A flag is a type of option, an option of boolean type, and is always false by default (e.g. --verbose, --quiet, --all, --long, etc).
> 
> An option tells the function how to act (e.g. -a, -l, --verbose, --output , -name , -c , etc), whist an arguments tells the function what > to act on/from (e.g. *, file1, hostname, database).


https://unix.stackexchange.com/questions/285575/whats-the-difference-between-a-flag-an-option-and-an-argument

One of those cases where the internet does not have a satisfying answer. :-(

PR #596, opened 2022-11-09
ast/expr: cleanup predicate checking if boxing is needed.

michaellilltokiwa:


fridis:
Good to move this into a separate method. Splitting a boolean expression up in a chain of `if`/`else if` IMHO does not make it more readable, but just more verbose. But this is a question of training...

PR #601, opened 2022-11-10
parser: report none disinct arg names as error

### LINKED ISSUE #170: feature argument name clash does not cause an error
See this little test
```
> echo "say (a 3 4 5); a(b,b,b i32) => b+b+b" | ./build/bin/fz -
15
```
This should complain about `b` being reused.

michaellilltokiwa:
fixes #170


src/dev/flang/ast/AstErrors.java
```diff
@@ -1524,6 +1524,12 @@ static void lossOfPrecision(SourcePosition pos, String _originalString, int _bas
       "Loss of precision for: " + _originalString,
       "Expected number given in base " + _base + " to fit into " + _type + " without loss of precision.");
   }
+
+  public static void argumentNamesNotDistinct(SourcePosition pos)
+  {
+    error(pos,
+      "Names of arguments must be distinct.", null);
```
fridis, line 1531: Be a bit more verbose. I usually add a detailed message and try to list the exact problems, in this case the names of all argument fields. Use `sbn(name)` in the text to get uniform color and quotation. Maybe also add the position of all arguments (by appending pos.show() for each name's position). 
Also, add a suggested fix at the end, e.g., "To solve this, rename argument `x` as `x1`.".

src/dev/flang/parser/Parser.java
```diff
@@ -1101,10 +1101,20 @@ else if (skipDot())
...
+    usedNames.add(name);
     while (skipComma())
       {
-        result.add(name());
+        var pos = pos();
+        var argName = name();
+        result.add(argName);
+        if (!argName.equals("_") && !usedNames.add(argName))
+          {
+            AstErrors.argumentNamesNotDistinct(posObject(pos));
```
fridis, line 1116: argNames is not the best location for this, this will detect `a, a i32` but not `a i32, a i32`. Better check this in `Feature.java`'s constructor that sets `Feature._arguments`.

fridis (commented):
Nice fix.
Some improvement suggestions from me, though...

src/dev/flang/ast/AstErrors.java
michaellilltokiwa: Now it looks like this.
image

Unfortunately the parser does not give us the correct position of the arguments yet so I use the position of the feature.
I added a NYI so this can be fixed in another PR.

src/dev/flang/parser/Parser.java
michaellilltokiwa: I should have realized this...

src/dev/flang/ast/AstErrors.java
fridis: You are using the Features of the arguments now, and they do have a `pos()` entry! 
Let me know if I should pull this PR into the repository (and you add the position in a new PR, or you want to update this.

michaellilltokiwa: yes, but right now all the arguments have the position of the first arg as far as I can tell.

michaellilltokiwa: @fridis Should I create an issue for this? This beeing all the args having the position of the first arg.

fridis:
@michaellilltokiwa, can you resolve the conflict?

michaellilltokiwa:
@fridis I resolved the conflicts

PR #625, opened 2022-11-14
move fzdocs to the main repository

michaellilltokiwa:
Should be merged at the same time as: https://git.tokiwa.software/tokiwa/flang_dev/pulls/104


assets/docs/style.css
```diff
@@ -0,0 +1,136 @@
+.d-grid{
```
fridis, line 1: Provided this was written by you, please add a GPLv3 and Tokiwa-header. Otherwise, where does this come from and what license does it have?

src/dev/flang/tools/docs/Docs.java
```diff
@@ -0,0 +1,346 @@
+/*
+
+This file is part of the Fuzion docs generator implementation.
```
fridis, line 3: Make this line the same as the other Java sources (without 'docs', I guess).

assets/docs/reboot.css
```diff
@@ -0,0 +1,485 @@
+/*!
```
fridis, line 1: Please add a reference (URL) to where this came from. 

Since this is MIT, we have the choice of either 

- adding the MIT license to the Fuzion repository (how is this usually done when there are several licenses, make LICENSE a directory?)
- or changing the license to GPLv3, which MIT should allow us. 

I prefer to keep this under MIT and add that license to our repository. Can you do this, maybe in a separate patch?

src/dev/flang/tools/docs/DocsOptions.java
```diff
@@ -0,0 +1,43 @@
+/*
+
+This file is part of the Fuzion docs generator implementation.
```
fridis, line 3: Make this line the same as the other Java sources (without 'docs', I guess).

src/dev/flang/tools/docs/Html.java
```diff
@@ -0,0 +1,452 @@
+/*
+
+This file is part of the Fuzion docs generator implementation.
```
fridis, line 3: Make this line the same as the other Java sources (without 'docs', I guess).

src/dev/flang/tools/docs/Util.java
```diff
@@ -0,0 +1,131 @@
+/*
+
+This file is part of the Fuzion docs generator implementation.
```
fridis, line 3: Make this line the same as the other Java sources (without 'docs', I guess).

fridis (commented):
Nice to join this into the fuzion repo. 

Just some comment on legal and formal issues...

assets/docs/reboot.css
michaellilltokiwa: @fridis I deleted deleted `reboot.css` now since it was only used when "debugging" the docs and even then it is not super important. Also `flang.dev` has its own `reboot.css` anyway.

Issue #609, opened 2022-11-10
envir.vars effect enhancements


maxteufel:
The `envir.vars` effect should provide features to:

- [ ] get all environment variables and their values as a key-value map.
- [x] set an environment variable.
- [x] unset an environment varable.
- [ ] Go also provides `Clearenv` to unset all environment variables, but I am not sure if we should implement that. If someone *really* wants that, they could iterate over the map provided by the first feature and use the third feature to unset each variable.

michaellilltokiwa:
I think this is fine for now. We can open new issue when we see the need for e.g. a map of env-vars.

PR #628, opened 2022-11-15
lib/envir/vars: implement set and unset

maxteufel:
Addresses #609 

There is no implementation in Java as Java does not provide a possibility to change the environment variables. Even calling to C code does not help with that as apparently the environment is cached when the JVM starts. Subsequent calls to `System.getenv` would not return the updated value.

### DIFF
--- a/lib/envir/vars.fz
+++ b/lib/envir/vars.fz
@@ -47,6 +47,19 @@ is
   #
   get(v string, default string) => (get v).get default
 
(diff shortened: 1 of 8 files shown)


michaellilltokiwa:
Good work.

If you set an env variable, can you then get it? I'm asking because I think if you set an env variable you need to replace the currently installed effect.
Also I would like to suggest adding some basic tests.

PR #631, opened 2022-11-15
adding `append` feature

WaelYoussfi:


WaelYoussfi:
@maxteufel I am well aware of the issues in the code and TBH this is just a prototype that I am working to change completely, hence why it is still WIP.

WaelYoussfi:
`open` will provide a way to open a stream in append mode. Then appending content will be dependent on the opening mode instead of having its own intrinsic.

PR #639, opened 2022-11-17
lib/bitset: fix equality when one of the arguments is empty

### LINKED ISSUE #638: lib/bitset: equality broken
```
ex_bitset is

  x := bitsets.empty
  y := (x.put 134).put 12

  say "x is $x"
  say "y is $y"

  say (x == y)
```
returns
```
x is {}
y is {12, 134}
true
```

This is obviously not what should happen.

maxteufel:
If one of the arguments is the empty bitset, then highest will be nil, and hence the bind operator in the definition of h will make the result nil as well. However, this does not mean that both bitsets are empty. Add the missing check for this case.

Fixes #638.

### DIFF
--- a/lib/bitset.fz
+++ b/lib/bitset.fz
@@ -78,7 +78,7 @@ is
   infix == (other bitset) bool is
      h := highest >>= i -> other.highest >>= j -> i.max j
      match h
-       _ nil => true
+       _ nil => highest!! && other.highest!!
        m u64 =>
          for
            r := true, r && ((has i) <=> other.has i)


lib/bitset.fz
```diff
@@ -78,7 +78,11 @@ is
   infix == (other bitset) bool is
      h := highest >>= i -> other.highest >>= j -> i.max j
      match h
-       _ nil => true
+       _ nil =>
```
michaellilltokiwa, line 81: I think you can omit the if-else and just write `highest!! && other.highest!!`

Issue #641, opened 2022-11-17
Changes required for language server


michaellilltokiwa:
Right now language server uses a branch of fuzion that contains a few hacks to make it work. These should be integrated into main.
These are:
- [x] Make it possible to clear all statically held things in Clazzes.java
- [x] System.exit should not be used. Instead we need something like Fuzion.exit where we can prevent exits.
- [x] Parser.match(): in case of missing matches return match instead of block.
- [x] public access to _effects

michaellilltokiwa:
@tokiwa-software/developers the language server now uses the `main`-branch instead of a `hacks`-branch

PR #643, opened 2022-11-17
air: add Clazzes.reset, a method to clear statically held things

michaellilltokiwa:
see also #641


fridis:
Would be better to not have static fields at all, my failure that I was too lazy when I wrote this. 
Would be good if you could make these instance fields for Clazzes and then have only one single static field Clazzes._instance_ for that instance, so reset() would just allocate a new instance. Then, we could remove that single static field by passing that instance around as an argument (or field of an argument) to all calls that need to access Clazzes.

PR #644, opened 2022-11-17
Errors: add ability to prevent System.exit

michaellilltokiwa:
see also: #641

### DIFF
--- a/src/dev/flang/tools/Tool.java
+++ b/src/dev/flang/tools/Tool.java
@@ -48,6 +48,7 @@ import dev.flang.opt.Optimizer;
 
 import dev.flang.util.ANY;
 import dev.flang.util.Errors;
+import dev.flang.util.FatalError;
 import dev.flang.util.List;
 import dev.flang.util.Profiler;
 
@@ -154,6 +155,10 @@ public abstract class Tool extends ANY
       {
         parseArgs(_args).run();
         Errors.showAndExit(true);
+      }
+    catch (FatalError e)
+      {
+        System.exit(e.getStatus());
       }
     catch(Throwable e)
       {
--- a/src/dev/flang/tools/Tool.java
+++ b/src/dev/flang/tools/Tool.java
@@ -26,6 +26,8 @@ Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
 
 package dev.flang.tools;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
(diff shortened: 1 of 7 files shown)


src/dev/flang/tools/Tool.java
```diff
@@ -155,8 +157,10 @@ protected void run()
       }
     catch(Throwable e)
       {
-        e.printStackTrace();
-        System.exit(1);
+        var sw = new StringWriter();
+        var pw = new PrintWriter(sw);
+        e.printStackTrace(pw);
+        Errors.fatal(sw.toString());
```
fridis, line 163: Would be nice to move this logic into a new method `Errors.fatal(Throwable)`.

src/dev/flang/util/Errors.java
```diff
@@ -349,7 +353,19 @@ public static void fatal(String s, String detail)
...
+
+  /**
+   * Call System.exit in case preventExit is false.
+   */
+  private static void exit(int status)
+  {
+    if (!preventExit)
+      {
+        System.exit(status);
+      }
```
fridis, line 368: Just returning from a `Error.fatal` is IMHO not a good idea, the code that calls `Error.fatal` can safely assume this not to return. If this returns anyway, we might run into more errors or endless loops, endless recursions, etc. 

The only alternative I see is to declare a new class FatalError that extends Error and throw this. Then, `Tool.run` or the corresponding routine in the language server can catch `FatalError` and handle it. 

Maybe we could even get rid of the `preventExit` flag and always terminate via `FatalError`, maybe `FatalError` could contain the `status` code as well, so `Tool.run` could then catch it and `System.exit` with this status. 

Futhermore, it might be good to check for repeated calls to `Errors.exit` and perform a `System.exit` in this case (just in case some code catches `Error` and runs into a loop peroducing fatal errors.

fridis (commented):
A bit worried about returning from Errors.fatal, see suggestion below.

src/dev/flang/tools/fzjava/FZJava.java
```diff
@@ -116,7 +117,14 @@ public class FZJava extends Tool
...
   {
-    new FZJava(args).run();
+    try
+      {
+        new FZJava(args).run();
+      }
+    catch (FatalError e)
+      {
+        System.exit(e.getStatus());
+      }
```
fridis, line 127: non needed if this is moved to  `Tool.run`.

PR #657, opened 2022-11-21
lib/closeable: new effect to handle opening and closing of files

maxteufel:
Eventually, the idea behind this effect should be generalized to other types of resources that can be opened and closed, e.g. network connections.

### DIFF
--- a/lib/closeable.fz
+++ b/lib/closeable.fz
@@ -27,56 +27,45 @@
 # file handler (currently an integer representing a file descriptor) in
 # mutable state.
 #
+Resource ref : hasHash Resource is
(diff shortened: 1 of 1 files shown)


lib/closeable.fz
```diff
@@ -0,0 +1,74 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature closeable
+#
+# -----------------------------------------------------------------------
+
+# closeable -- effect for handling a resource that can be opened an closed
+#
+# closeable provides an operation 'open' that opens and file and stores a
```
michaellilltokiwa, line 26: that opens a file

PR #658, opened 2022-11-21
fix windows test

michaellilltokiwa:


michaellilltokiwa:
works on my machine, but does not yet work with github actions. not sure why...
image

PR #660, opened 2022-11-21
flake.nix: init

maxteufel:
This provides:

- a Fuzion package for Nix, if Nix is installed a shell with the `fz` and `fzjava` executables can be launched with `nix shell github:tokiwa-software/fuzion` (once merged)
- if Nix is installed and one uses `nix develop`, all necessary dependencies will be installed and `PATH` and `MAKEFLAGS` will be configured automatically.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,42 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1659877975,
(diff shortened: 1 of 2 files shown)


fridis:
Can you add GPLv3 and Tokiwa Software headers to the files and include a description what these files are and how they are used. Try to reduce the confusion for those not using nix as much as possible. 
Also, would it be possible to move these files to a sub-directory?

fridis:
> Do you have any name in mind?

What do other repositories use? `nix`, `.nix`, `nixOS` or `flake` or something similar might make sense.

PR #682, opened 2022-11-24
Sequence and friends: improve documentation

maxteufel:
Add some clarifications to some comments, add missing comments, change list to Sequence in comments where the latter was actually meant, add a warning about the mutability of stream.


lib/stream.fz
```diff
@@ -65,13 +80,10 @@ stream(redef T type) ref : Sequence T is
...
 
-  # create a stream of T.
-  #
-  # A stream contains mutable state, so it cannot be reused or shared
-  # between threads.
+  # Return this stream as a stream.
   #
-  # Default implementation uses asList.  Heirs must redefine at least
-  # one of asList or asStream.
+  # This is a helper function that needs to be defined because stream is a heir
```
michaellilltokiwa, line 85: I think it's `an heir`

michaellilltokiwa (approved):

PR #700, opened 2022-11-30
lib: implement trees of bounded balance

maxteufel:
This implements trees of bounded balance as described and implemented in the paper at <https://groups.csail.mit.edu/mac/users/adams/BB/>. This type of tree is also implemented in Haskell's [containers](https://hackage.haskell.org/package/containers) library; tree operations have the complexity as usual on self-balancing trees.

This can be in particular used to implement sets and maps of types that have an order. The current implementations of sets and maps in Fuzion either suffer from using mutable state or from having a higher complexity.

Not yet implemented:

- [x] `hedge_union` as an optimized replacement for `union` (also see the paper by Adams)
- [ ] rename some features to match the Fuzion `Set` feature (delete vs. remove)


michaellilltokiwa:
Cool data structure. What are the space/time, worst/avg/best complexities?

lib/TreeNode.fz
```diff
@@ -0,0 +1,30 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature TreeNode
+#
+# -----------------------------------------------------------------------
+
+TreeNode(A, B, C type) ref is
+  datum A is abstract
+  count u32 is abstract
```
michaellilltokiwa, line 26: maybe u64 even though a count higher than 2^32 seems unlikely

michaellilltokiwa:
> However, unlike orderedMap, which has to copy an array when adding a new element, we can share the nodes that remain unchanged.

The downside is worse cache locality though which seems to be very important for best performance on modern hardware.

> For union, intersection, and difference see https://arxiv.org/abs/1602.02120v3; note that it is possible parallelize these features. 

Parallelize by _divide and conquer_, I assume?

lib/tree.fz
```diff
@@ -0,0 +1,418 @@
...
+  # The gamma constant is used to determine whether it is necessary to do a single or
+  # a double rotation to resolve the imbalance.
+  #
+  # According to the paper
+  #
+  #   Hirai, Y. and Yamamoto, K. (2011): Balancing weight-balanced trees. In: Journal of
+  #   Functional Programming, 21(3):287, DOI: https://dx.doi.org/10.1017/S0956796811000104
+  #
+  # the only integer solution is <3, 2>. Note that by the convention of that paper, we use
+  # the original WBT algorithm.
```
fridis, line 104: I do not understand the sentence "Note that by the convention of that paper..."

fridis (commented):
did not look into details yet, but at least some first comments...

PR #701, opened 2022-11-30
be/c: enable -Wall -Werror by default

michaellilltokiwa:
some warnings are excluded though


fridis:
I would be happy to pull this in, but I would remove `-pedantic`.

PR #704, opened 2022-11-30
be/c: use c11 threads instead of pthreads

michaellilltokiwa:
generated code:
```c
  thrd_t pt[1];
  int res;
  struct fzThrd_startRoutineArg* arg;
  arg = malloc(sizeof(struct fzThrd_startRoutineArg));
  if (arg==NULL)
  {
    fprintf(stderr,"*** malloc(%zu) failed\012",sizeof(struct fzThrd_startRoutineArg));
    exit(1);
  }
  arg->fzThrd_startRoutineArgFun = (void *)&fzC__L1357fuzion__sy__all__call;
  arg->fzThrd_startRoutineArgArg = (void *)arg0;
  res = thrd_create(&pt[0],(thrd_start_t)fzThrd_startRoutine,arg);
  if (res==thrd_error)
  {
    fprintf(stderr,"*** thrd_create failed with return code %d\012",res);
    exit(1);
  }
```

### DIFF
--- a/src/dev/flang/be/c/C.java
+++ b/src/dev/flang/be/c/C.java
@@ -551,8 +551,8 @@ public class C extends ANY
       {
         command.addAll("-lgc");
       }
-    // NYI link libmath, libpthread only when needed
-    command.addAll("-lm", "-lpthread", "-o", name, cname);
+    // NYI link libmath only when needed
+    command.addAll("-lm", "-o", name, cname);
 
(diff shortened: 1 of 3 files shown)


michaellilltokiwa:
@tokiwa-software/developers I want your opinion here. Do we want to keep pthreads use c11 threads or have an option to choose either? If so what should be the default?

michaellilltokiwa:
c11 threads are not widely used/supported, sticking with posix threads for now.

PR #707, opened 2022-11-30
be/c,dfa: evaluate pre-condition of main feature, #270

### LINKED ISSUE #270: pre-Condition in c-backend not always evaluated
This example runs fine when using c-backend but should fail:
```
ex
  pre false
  =>
  say "hello"
```

the same problem occurs when applying this patch:

```patch
diff --git a/lib/stream.fz b/lib/stream.fz
index 28b11e67..ef74b2bd 100644
--- a/lib/stream.fz
+++ b/lib/stream.fz
@@ -39,7 +39,10 @@ stream(redef T type) ref : Sequence<T> is
         ref : Cons<T,list<T>>
           memoizedTail option<some<list<T>>> := nil
           head => h
-          tail =>
+          tail
+            pre false
+            =>
+            say "tail"
             if !memoizedTail.exists
               set memoizedTail := some fromStream
             memoizedTail.get.val
```
example test-code for the diff above:
```
ex
  =>
  say [1,2,3,4,5,6,7].asStream.take 2
```

michaellilltokiwa:
fixes #270


tests/reg_issue270/issue270.fz
```diff
@@ -0,0 +1,4 @@
+ex
+  pre false
```
fridis, line 2: The idea of a negative test is that it contains numbered lines with comment `# 1. should flag an error: ...this or that should be detected...` and then negative.mk checks that they all are actually reported.

fridis (commented):
Looks good, just a comment on the regression test.

tests/reg_issue270/issue270.fz
michaellilltokiwa: @fridis I did not add this since the precondition failure occurs at run time.
I added a `should flag an error` comment now and a `skip_c` file. The minor downside to is that now it does not check anymore if the compiled program also terminates erroneously.

fridis:
@michaellilltokiwa, can you check why the checks failed?

michaellilltokiwa:
@fridis fixed

PR #712, opened 2022-12-01
lib/unit: provide a unit.type.monoid feature

### LINKED ISSUE #686: lib: introduce a unit monoid?
Should we introduce this unit monoid:
```
unit_monoid : Monoid unit is
  redef infix ∙ (a, b unit) unit is unit
  redef infix == (a, b unit) bool is true
  redef e unit is unit
```
This would for example allow to define the stream/list forAll feature in a functional way:
```
(a.map f).fold unit_monoid
```

maxteufel:
This feature implements a Monoid of unit that essentially does not do anything, however it can be used when a `Monoid unit` type is expected, such as in some calls to `fold`.

Fixes #686.

### DIFF
--- a/lib/unit.fz
+++ b/lib/unit.fz
@@ -92,3 +92,8 @@ unit is
 
   # human readable string
   redef asString => "unit"
+
+  type.monoid : Monoid unit is
+    redef infix ∙ (a, b unit) unit is unit
+    redef infix == (a, b unit) bool is true
+    redef e unit is unit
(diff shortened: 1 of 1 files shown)


michaellilltokiwa (approved):

lib/unit.fz
```diff
@@ -92,3 +92,8 @@ unit is
 
   # human readable string
   redef asString => "unit"
+
+  type.monoid : Monoid unit is
```
fridis, line 96: please add an explanation what this does and what it is good for. Even though this is inherently simple, it is not at all obvious for a newbie...

Issue #697, opened 2022-11-28
base library: string manipulation


michaellilltokiwa:
There are some basic string manipulation functions missing in our std lib:

From Basic (https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/keywords/string-manipulation-summary):
- [x] Compare two strings
- [ ] Convert strings
- [ ] Reverse a string
- [x] Convert to lowercase or uppercase
- [x] Create a string of repeating characters
- [x] Find the length of a string
- [ ] Format a string
- [ ] Manipulate strings
- [ ] Set string comparison rules
- [ ] Work with ASCII and ANSI values
- [ ] Replace a specified substring
- [ ] Return a filter-based string array
- [ ] Return a specified number of substrings

From golang (https://pkg.go.dev/strings):
- [x] Clone
- [x] Compare
- [x] Contains
- [ ] ContainsAny
- [ ] ContainsRune
- [x] Count
- [x] Cut
- [ ] EqualFold
- [ ] Fields
- [x] FieldsFunc
- [x] HasPrefix
- [x] HasSuffix
- [x] Index
- [ ] IndexAny
- [ ] IndexByte
- [ ] IndexFunc
- [ ] IndexRune
- [x] Join
- [x] LastIndex
- [ ] LastIndexAny
- [ ] LastIndexByte
- [ ] LastIndexFunc
- [ ] Map
- [x] Repeat
- [ ] Replace
- [x] ReplaceAll
- [x] Split
- [x] SplitAfter
- [x] SplitAfterN
- [x] SplitN
- [ ] Title
- [x] ToLower
- [ ] ToLowerSpecial
- [x] ToTitle
- [ ] ToTitleSpecial
- [x] ToUpper
- [ ] ToUpperSpecial
- [x] ToValidUTF8
- [x] Trim
- [ ] TrimFunc
- [ ] TrimLeft
- [ ] TrimLeftFunc
- [ ] TrimPrefix
- [ ] TrimRight
- [ ] TrimRightFunc
- [ ] TrimSpace
- [ ] TrimSuffix

michaellilltokiwa:
For some we might want two flavors, e.g. `join`
- `strings.join "string 1" "string 2"`
- `"my string".join "another string"`

michaellilltokiwa:
most important things are implemented. closing this. if need arises we can open more specific issues.

PR #724, opened 2022-12-05
lib/string: add cut, fields_func, and strings.join features

maxteufel:
All three features are inspired by the Go standard library [1] [2] [3].

[1]: https://pkg.go.dev/strings#Cut
[2]: https://pkg.go.dev/strings#FieldsFunc
[3]: https://pkg.go.dev/strings#Join

Reference: #697

### DIFF
--- a/lib/string.fz
+++ b/lib/string.fz
@@ -372,7 +372,7 @@ string ref : has_equality, hasHash string, ordered string, strings is
           fuzion.std.panic "string.codepointsAndErrors: missing case for $b1"
 
 
-  # create substring of this string consisting of codepoints from (inclusive) .. to (exclusive).
+  # create substring of this string consisting of bytes from (inclusive) .. to (exclusive).
   #
   substring(from, to i32) ref : string
     pre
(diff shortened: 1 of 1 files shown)


lib/string.fz
```diff
@@ -518,6 +518,50 @@ string ref : has_equality, hasHash string, ordered string, strings is
       (s * missing) + string.this
 
 
+  # Takes a predicate p that takes a codepoint and returns true or false. Splits
```
michaellilltokiwa, line 521: I suggest dropping the first sentence of the comment as it does not add any valuable information for understanding the purpose of the feature. 

I think the special handling of multiple, neighboring codepoints should be possible to turn off. I.e. if we have a CSV with lots of empty fields.

Would it make sense for `p` -  instead of taking a single codepoint - to take a sequence of all of the following codepoints?

lib/string.fz
```diff
@@ -518,6 +518,50 @@ string ref : has_equality, hasHash string, ordered string, strings is
...
+          match r.in_run
+            TRUE => state_wrapper r.l false r.current_pos (r.current_pos + 1)
+            FALSE => state_wrapper r.l false r.start_pos (r.current_pos + 1))
+
+    match last_state.in_run
+      TRUE => last_state.l
+      FALSE => last_state.l ++ [(substring last_state.start_pos)]
+
+
+  # Cuts out the first appearance of the string sep from this string, that is returns a
```
michaellilltokiwa, line 546: that is returns

lib/string.fz
```diff
@@ -548,6 +592,19 @@ strings is
       utf8 Sequence u8 is a.utf8 ++ b.utf8
 
 
+  # Takes a Sequence of string and concatenates its elements, while adding the separator
```
michaellilltokiwa, line 595: sequence of strings

I think I prefer not using _code language_ but that is certainly a matter of taste.

lib/string.fz
```diff
@@ -518,6 +518,49 @@ string ref : has_equality, hasHash string, ordered string, strings is
       (s * missing) + string.this
 
 
+  # Splits this string at codepoints where p is true and returns the result as a
+  # list of strings. In case multiple, neighboring codepoints in the string are
+  # evaluated to be true by p, this does not cause empty strings to be added to
+  # the result list, rather this case is being treated as being one big separator.
+  fields_func(p codepoint -> bool) list string is
```
fridis, line 525: can you add an empty comment line here

lib/string.fz
```diff
@@ -518,6 +518,49 @@ string ref : has_equality, hasHash string, ordered string, strings is
...
+
+
+  # Cuts out the first appearance of the string sep from this string, in other words,
+  # returns a tuple of two strings and a bool, the first string is the substring before
+  # the first appreance of sep, the second string is the substring after the first
+  # appearance of sep. The bool result is true iff sep appears in this string.
+  #
+  # If sep does not appear in this string at all, return this string as the first string,
+  # the empty string as the second, and false as the bool.
+  cut(sep string) tuple string string bool is
```
fridis, line 552: can you add an empty comment line here

fridis (commented):
Looks good, but I found one bug that should lead to a compilation error and have a suggestion to avoid this and simplify the code.

PR #729, opened 2022-12-06
lib/string: add split_n, split_after, split_after_n features

maxteufel:
These features are similar to the eponymous ones found in the Go standard library [1]. They are however, not fully identical.

[1]: https://pkg.go.dev/strings

Reference: #697

### DIFF
--- a/lib/String.fz
+++ b/lib/String.fz
@@ -525,13 +525,75 @@ String ref : has_equality, hasHash String, ordered String is
     pre
       !s.is_empty
     is
-    match (find s)
-      nil => [as_string].as_list
-      idx i32 => ref : Cons String (list String)
-                   head => substring 0 idx
-                   tail =>
-                     rest := substring (idx + s.byte_length) String.this.byte_length
-                     rest.split s
+      split0 s nil false
+
+
+  # split string after s, that is do the same thing as split but
+  # include the separator s in the resulting strings
+  #
+  split_after(s String) list String
+    pre
+      !s.is_empty
+    is
+      split0 s nil true
+
+
+  # split string at s, for at most n occurences of s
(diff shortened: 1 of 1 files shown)


lib/string.fz
```diff
@@ -454,13 +454,63 @@ string ref : has_equality, hasHash string, ordered string, strings is
     pre
```
michaellilltokiwa, line 454: This does only tangentially belong to this PR but I'm wondering if this precondition !s.isEmpty what you would expect when using a split feature? I would expect to get an empty list in return in such cases.

lib/string.fz
```diff
@@ -454,13 +454,63 @@ string ref : has_equality, hasHash string, ordered string, strings is
...
+  private split0(s string, n i32, split_after bool) list string
+    pre
+      !s.isEmpty,
+      n >= 0 : n > 0,
+      n < 0 : n ≟ -1
+    is
+      unlimited := n < 0
+      match (find s)
+        nil => [string.this].asList
+        idx i32 => ref : Cons string (list string)
```
michaellilltokiwa, line 503: I prefer less indentation. Can you put `ref : Cons` on the next line?

lib/string.fz
michaellilltokiwa: Oh, yes. My bad, sorry. :-)

fridis:
Looks good. 
Can you use consistent naming (`n` vs.  `limit`) and add comments to the argument fields to explain their meaning?
Splitting 'after each utf8-sequence' (codepoint?) is IMHO not useful, would destroy emojis.

fridis:
@maxteufel, can you check why the checks failed?

PR #732, opened 2022-12-06
lib/string: add find_last feature

maxteufel:


michaellilltokiwa (approved):

lib/string.fz
```diff
@@ -397,6 +397,17 @@ string ref : has_equality, hasHash string, ordered string, strings is
     (searchableSequence utf8).find substring.utf8
 
 
+  # find (utf8-byte-) index of last occurrence of 'substring'
+  # within this string.
+  find_last(substring string) option i32 is
+    match find substring
+      nil => nil
+      idx i32 =>
+        match (strings.fromBytes (utf8.drop (idx + substring.byteLength))).find_last substring
```
michaellilltokiwa, line 406: @fridis Is code like this detected to be tail recursive?

fridis: yes, should be!

fridis: One caveat though: If there are heirs of string that redefine `find_last`, and DFA thinks that they could be called here, this is no longer the case.

fridis: Nope, this is not tail recursive since the result of the call is not returned directly. To make this tail recursive, we need code like this:

```
  # find (utf8-byte-) index of last occurrence of 'substring'
  # within this string.
  find_last(substring string) option i32 is
    find_last string nil

  # find (utf8-byte-) index of last occurrence of 'substring'
  # within this string.
  private find_last(substring string, found option i32) option i32 is
    match find substring
      nil => found
      idx i32 => (substring idx+1).find_last substring (found.orElse idx)
```

I also added a call to `substring` instead of the utf8-bytes-twiddling.

fridis: Looks correct, but  a bit over-complex, so go ahead!

I think that indices in substring should better be utf8-byte indices, then my version using substring would do. 

In another PR, you can check all indices used in string and either make them utf8-byte indices or rename the feature to make clear that they work on codepoints, composite glyphs (whatever the best name is for those), utf16 or whatever else we might want to support.

PR #734, opened 2022-12-06
lib/string: changed utf8 to return a list, was sequence

michaellilltokiwa:
to be considered...

### DIFF
--- a/lib/codepoint.fz
+++ b/lib/codepoint.fz
@@ -34,17 +34,17 @@ is
   # the utf8 encoded bytes for the string representation
   # of this codepoint
   #
-  redef utf8 Sequence u8 is
-    if      utf8EncodedInOneByte   .contains val then [ val.low8bits ]
+  redef utf8 list u8 is
+    if      utf8EncodedInOneByte   .contains val then [ val.low8bits ].asList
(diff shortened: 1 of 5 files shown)


fridis:
My first instinctive reaction to this was the following paragraph

> `Sequence` is the more generic type that can have features more powerful than a `list`, e.g., it could provide predicates like `hasEfficientIndexedAccess` or `asUnderlyingArray` or features like `asArray` that are O(1) if there is an underlying array. For a `list` of size `n`, `asArray` is always `O(n)`.

But thinking of it further, `list` could actually provide this functionality as well if `Cons` would provide corresponding features `hasEfficientIndexedAccess`  etc. and they were redefined by the `Cons` instance created, e.g., by `array.asList`. 

This is a fundamental API decision: What should the default type for a ordered set of values of the same type be? Maybe a base library feature receiving an argument should receive a `Sequence` since `list` can be assigned to `Sequence`, while a feature's result should be `list` for the same reason. Or we should use the same throughout, but which one?

michaellilltokiwa:
Hmm, it seems a bit weird to me that a cons cell is supposed to answer the question if it is part of an array.

michaellilltokiwa:
Loosely related is the discussion if we can/should get rid of streams. To Go people recently revisited this problem as well:
https://github.com/golang/go/discussions/54245

PR #758, opened 2022-12-08
lib/list: add prepend_to_all and intersperse features

maxteufel:
`prepend_to_all` adds an element `sep` in front of every element of a list, `intersperse` adds an element `sep` only in between elements of a list.

### DIFF
--- a/lib/list.fz
+++ b/lib/list.fz
@@ -250,6 +250,14 @@ list(A type) : choice nil (Cons A (list A)), Sequence A is
       x A => tail.prepend_to_all sep (res ++ [sep, x])
 
 
+  # add an element sep between every element of this list.
+  #
+  intersperse(sep A) list A is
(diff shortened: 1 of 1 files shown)


lib/list.fz
```diff
@@ -231,6 +231,34 @@ list(A type) : choice nil (Cons A (list A)), Sequence A is
...
+  #
+  prepend_to_all(sep A) list A is
+    prepend_to_all0 sep nil
+
+
+  # add an element sep in front of every element of this list, helper to
+  # allow tail recursion.
+  #
+  # if this list is the empty list, return the given result list res, if
+  # not, make a recursively call this on the tail of this list, feeding it
```
fridis, line 244: _if not_ -> _otherwise_
I cannot parse _make a recursively call this on the tail of this list_

PR #785, opened 2022-12-13
docs: link to redefined features

maxteufel:
And drop redundant copy of comments of redefined features.

### DIFF
--- a/src/dev/flang/tools/docs/Html.java
+++ b/src/dev/flang/tools/docs/Html.java
@@ -160,6 +160,38 @@ public class Html
       + "</div>";
   }
 
+  /**
+   * list of features that are redefined by feature af
+   * @param af
+   * @return list of redefined features, as HTML
+   */
+  private String redefines(AbstractFeature af, boolean recurse)
+  {
+    if (af.redefines().isEmpty())
+      {
+        return "";
+      }
+    else
+      {
+        var links = af
+          .redefines()
+          .stream()
+          .map(f -> """
+            <li><a href="$1">$2</a></li>$3
+          """.replace("$1", featureAbsoluteURL(f)).replace("$2", f.qualifiedName()).replace("$3", redefines(f, true)))
+          .collect(Collectors.joining(System.lineSeparator()));
+
+        if (recurse)
+          {
+            return links;
+          }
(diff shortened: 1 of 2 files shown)


src/dev/flang/tools/docs/Html.java
```diff
@@ -137,6 +137,38 @@ private String summary(AbstractFeature af)
       + "</div>";
   }
 
+  /**
+   * list of features that are redefined by feature af
+   * @param af
+   * @return list of redefined features, as HTML
+   */
+  private String redefines(AbstractFeature af, boolean recurse)
```
michaellilltokiwa, line 145: I had trouble understanding the `recurse` option. If you split this up into a recursive part and one that wraps the result in the `div>ul` then it should be easier to read. I also think the `af.redefines().isEmpty()` should not be necessary?

src/dev/flang/tools/docs/Html.java
```diff
@@ -138,6 +138,34 @@ private String summary(AbstractFeature af)
...
+    if (af.redefines().isEmpty())
+      {
+        return "";
+      }
+
+    return "<div class='fd-redefines'><br />redefines: <br /><ul>" + redefines0(af) + "</ul><br /></div>";
+  }
+
+
+  private String redefines0(AbstractFeature af)
```
fridis, line 157: Can you add a comment, in particular explaining  —if correct— that this is a helper for `redefines` and what this does.

src/dev/flang/tools/docs/Html.java
```diff
@@ -138,6 +138,34 @@ private String summary(AbstractFeature af)
...
+  /**
+   * list of features that are redefined by feature af
+   * @param af
+   * @return list of redefined features, as HTML
+   */
+  private String redefines(AbstractFeature af)
+  {
+    if (af.redefines().isEmpty())
+      {
+        return "";
```
fridis, line 150: try to avoid having multiple `return` statements, makes the code harder to understand and debug.

fridis (commented):
just my nitpicking...

fridis:
@maxteufel, can you resolve the conflicts?

PR #789, opened 2022-12-13
lib/list: add init feature

maxteufel:


fridis:
@maxteufel, please resolve the conflicts.

PR #799, opened 2022-12-15
lib: io: remove exists effect and intrinsics

maxteufel:
Superseded by the stats/lstats intrinsics and the stat effect.

### DIFF
--- a/tests/fileio/fileiotests.fz
+++ b/tests/fileio/fileiotests.fz
@@ -36,17 +36,22 @@ fileiotests =>
   writer := f.write
   reader := f.read
 
-  say "$dir exists: {(f.exists dir).val}"
+  exists(path string) bool is
+    match f.stat path true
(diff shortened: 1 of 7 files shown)


tests/fileio/fileiotests.fz
```diff
@@ -36,17 +36,22 @@ fileiotests =>
   writer := f.write
   reader := f.read
 
-  say "$dir exists: {(f.exists dir).val}"
+  exists(path string) bool is
```
michaellilltokiwa, line 39: I think we want this in the stdlib.

michaellilltokiwa: - I'd say separate. The feature then uses the stat effect but that is not really a problem, is it?
- Good points, if the feature returns a bool then it should always return true unless really nothing exists but maybe we should rather not return a bool but a more complex answer that includes your mentioned cases.

PR #800, opened 2022-12-15
lib: io: implement get_file_size as a fuzion feature

maxteufel:
Previously, get_file_size was implemented as an intrinsic which, in the C backend, called to stat. So does the stats intrinsic, therefore we can actually move to using that.

Also mark get_file_size private for now. It uses the stats intrinsic, but without using the stat effect. We might want to change this. Keep it private until a decision is made.

### DIFF
--- a/lib/fuzion/sys/fileio.fz
+++ b/lib/fuzion/sys/fileio.fz
@@ -59,23 +59,13 @@ public fileio is
 
   # retrieves the file size in bytes and returns an outcome of error in case of an error
   #
-  public get_file_size(
-                       # the (relative or absolute) file name, using platform specific path separators
-                       path String) outcome i64 is
-    arr := fuzion.sys.c_string path
-    size := get_file_size arr unit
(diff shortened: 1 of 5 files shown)


lib/fuzion/std/fileio.fz
```diff
@@ -62,23 +62,15 @@ public fileio is
...
-                       path string) outcome i64 is
-    arr := fuzion.sys.c_string path
-    size := get_file_size arr unit
-    if size != -1
-      size
-    else
-      error "an error occurred while retrieving the size of the file/dir: \"$path\""
-
-  # intrinsic that returns the file size in bytes or -1 in case of an error
+  # NYI: should this use effects?
```
michaellilltokiwa, line 65: I think, yes it should.

fridis: what is in `fuzion.std.` should be private helpers used by public effects, so this should not use an effect.

PR #801, opened 2022-12-15
list, Sequence: clarify behaviour of first, last and forceTail when empty

maxteufel:
See also: <https://github.com/tokiwa-software/fuzion/pull/789#discussion_r1049523594>

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -72,7 +72,9 @@ Sequence(T type) ref is
   count => (mapSequence i32 (_ -> 1)).fold i32.sum
 
 
-  # get the first element of this Sequence, panic if Sequence is empty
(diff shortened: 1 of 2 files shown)


lib/Sequence.fz
```diff
@@ -72,7 +72,7 @@ Sequence(T type) ref is
   count => (mapSequence i32 (_ -> 1)).fold i32.sum
 
 
-  # get the first element of this Sequence, panic if Sequence is empty
+  # get the first element of this Sequence, undefined behaviour if Sequence is empty
```
fridis, line 75: Not respecting a pre-condition is a bug in the caller. 

I think it is better to say: _Sequence must not be empty, causes precondition failure if debug is enabled._

PR #803, opened 2022-12-15
lib: io: improve error handling for stat/lstat

maxteufel:
For the C backend, return the error number directly in case of a failure. For the interpreter backend, match the possible exceptions returned by Files.readAttributes to similar C error numbers.

### DIFF
--- a/lib/io/file/stat.fz
+++ b/lib/io/file/stat.fz
@@ -57,7 +57,7 @@ private stat(
       meta_data data[0] data[1] (data[2] == 1) (data[3] == 1)
     else
       replace
-      error "error while retrieving meta_data for the following file/dir: $path"
+      error "stat error {data[0]} for {path}"
 
   # returns stats of the file/dir passed in the pathname
   # in success it will return a meta_data outcome storing stats regarding the file/dir
@@ -73,7 +73,7 @@ private stat(
       meta_data data[0] data[1] (data[2] == 1) (data[3] == 1)
     else
       replace
-      error "error while retrieving meta_data for the following file/dir: $path"
+      error "lstat error {data[0]} for {path}"
 
 # short-hand for installing the default stat effect to the provided code
 #
--- a/src/dev/flang/be/interpreter/Intrinsics.java
+++ b/src/dev/flang/be/interpreter/Intrinsics.java
@@ -75,6 +75,26 @@ public class Intrinsics extends ANY
   }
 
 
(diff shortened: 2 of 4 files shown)


src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -438,8 +438,28 @@ private static synchronized void releaseFileDescriptor(long fileDescriptor)
               stats[3] = metadata.isDirectory()? 1:0;
               return new boolValue(true);
             }
-          catch (Exception e)
+          catch (UnsupportedOperationException e)
+            {
+              stats[0] = 95; // ENOTSUP 95 Operation not supported
```
fridis, line 443: Can you add a class or enum that defines named constants for errno? As a general rule, any number except 0 and maybe 1 should usually be declared as a named constant.

lib/io/file/stat.fz
```diff
@@ -67,7 +67,7 @@ private stat(
       meta_data data[0] data[1] (data[2] == 1) (data[3] == 1)
     else
       replace
-      error "error while retrieving meta_data for the following file/dir: $path"
+      error "error number: {data[0]}"
```
fridis, line 70: Don't you think it helps to include `$path` as well?

fridis (commented):
Looks good. What worries me is platform independence. What do the errno values look like on Windows, QNX, MacOS, etc.?

lib/io/file/stat.fz
fridis: sure, the caller will know this. But the caller will perform lazy error handling and propagate an error further up to the caller's caller who does not know :-)

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -75,6 +75,22 @@ public class Intrinsics extends ANY
   }
 
 
+  /*------------------------------  enums  ------------------------------*/
+
+
+  enum IntrinsicError
```
fridis, line 81: Please add a comment here. 

`IntrinsicError` is maybe not the best name, there is nothing wrong with the intrinsic mechanism here. I think these are declared in _errno.h_, so `ErrNo` or `SystemErrNo` would maybe be better names.

In a second patch, I think we will need a counterpart to this in _fz/intrinsic_error.fz_ or similar to map these back to human-readable error messages.

Issue #637, opened 2022-11-16
type.equality features and generic types


maxteufel:
I made the following change:
```diff
diff --git i/lib/complex.fz w/lib/complex.fz
index af92d163..41633659 100644
--- i/lib/complex.fz
+++ w/lib/complex.fz
@@ -61,6 +61,9 @@ is
   redef infix == (b complex C) => a.real  = b.real && a.imag  = b.imag
   redef infix != (b complex C) => a.real != b.real || a.imag != b.imag
 
+  complex.type.equality(C (numeric C).type, a, b complex C) =>
+    a.real = b.real && a.imag = b.imag
+
   # NYI: total order ignoring imag
   redef infix <  (b complex C) => a.real <  b.real
   redef infix <= (b complex C) => a.real <= b.real
```
I have the following example:
```
ex is
  a := complex 42 17
  b := complex 40 7
  c := complex 2 10

  say a
  say b
  say c
  say (b + c)

  say (complex.type.equality a (b + c))
  say (a ≟ (b + c))
  say (infix ≟ (complex i32) (b + c) a)
```
I get these errors:
```
ex.fz:11:8: error 1: Wrong number of generic arguments
  say (complex.type.equality a (b + c))
-------^
Wrong number of actual generic arguments in type:
Type: complex
expected one generic argument for 'C'
found none.


$FUZION/lib/complex.fz:31:1: error 2: Used abstract feature equality is not implemented
complex (C (numeric C).type,
^
Feature 'complex.#type_STATIC' instantiated at $FUZION/lib/equals.fz:29:27:
equals(T type, a, b T) => T.equality a b
--------------------------^
inherits or declares abstract feature 'Type_STATIC.equality' declared at $FUZION/lib/Type.fz:69:3:
  equality(a, b STATIC_THIS_TYPE) bool is abstract
--^
which is called at $FUZION/lib/equals.fz:29:29:
equals(T type, a, b T) => T.equality a b
----------------------------^
without providing an implementation
```

How am I supposed to make this work?

michaellilltokiwa:
meanwhile fixed

PR #819, opened 2022-12-20
lib/searchablelist: add a variant of replace with a limit

maxteufel:
Given a limit, replace will only replace the first n occurrences of the sub-list to be replaced. Also, we add the same variants to the replace features of searchableSequence and string.

### DIFF
--- a/lib/searchablelist.fz
+++ b/lib/searchablelist.fz
@@ -72,21 +72,31 @@ is
         c1 Cons => ((searchablelist c1.tail).find l) >>= n -> n + 1
 
 
-  # replace all occurances of f by r
+  # replace all occurrences of f by r
   #
-  replace (f, r list A) =>
+  replace (f, r list A) => replace f r nil from nil
 
-    # tail recursive helper
-    #
-    repl (h,       # the head part of from with f already replaced by r
-          t list A # the tail that still needs to be searched for f
-         ) list A
-    is
-      match (searchablelist t).find f
-        nil   => h ++ t
-        n i32 => repl h++(t.take n)++r (t.drop n+f.count)
 
-    repl nil from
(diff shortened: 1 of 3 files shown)


lib/searchablelist.fz
```diff
@@ -72,21 +72,31 @@ is
...
+
+
+  # replace the first n occurrences of f by r
+  #
+  replace (f, r list A, n u64) => replace f r nil from n
+
+
+  # tail recursive helper for the replace features
+  #
+  private replace (f, r,
```
michaellilltokiwa, line 87: I took a while for me to guess that f is find and r replace. A comment might help. I would actually prefer full words even but I think @fridis disagrees here.

lib/searchablelist.fz
```diff
@@ -72,21 +72,31 @@ is
...
+                   h,        # the head part of from with f already replaced by r
+                   t list A, # the tail that still needs to be searched for f
+                   limit option u64
+                  ) list A
+  is
+    match (searchablelist t).find f
+      nil   => h ++ t
+      n i32 =>
+        if (limit.map (v -> v ≟ (u64 0))).get false
```
michaellilltokiwa, line 96: @fridis Will it be possible to implement `type.equality` for `option u64`? Then this could be `limit ≟ option 0`

michaellilltokiwa: Then it makes sense to add a NYI with #637.

lib/searchablelist.fz
michaellilltokiwa: Hmm, right. We should not use verbs here. What about treasure and replacement?

michaellilltokiwa: Old and new seems easy enough.

PR #854, opened 2023-01-03
lib/cache: improve documentation using better example

maxteufel:
This also makes use of the new functionality that allows code to be added to the comments of features, which will be executable on flang.dev.


fridis:
Good, I got confused by the original documentation as well!

PR #863, opened 2023-01-04
tests: increment stack size for fileio test

fridis:


fridis:
The interpreter does not have an tail recursion optimization, that's why!

I guess the JIT is the reason why this works sometimes.

fridis:
I think you have not understood: Loop in Fuzion will be translated to recursive calls, and there is no tail recursion optimization in the interpreter, and there never will be. There will be a new JVM/Bytecode back-end and some point that should have tail recursion optimization. 
So, for the interpreter backend, the only solution is to increase the Java stack.

PR #865, opened 2023-01-04
rename Object to Any

maxteufel:


fridis:
Can you try to solve the conflicts... first time I see github complaining that conflicts are too complex for the web editor.
Apart from that, the patch looks good. Maybe another patch should rename `FuzionConstants.OBJECT_NAME` as `FuzionConstants.ANY_NAME`.

fridis:
> > Maybe another patch should rename `FuzionConstants.OBJECT_NAME` as `FuzionConstants.ANY_NAME`.
> 
> I was not sure about this. Perhaps there is a better name for this? Because renaming this every time we change it defeats the point of such a constant. On the other hand, we probably don't want to rename Any as anything else soon.

The main point of having such a constant is not miss any uses of the constant when the constant changes or we need to know everything that depends on the constant. So renaming it in case the constant changes is no problem, the main purpose of having the constant is still fulfilled (and an IDE would even point you to all the uses automatically, which cannot be done for constants).

PR #875, opened 2023-01-09
lib: io: write improvements

maxteufel:


lib/fuzion/std/fileio.fz
```diff
@@ -78,22 +78,17 @@ public fileio is
...
-               path String,
-               # the content of the file as an u8 array
-               content array u8) outcome unit is
-    (open path (i8 1)).bind unit (fd->
-      write_results := write fd content.internalArray.data content.length
-      if write_results == 0
-        unit
-      else
-        error "error number: $write_results")
+  # this might overwrite parts or the whole of an existing file.
```
michaellilltokiwa, line 83: what about: parts or all of

PR #898, opened 2023-01-10
Lib/modules/nom

michaellilltokiwa:
This a port of https://github.com/rust-bakery/nom
Also a basic json parser has been implemented on top of this.


michaellilltokiwa:
Good point but I think it is to early to put this in a separate repository. Like `modules/terminal` we should move this to a separate repository once we have a working package/module manager.

michaellilltokiwa:
succeeded by #2808

PR #936, opened 2023-01-16
fe: write/read feature visibility to/from fum-file

michaellilltokiwa:

PR #939, opened 2023-01-16
Docs ignore feature based on visibility

michaellilltokiwa:
#936 needs to be merged first


fridis:
> I don't think private features should be entirely hidden; they should be documented and thus also their documentation should be accessible in _some_ way. I think they should be in a `<div class="hidden-feature">` which is not shown by default but can be toggled by a button "Expert Mode" (or whatever) on the documentation page.

For the public documentation of a Fuzion module, any private features should be completely hidden. What might however be useful is to have a way to create documentation for the module developers that includes all internal features. Maybe an option could allow this, the resulting html might have ways to chose between public API and complete features.

PR #956, opened 2023-01-17
Errors: reduce default maximum error count to 10

### LINKED ISSUE #951: reduce default maximum error count?
Being given 20 errors is rarely useful even if they were all valid errors. Should we choose a 'better' default?

maxteufel:
Fixes #951


michaellilltokiwa (approved):

michaellilltokiwa:
@tokiwa-software/developers Warning count is unlimited right now. Since there are only few warnings produced currently it does not make a big difference. But I think it makes sense to also limit those to five?

PR #986, opened 2023-01-20
lib/ryu: a float to string algorithm

### LINKED ISSUE #921: implement float.asString in fuzion
We should implement [ryu](https://github.com/ulfjack/ryu) or similar in fuzion. Currently (f32|f64).asString depends on the backend.

michaellilltokiwa:
fixes  #921


michaellilltokiwa:
This is blocked by the same issue as #355. The DFA does not reach a fix point quickly because of the large table/arrays.

michaellilltokiwa:
Not part of this PR:
- replace intrinsics `fXX.asString`
- formatting options beyond scientific/full-notation like controlling precision, decimal separator.

lib/ryu.fz
```diff
@@ -0,0 +1,994 @@
...
+      else if v >= 1000000000          then 10
+      else if v >= 100000000           then  9
+      else if v >= 10000000            then  8
+      else if v >= 1000000             then  7
+      else if v >= 100000              then  6
+      else if v >= 10000               then  5
+      else if v >= 1000                then  4
+      else if v >= 100                 then  3
+      else if v >= 10                  then  2
+      else                                   1
```
fridis, line 901: Might help to do this the other way around, i.e., start checking `v < 10`, then `v < 100`, etc. to shorten the most common paths.

fridis (approved):

PR #996, opened 2023-01-23
ast: Improve error created when arguments have equal names

fridis:
This error was helpful when working on type features where I ran into a problem due to repeated arguments.  But I needed more information to understand the problem, so this patch.

Moved the check for this error to the end of the Feature constructor such that all fields in feature needed by the error are initialized.

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -1579,15 +1579,23 @@ public class AstErrors extends ANY
       "Expected number given in base " + _base + " to fit into " + _type + " without loss of precision.");
   }
 
-  public static void argumentNamesNotDistinct(SourcePosition pos, Set<String> duplicateNames)
+  public static void argumentNamesNotDistinct(Feature f, Set<String> duplicateNames)
   {
-    error(pos,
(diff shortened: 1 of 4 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -1579,15 +1579,23 @@ static void lossOfPrecision(SourcePosition pos, String _originalString, int _bas
...
-      "Names of arguments used in this feature must be distinct.",
+    int[] cnt = new int[1];
+    error(f.pos(),
+          "Names of arguments used in this feature must be distinct.",
           "The duplicate" + (duplicateNames.size() > 1 ? " names are " : " name is ")
           + duplicateNames
             .stream()
             .map(n -> sbn(n))
             .collect(Collectors.joining(", ")) + "\n"
+          + "Feature with equally named arguments: "+ s(f) + "\n"
```
michaellilltokiwa, line 1592: Somehow this sounds a bit strange to me. Maybe "Feature contains duplicate argument names"?

fridis: Not sure, but sounds alright to me. We need a native speaker to proof read everything at some point...

PR #1010, opened 2023-01-24
GitHub Actions: speed up tests by running them in parallel

maxteufel:


michaellilltokiwa:
Even though I use this a lot on my machine the parallelizing of the tests is kind of hacky.  I would expect more problems than this solves. Also you @maxteufel mentioned this does not seem to result in faster runs on Github anyway?

michaellilltokiwa:
@maxteufel Fair point :-D

PR #1014, opened 2023-01-25
ast/Call: loadCalledFeature, try find operatorOnOuter before chained …

### LINKED ISSUE #1013: type inference fails in expression with multiple ≟
example:
```
ex_i64 is
  for m in (i64 -1)..1 do
    for n in (i64 -1)..1 do
      a bool := (m ≟ n)
      b bool := ((int m) ≟ (int n))
      c bool := (a ≟ b)

      x bool := ((m ≟ n) ≟ ((int m) ≟ (int n)))
```
for `a`, `b`, `c`, this works as expected, but type inference suddenly finds an `i64` in `x`:
```
ex_i64.fz:8:26: error 1: Incompatible types found during type inference for type parameters
      x bool := ((m ≟ n) ≟ ((int m) ≟ (int n)))
-------------------------^
Types inferred for first type parameter 'T':
i64 found at ex_i64.fz:8:26:
      x bool := ((m ≟ n) ≟ ((int m) ≟ (int n)))
-------------------------^
bool found at ex_i64.fz:8:37:
      x bool := ((m ≟ n) ≟ ((int m) ≟ (int n)))
------------------------------------^


ex_i64.fz:8:37: error 2: Incompatible types when passing argument in a call
      x bool := ((m ≟ n) ≟ ((int m) ≟ (int n)))
------------------------------------^
Actual type for argument #2 'b' does not match expected type.
In call to          : 'infix ≟'
expected formal type: 'i64'
actual type found   : 'bool'
assignable to       : 'bool'
for value assigned  : 'infix ≟<int>(int(deref(#loop0.this.#^ex_i64.#loop1.#loop0).m), int(#loop0.this.n))'
To solve this, you could change the type of the target 'b' to 'bool' or convert the type of the assigned value to 'i64'.

2 errors.
```

michaellilltokiwa:
…booleans

fixes #1013

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -706,11 +706,11 @@ public class Call extends AbstractCall
                   }
                 if (_calledFeature == null)
                   {
-                    findChainedBooleans(res, thiz);
+                    findOperatorOnOuter(res, thiz);
                   }
                 if (_calledFeature == null)
                   {
-                    findOperatorOnOuter(res, thiz);
+                    findChainedBooleans(res, thiz);
                   }
                 if (_calledFeature == null) // nothing found, so flag error
                   {


michaellilltokiwa:
This PR does not work.
The actual problem seems to be that `findChainedBooleans` is not considering the parentheses...

PR #1030, opened 2023-01-30
ast: type inference failed for operator calls

### LINKED ISSUE #1013: type inference fails in expression with multiple ≟

michaellilltokiwa:
- reason was that parentheses were not represented in AST
- this resulted in Call.loadCalledFeature() to try and find (and erroneously succeed in this case) a chain booleans before trying to find the operator on outer.
- fixes  #1013

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -1609,7 +1609,12 @@ public class Call extends AbstractCall
                     inferGeneric(res, t, actualType, actual.pos(), conflict, foundAt);
                     checked[vai] = true;
                   }
+                // NYI cleanup/merge the two cases below
                 else if (actual instanceof Function af)
+                  {
+                    checked[vai] = inferGenericLambdaResult(res, outer, t, af, actual.pos(), conflict, foundAt);
+                  }
+                else if (actual instanceof Block b && b.resultExpression() instanceof Function af)
                   {
                     checked[vai] = inferGenericLambdaResult(res, outer, t, af, actual.pos(), conflict, foundAt);
                   }
--- a/src/dev/flang/ast/Function.java
+++ b/src/dev/flang/ast/Function.java
@@ -244,7 +244,7 @@ public class Function extends ExprWithPos
 
 
   /**
-   * Special version of propagateExpetedType(res, outer, t) tries to infer the
+   * Special version of propagateExpectedType(res, outer, t) tries to infer the
    * result type of a lambda.
    *
    * @param res this is called during type inference, res gives the resolution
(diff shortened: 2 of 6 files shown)


michaellilltokiwa:
@fridis what should we do with these failed cases? disallow and thus remove from the tests or add special handling for numeric literals enclosed in braces or ...?
```
s9a i16 := -(32768)
s9c i16 := -(-(-32768))
s9a := i16 -(32768)
s9c := i16 -(-(-32768))
```

fridis:
> @fridis what should we do with these failed cases? disallow and thus remove from the tests or add special handling for numeric literals enclosed in braces or ...?
> 
> ```
> s9a i16 := -(32768)
> s9c i16 := -(-(-32768))
> s9a := i16 -(32768)
> s9c := i16 -(-(-32768))
> ```

I do not fully remember how I made these work, but I guess that they would work if instead of wrapping an actual in a Block, you would do `actual instanceof OpExpr oe ? oe.toExpr : actual`.

fridis:
Now I understand: In OpExpr, there is special handling for `prefix +` and `prefix -` when applied to `NumLiteral`. This no longer works when the `NumLiteral` was wrapped in a `Block`. So the solution is to either not wrap `NumLiteral` into a block (just like you do not wrap `Function`), or to extend the special handling in `OpExpr` to work for `NumLiteral` wrapped into a `Block`.

PR #1056, opened 2023-02-02
workflow linux: disable parallel test runs

michaellilltokiwa:
exit codes are not properly handled in parallel test execution.


michaellilltokiwa:
fixed in #1057

PR #1060, opened 2023-02-02
ast: AbstractType.constraintAssignableFrom, relax precondition, actua…

michaellilltokiwa:
…l might be error


michaellilltokiwa:
I jumped to conclusions too quickly here... Closing

PR #1072, opened 2023-02-06
lib: stdin cleanup, remove temporary assignments

michaellilltokiwa:


lib/io/stdin.fz
```diff
@@ -90,19 +90,13 @@ private read_line0 outcome String : character_encodings is
             if r.is_empty # we read nothing yet, so we just skip the lf/cr
               r
             else
-              # NYI explicit tagging should not be necessary #671
-              tmp outcome String := r
-              abort tmp
+              abort (outcome String) r
```
fridis, line 93: I think you should be able to go one step further and do
```
  abort (outcome r)
```
can you try that?

michaellilltokiwa: I changed it.

PR #1074, opened 2023-02-06
fix spelling mistakes

michaellilltokiwa:


src/dev/flang/ast/Function.java
```diff
@@ -378,7 +378,7 @@ public Expr visit(FeatureVisitor v, AbstractFeature outer)
   public void findGenerics(FeatureVisitor v, AbstractFeature outer)
   {
     if (this._feature != null)
-      { /* NYI: Neeed? The following comment seems wrong: */
+      { /* NYI: Need? The following comment seems wrong: */
```
fridis, line 381: This should be `Needed?`

fridis (commented):
I start to believe I am dyslexic... Thanks for all the corrections!

Found a minor correction that was wrong.

PR #1080, opened 2023-02-07
move entry back to ordered_map.entry

### LINKED ISSUE #1044: `entry` should be moved back to `ordered_map.entry`
This currently does not work since handling of `this.type` fails for nested features: 
```
ordered_map(OK type : has_total_order,
            redef V type,
            ks array OK,
            vs array V) : Map OK V
  pre
    ks.length ≟ vs.length

is

  # entry is an index in ks/vs
  #
  entry(i i32) : has_total_order is

    key => ks[i]
    val => vs[i]

    fixed type.lteq(a, b (ordered_map OK V).entry) => a.key ≤ b.key
```
causes the error
```
./build/lib/ordered_map.fz:51:21: error 1: Wrong argument type in redefined feature
    fixed type.lteq(a, b (ordered_map OK V).entry) => a.key ≤ b.key
--------------------^
In 'ordered_map.entry.type.lteq' that redefines 'has_partial_order.type.lteq' argument type is 'ordered_map<ordered_map.type.OK, ordered_map.type.V>.entry', argument type should be 'ordered_map.entry.this.type (in type feature)' (from 'has_partial_order.this.type (in type feature)').  Original argument declared at ./build/lib/has_partial_order.fz:48:13:
  type.lteq(a, b has_partial_order.this.type) bool is abstract
------------^
To solve this, change type of argument to 'ordered_map.entry.this.type (in type feature)' at ./build/lib/ordered_map.fz:51:21:
    fixed type.lteq(a, b (ordered_map OK V).entry) => a.key ≤ b.key
--------------------^

one error.
```

maxteufel:
Remove the hack using a ref to the ordered_map.

Closes #1044.

### DIFF
--- a/lib/ordered_map.fz
+++ b/lib/ordered_map.fz
@@ -42,9 +42,24 @@ ordered_map(OK type : has_total_order,
 is
 
 
+  # entry represents the pair of key and value at the given
+  # index i of the ordered map.
+  #
+  entry(i i32) : has_total_order is
+
+
+    key => ks[i]
+    val => vs[i]
+
+
+    # total order over entries.
+    #
+    fixed type.lteq(a, b entry.this.type) => a.key ≤ b.key
+
+
   # a sorted array of entries of this map
   #
-  sortedEntries := sorted_array_of (ks.indices.mapSequence (entry OK V) (i -> entry ordered_map.this i)).as_array
+  sortedEntries := sorted_array_of (ks.indices.mapSequence (i -> entry i)).as_array
 
 
   # number of entries in this map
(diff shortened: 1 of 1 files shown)


lib/ordered_map.fz
```diff
@@ -42,9 +42,23 @@ ordered_map(OK type : has_total_order,
 is
 
 
+  # entry is an index in ks/vs
```
michaellilltokiwa, line 45: maybe change this comment to sth. like:
```
# an entry is the key/value
# in this ordered map at index i.
#
```

PR #1082, opened 2023-02-07
wip: remove pos from type

michaellilltokiwa:


fridis:
`Type.pos` does not make sense after a Type was interned or written to a .fum file, so yes, these should be removed. But I would suggest not to start removing the code for AbstractType.pos, but instead maybe rename it to something like `pos_TO_BE_REMOVED` or similar, and then check all the locations that use it.  If these locations use a `Type` before it was interned, then they are probably ok. Otherwise, it should be checked what position the error message should contain and some way to get that position should be implemented (e.g., by storing the position of a feature's result type as part of the feature).

PR #1106, opened 2023-02-11
fe: Remove fallback for modules to load source files instead of .fum

fridis:
This fallback is no longer needed since modules are sufficiently stable.

### DIFF
--- a/src/dev/flang/fe/FrontEnd.java
+++ b/src/dev/flang/fe/FrontEnd.java
@@ -168,19 +168,9 @@ public class FrontEnd extends ANY
       {
         _baseModule = null;
       }
-    for (int i = 0; i < options._modules.size(); i++)
-      {
-        var m = _options._modules.get(i);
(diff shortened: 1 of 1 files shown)


src/dev/flang/fe/FrontEnd.java
```diff
@@ -265,37 +255,23 @@ private LibraryModule module(String m, Path p)
...
    */
   LibraryModule loadModule(String m)
-  {
-    return loadModule(m, false);
-  }
-  LibraryModule loadModule(String m,
-                           boolean ignoreNotFound // NYI: remove when module support is stable
-                           )
   {
     var result = _modules.get(m);
```
michaellilltokiwa, line 259: I like the method `computeIfAbsent` from the `Map<K,V>` interface for lazily filling a TreeMap.

michaellilltokiwa (approved):

PR #1114, opened 2023-02-13
actualType: when g.formalGenerics() != f.generics()

michaellilltokiwa:


michaellilltokiwa:
@fridis
this commit https://github.com/tokiwa-software/fuzion/commit/00c390d8fd2d10823f2c2034fe36091cd6b6a19c broke all idioms using `ordered_map`.
This patch here fixes this but I don't really understand what I'm doing here...
In case `f` is a type feature `g.formalGenerics() == f.generics()` seems to always false and also `g.index()` is the wrong index in actualGenerics then because to first element is some kind of thisType?

fridis:
This was fixed by #1124.

PR #1117, opened 2023-02-14
workflow: check idioms

michaellilltokiwa:


michaellilltokiwa:
@maxteufel @fridis My idea would be to rewrite some of those shell scripts in Fuzion. We will 'just' need to implement spawning processes and capturing their output first and maybe other minor things?

michaellilltokiwa:
@maxteufel but that will change. Also compiled fuzion code should be able to compete with a bash script even today.

michaellilltokiwa:
closing since we have the Jenkins now.

PR #1131, opened 2023-02-16
lib/array: move array initialization to type feature new

maxteufel:
Move the three-argument feature array(T type, length i32, init i32 -> T) to the type feature array.type.new(length i32, init i32 -> T). This resolves name clashes with the array feature which takes a type and an internal array representation, and up until now used three dummy unit arguments to avoid name clashes.

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -116,7 +116,7 @@ Sequence(T type) ref is
   #
   as_array array T is
     s := as_stream
-    array T count i->s.next
+    (array T).type.new count i->s.next
 
 
   # create a stream and call 'for_each f' on it
--- a/lib/array.fz
+++ b/lib/array.fz
@@ -25,21 +25,14 @@
 
 # array -- one-dimensional immutable array
 #
-# This is the result type of array(type, i32, i32 -> T) which creates an
-# initalized immutable array
-#
-# NYI: This uses three dummy unit-type args (i.e. total of 5 args or 4 value
-# args) to avoid name clashes with routine array(T,length,init) (i.e., 3 args
-# or  2 value args).  These unit-type args can go if the routine would be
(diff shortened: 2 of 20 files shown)


michaellilltokiwa (approved):
Cool! Much cleaner imho.

fridis:
The workaround with the unit-type arguments to avoid the name clash is admittedly ugly, but this is an internal API only.

The consequence of this patch is that code to initialize an array (using type inference)
```
  a := array length (f i)
``` 
now would b
```
  a := (array type).type.new length (f i)
``` 
This is much more typing and IMHO confusing.

My suggestions:

- keep `array(T type, length i32, init i32 -> T)` as a convenience feature
- either have this feature call `array.type.new` (so we still need the dummy `unit` args), or
- move the code to create the internal array to the constructor of `array`. This would probably require some additional changes where array are created by different ways, e.g., `conststring` instances created in the backends...

michaellilltokiwa:
I think the dummy `unit` hack - even though just internal in stdlib - makes things harder to read and comprehend. Would much prefer just using a different name. Maybe `array_by_internal` or `array_init`  or  sth. else?

fridis:
> That is a bit unfortunate because that would rename the type itself (at least as implemented currently).
> 
> I'm currently looking into resolving this as follows: `array` the type could take a single type argument `T`, and provide `internalArray` as `abstract`. Then defining
> 
> ```
> array_from_internal(T type, internalArray fuzion.sys.internal_array T) is
> ```
> 
> should be enough to provide an array from an internal array. The difference of arguments should make the previous usage of `array length init` with type inference possible.

having `array` with abstract features means that there will be a heir of array that implements this abstract feature, but then `array` will have to be a `ref` feature `Array`, causing one more heap allocation...

PR #1148, opened 2023-02-20
lib: introduce Mutable_Linked_List

maxteufel:
A Mutable_Linked_List implements a singly linked list using a given local mutate environment. Currently, only basic features are supported: creation of a list with a single element, appending single elements to a given list, and concatenating two lists.


lib/Mutable_Linked_List.fz
```diff
@@ -0,0 +1,111 @@
...
+  # a list with a single element.
+  #
+  private n := LM.env.new (option (Mutable_Linked_List LM T)) nil
+
+
+  # short-hand features to get the mutable references from `n`
+  #
+  next option (Linked_List T) is
+    match n.get
+      ll Mutable_Linked_List LM T => ll
```
fridis, line 50: The type parameters `LM T` can be omitted here, they will be inferred.

fridis (commented):
Looks pretty good, just a minor suggestion to use type inference. 

What is a little unfortunate (and I do not have a nice solution yet) is that for the static analysis `mutate.new.get` always requires the mutate effect, you should see this when your run `fz -effects` on your test. 

One way around this is to use `val` instead of `get`,  but this only works after `freeze`, so `as_string` would either work before or after `freeze`, but not both. 

What might work is to create a new, immutable linked list on `freeze`, but then we are creating the list twice, not nice. 

A better solution I might implement some time is to instantly freeze all mutable values when leaving the mutate effect, and let the DFA analysis know about this such that it does not report the dependency on the mutate effect if there is a call to `get`.

PR #1150, opened 2023-02-22
lib: add `list_of`, create list from head and tail

michaellilltokiwa:


fridis (commented):
This is basically a duplicate of `list(T, h, t)`  without lazy evaluation. Similar to `type_of` vs. `type_of_lazy` from #1163. I will create an issue that we need better handling of lazily evaluated arguments such that we do not need to duplicate features.

Can you add a NYI-comment with a reference to that issue and rename `list_of` as `list_of_non_lazy` or similar?

fridis (approved):

fridis (approved):

PR #1151, opened 2023-02-22
lib: list.type.infinite, create infinite list of only one unique element

michaellilltokiwa:


michaellilltokiwa:
The use case I had in mind was sth. like this:
```
list.type.infinite.reduce "" ((r,_)-> 
  if stream_exhausted then abort r else r + read_next_chunk_from_input_source.as_string
  )
```

fridis (commented):
I prefer sticking to the name `cycle`, but maybe adding it to `Sequence` such that you can write `[a].cycle`. 
If you wan to add it to `type`, you could to 
```
  type.cycle (el A) => el : ()->(type.cycle el)
```
but calling this 
```
  els => (list A).type.cycle el
```
is more typing than just doing
```
  els => el : ()->els
```
:-).

michaellilltokiwa:
I changed my mind about this PR. Having `cycle` is fine / good enough.

PR #1152, opened 2023-02-22
lib: list.concat_lazy, variant that takes function to create second list

michaellilltokiwa:


fridis (commented):
Can you add an example when this should be used instead of concat?
Also add a ref to #1166, we might want to make `concat` itself lazy and avoid this duplication.

fridis (approved):

PR #1154, opened 2023-02-22
lib: integer.as_bytes, convert u8-u64 to an array of bytes

michaellilltokiwa:


lib/wrapping_integer.fz
```diff
@@ -204,3 +218,8 @@ wrapping_integers(W type : wrapping_integer W) : numerics W is
 
   # returns the number in whose bit representation all bits are ones
   all_bits_set W is abstract
+
+
+  # how many bytes does this integer use?
+  bytes i32 is
```
fridis, line 224: This name confused me, should maybe be `byte_count`  or `byte_size`?

Also, this should be moved to `wrapping_integer.type`.

lib/wrapping_integer.fz
```diff
@@ -194,6 +194,20 @@ wrapping_integer(redef W type : wrapping_integer W) : integer W, wrapping_intege
...
+  # integer.
+  #
+  ones_count i32 is abstract
+
+
+  # this integer as an array of bytes
+  as_bytes array u8 is
+    array u8 bytes (idx ->
+      shift W := from_u32((bytes.as_u32-idx.as_u32-1)*8)
+      (wrapping_integer.this>>shift).low8bits)
```
fridis, line 207: This needs to document what byte ordering is used (little or big endian or platform dependent?). My preference is little endian as default, since then you can read it as a different size integer and you get the same value (provided it fits or padded with zeros). 

For integers in .fum-files, we use little endian for Fuzion constants and big endian for other integers (I don't know why I did that, maybe Java influence), this should all be little endian except for the magic numbers to identify files (MIR_FILE_MAGIC etc.).

fridis (commented):
Very nice to implement this in a way independent of the actual integer type!
Just some minor comment.

fridis (approved):

PR #1155, opened 2023-02-22
int: openresources, abstract class to hold open resources like files,…

michaellilltokiwa:
… socket, threads.

### DIFF
--- a/src/dev/flang/be/interpreter/OpenResources.java
+++ b/src/dev/flang/be/interpreter/OpenResources.java
@@ -1,0 +1,88 @@
+/*
+
+This file is part of the Fuzion language implementation.
+
+The Fuzion language implementation is free software: you can redistribute it
+and/or modify it under the terms of the GNU General Public License as published
+by the Free Software Foundation, version 3 of the License.
+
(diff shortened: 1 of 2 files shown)


src/dev/flang/be/interpreter/OpenResources.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * This acts as container for open/active resources like files, sockets, threads etc.
+ */
+public abstract class OpenResources<T>
+{
+  private ConcurrentHashMap<Long, T> data = new ConcurrentHashMap<Long, T>();
+
+  private AtomicLong cnt = new AtomicLong(0);
```
fridis, line 39: I prefer to start with a large number, maybe `System.identityHashCode(this)*(1L<<32)` or similar. The reason is to see the code crash early in case ids for different resources get mixed up.

fridis (commented):
Much nicer. 
Two small comments from my side.

fridis (approved):

PR #1177, opened 2023-02-28
lib: introduce mutable_tree_map

maxteufel:
mutable_tree_map implements a mutable map based on AVL trees.

### DIFF
--- a/lib/Entry.fz
+++ b/lib/Entry.fz
@@ -1,0 +1,122 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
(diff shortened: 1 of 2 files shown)


lib/Entry.fz
```diff
@@ -0,0 +1,122 @@
...
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature Entry
+#
+# -----------------------------------------------------------------------
+
+
+# Entry -- an entry of a mutable tree map
+#
+private Entry(LM type : mutate, K type : has_total_order, V type, k K, v V) ref is
```
michaellilltokiwa, line 27: probably better to choose a more specific name here: `Tree_Map_Entry`?

lib/mutable_tree_map.fz
```diff
@@ -0,0 +1,320 @@
...
+
+  # returns an empty tree of elements of type A.
+  #
+  type.empty mutable_tree_map LM K V is
+    mutable_tree_map LM K V
+
+
+  # returns a tree of elements of type A that contains just the element a.
+  #
+  type.singleton(k K, v V) mutable_tree_map LM K V is
```
michaellilltokiwa, line 315: `type.new` would be handy, similar to `map_of(K type : has_total_order, V type, kvs array (tuple K V))`

fridis (changes requested):
Looks good. 
Just some general comments that can be implemented in separate PR: 

- we should move features from `universe` into sub-features like `adt` to reduce the noise in the global name space
- we should think about higher level abstractions early on. The Java world is going through the painful process of adding these afterwards [JEP431](https://openjdk.org/jeps/431). We have several sorted data structures already.

fridis (approved):

PR #1181, opened 2023-03-01
lib: pass array and length to intrinsics instead of c_string

### LINKED ISSUE #1143: replace `c_string` in stdlib by array and length
Right there are two ways how we pass strings in intrinsics.
One is passing an array and a length separately as in `fuzion.sys.out.write`, number two is passing a null terminated array as in `fuzion.sys.env_vars.has0`. We should use only one mechanism.

maxteufel:
Closes #1143.

### DIFF
--- a/include/fz.h
+++ b/include/fz.h
@@ -68,4 +68,18 @@ int fzE_unsetenv(const char *name){
 }
 
 
+
+// \0-terminate a string that is given as an array plus length
+char * fzE_zero_terminate(const char *src, size_t n){
+#if GC_H
+  char *dest = GC_MALLOC(n + 1);
+#else
+  char *dest = malloc(n + 1);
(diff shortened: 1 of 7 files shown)


include/fz.h
```diff
@@ -68,4 +68,18 @@ int fzE_unsetenv(const char *name){
 }
 
 
+
+// \0-terminate a string that is given as an array plus length
+char * fzE_zero_terminate(const char *src, size_t n){
+#if GC_H
```
miclill, line 74: We should probably have a `fzE_malloc()`

fridis:
I prefer not to add additional allocation of C strings at this point since this would make the implementation of a future GC more complicated. We might revisit this once we have a GC and we see that the string copying is a performance issue.

PR #1190, opened 2023-03-03
air: Fix some preconditions and checks that caused failures during debugging

fridis:
Making the code more robust.


src/dev/flang/air/Clazz.java
```diff
@@ -2156,7 +2158,7 @@ else if (!t.dependsOnGenerics())
                   }
                 return res;
               }
-            else
+             else
```
michaellilltokiwa, line 2161: accidental space?

michaellilltokiwa (approved):

PR #1195, opened 2023-03-07
ast: propose solution if feature is called with the wrong number of args

### LINKED ISSUE #1184: error `Could not find called feature` could list features with different argument counts
In this little example
```
test_three_args is
  x(a, b, c Any) =>
    say "$a $b $c"

  x 1
  x 1 2
  x 1 2 3
  x 1 2 3 4
```
Three errors are produced as expected
```
 > ./build/bin/fz test_three_args.fz

/home/fridi/fuzion/work/test_three_args.fz:5:3: error 1: Could not find called feature
  x 1
--^
Feature not found: 'x' (one argument)
Target feature: 'test_three_args'
In call: 'x 1'


/home/fridi/fuzion/work/test_three_args.fz:6:3: error 2: Could not find called feature
  x 1 2
--^
Feature not found: 'x' (2 arguments)
Target feature: 'test_three_args'
In call: 'x 1 2'


/home/fridi/fuzion/work/test_three_args.fz:8:3: error 3: Could not find called feature
  x 1 2 3 4
--^
Feature not found: 'x' (4 arguments)
Target feature: 'test_three_args'
In call: 'x 1 2 3 4'

3 errors.
```
It would be nice if the errors could explain that an `x` with three arguments exists, so in the calls there are arguments missing or too many arguments given.

maxteufel:
Closes #1184.

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -77,6 +77,15 @@ public class AstErrors extends ANY
   {
     return sbn(fn.baseName()) + fn.argCountAndIdString();
   }
+  static String slbn(List<FeatureName> l)
+  {
+    var sl = new List<String>();
+    for (var fn : l)
+      {
+        sl.add(sbn(fn));
+      }
+    return sl.toString();
+  }
   static String sbn(String s) // feature base name
   {
     return code(s);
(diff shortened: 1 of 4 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -1038,6 +1038,39 @@ static String solutionDeclareReturnTypeIfResult(String name, int argCount)
...
+          }
+        else if (fn.equalsBaseName(name))
+          {
+            found.add(fn);
+          }
+      }
+
+    if (!found.isEmpty())
+      {
+        solution = "Did you perhaps mean to use " + (found.size() > 1 ? "one of these features" : "this feature") + ": " + found + "?";
```
fridis, line 1068: Please reformulate this and start with `To solve this, `, this is used in all other error messages when proposing a workaround: something like `To solve this, you might change the actual number of arguments to match ...

src/dev/flang/ast/AstErrors.java
```diff
@@ -1038,6 +1038,39 @@ static String solutionDeclareReturnTypeIfResult(String name, int argCount)
...
+        if (f.isChoice() && !f.isBaseChoice())
+          {
+            /* ignore call to choice type */
+          }
+        else if (fn.equalsBaseName(name))
+          {
+            found.add(fn);
+          }
+      }
+
```
fridis, line 1065: For code maintenance, it would be much better not to have this logic duplicated far away from the call lookup code. Instead, you couldadd it to FeatureAndOuter, maybe add a variant of `filter` that does this. Use this variant in the existing FeatureAndOuter.filter and call the new variant in Call to pass this list to AstErrors.

fridis (commented):
Good work!

Just minor suggestions to improve consistency and code maintenance. 

Also, would be good if you could show an example output in the PR or commit message, make review easier... :-)

fridis (approved):

PR #1221, opened 2023-03-14
switch from msys (cygwin) to clang64 (ucrt)

michaellilltokiwa:
On windows there are several options of a C standard library. ucrt (Universal C Runtime) seems to be the better choice than cygwin.
For more information see also:
https://www.msys2.org/docs/environments/
image


fridis:
I do not have any opinion on this, maybe @maxteufel has one?

PR #1224, opened 2023-03-14
add a `Unary` function type which allows function composition

### LINKED ISSUE #154: base Library: Add Function.infix ∘ for function composition
should be possible to write
```
h := f ∘ g
```
instead of 
```
h t1->t3 := x -> f (g x)
```

maxteufel:
Closes #154.


tests/unary/ex_unary.fz
```diff
@@ -0,0 +1,72 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test ex_unary
+#
+# -----------------------------------------------------------------------
+
+ex_unary is
+
+  g ref : Unary i32 i32 is
```
michaellilltokiwa, line 26: There could be a type feature `Unary.type.create i32 i32 (x->2*x)` but this seems only marginally better than `ref : ...`

lib/Unary.fz
```diff
@@ -0,0 +1,39 @@
...
+#
+# For unary functions, function composition is possible if the result type of
+# the first function matches the argument type of the second function.
+#
+Unary(T, U type) ref : Function T U is
+
+
+  # function composition
+  #
+  infix ∘ (V type, other Unary U V) Unary T V is
```
fridis, line 36: please rename this as `compose` and make `infix ∘` call `compose`. Any operator that cannot be found easily on a standard keyboard should be just an alternative for function with a normal name.

lib/Unary.fz
```diff
@@ -0,0 +1,39 @@
...
+    ref : Unary T V
+      redef call (v V) T is
+        Unary.this.call (other.call v)
```
fridis, line 39:
```
v -> call (other.call v) 
```
should work as an implementation as well, maybe even 
```
v -> Unary.this (other v) 
```

src/dev/flang/ast/Function.java
```diff
@@ -314,8 +317,18 @@ public AbstractType propagateExpectedType2(Resolution res, AbstractFeature outer
...
 
+            var _inheritsName = Types.FUNCTION_NAME;
+            if (t.featureOfType() == Types.resolved.f_Unary && gs.size() == 2)
+              {
+                _inheritsName = Types.UNARY_NAME;
+              }
+            else if (t.featureOfType() == Types.resolved.f_Lazy && gs.size() == 1)
+              {
+                _inheritsName = Types.LAZY_NAME;
+              }
```
fridis, line 328: A perfect candidate for Java's new switch syntax using `->`.

src/dev/flang/ast/Function.java
```diff
@@ -314,8 +317,18 @@ public AbstractType propagateExpectedType2(Resolution res, AbstractFeature outer
             var f = new Feature(pos(), rt, new List<String>("call"), a, _inherits, _contract, new Impl(_expr.pos(), _expr, im));
             this._feature = f;
 
+            var _inheritsName = Types.FUNCTION_NAME;
```
fridis, line 320: The idea of `_` prefix is that `_` is used only for fields in the surrounding class. This is mainly to avoid confusion in Java constructors that tend to have argument names equal to field names in the class.

fridis (changes requested):
Nice work. 
Just minor requests for changes.

src/dev/flang/ast/Function.java
fridis: right, my fault...

lib/Unary.fz
fridis: The type of the outer `call` is `T`, but the type of `v -> call..` should then be `Unary T V`. Anyway, this should be reproducible in a small test, can you check and create issues for these two cases?

fridis (approved):

PR #1225, opened 2023-03-14
ast: Generic.replaceOpen(), improve behaviour on error.

michaellilltokiwa:


fridis:
Would be good if you would be more specific in the PR message on what this does. From the commit, I guess you ran into a precondition failure as a consequence of previous errors, and you avoid this through this PR. But without reading the code, I thought you just improved some error message.

fridis (approved):

PR #1231, opened 2023-03-15
ast: reset statically held FeatureAndOuter.ERROR

michaellilltokiwa:
Led to problems in language server. Because `Types.f_ERROR` is reset but `FeatureAndOuter.ERROR` holds first instance of `Types.f_ERROR` forever.


fridis:
oops, my fault....

PR #1233, opened 2023-03-15
lib/set: fix rename mistake in comment: values V -> values E

michaellilltokiwa:


fridis:
> actually, I'm not sure. Does the `V` here really refer to the type of the values in the set, I personally think it can also be read as referring to the set itself. Perhaps the best option would be to simply remove this?

The intention was probably `set of values of type E`.

PR #1235, opened 2023-03-15
lib/map_of_total_order: add effect map_of_total_order

michaellilltokiwa:
Abstracts implementation of Map with totally ordered keys. This could be done for other data structures as well.

example:

```
ex =>
  a := (map_of_total_order i32 String)
    .new [1,2,3] ["a", "b", "c"]

  say a

  # install CTrie as default map implementation
  map_of_total_order i32 String (ks,vs->
    ct := CTrie i32 String
    for idx in ks.indices do
      ct.add ks[idx] vs[idx]
    ct
  )

  b := (map_of_total_order i32 String)
    .new [1,2,3] ["a", "b", "c"]

  say b

```
```
(1 => a), (2 => b), (3 => c)
(1 => a), (2 => b), (3 => c)
```

### DIFF
--- a/lib/map_of_total_order.fz
+++ b/lib/map_of_total_order.fz
@@ -1,0 +1,53 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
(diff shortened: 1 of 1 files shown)


fridis (changes requested):
Good idea. 

The default handler should not be replaceable, the default handler is just a preliminary solution until `fz` would wrap the application in code that installs a handler and executes the application inside that effects `run`. 
So I suggest that to use a different map implementation, one would install one via `effect.run`, only the code executed by lambda  passed to `run` should then be affected.

fridis (approved):

PR #1239, opened 2023-03-15
ast: Fix result type inference from lazy expression, make `infix :` lazy

fridis:
Moved the code to expand a call to a lazy feature `f` to `f.call` to Call.resolveTypes such that the new, non-lazy result type will be used in type inference.

This enables changing the last argument of `infix :` and `list(T, h t)` to be lazy.

This permits lazy lists like the following to produce all integers starting at `i`:

```
  ints(i i32) =>
    i : ints i+1
```

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -2066,7 +2066,6 @@ public class Call extends AbstractCall
 
                 // Convert a call "f.g a b" into "f.g.call a b" in case f.g takes no
                 // arguments and returns a Function or Routine
-                if (_resolvedFor != null)
                 result = resolveImmediateFunctionCall(res, outer); // NYI: Separate pass? This currently does not work if type was inferred
               }
           }


src/dev/flang/ast/Call.java
```diff
@@ -1080,17 +1080,20 @@ public Expr visit(FeatureVisitor v, AbstractFeature outer)
...
+    if (!_forFun                                     && // not a call to "b" within an expression of the form "fun a.b", will be handled after syntactic sugar
+        (_type.isFunType() &&
+         _calledFeature != Types.resolved.f_function && // exclude inherits call in function type
+         _calledFeature.arguments().size() == 0      &&
+         hasParentheses()
+         ||
+         _type.isLazyType()                          &&   // we are `Lazy T`
+         _calledFeature != Types.resolved.f_Lazy     &&   // but not an explicit call to `Lazy` (e.g., in inherits clause)
+         _calledFeature.arguments().size() == 0      &&   // no arguments (NYI: maybe allow args for `Lazy (Function R V)`, then `l a` could become `c.call.call a`
+         _actualsNew.isEmpty()                       &&   // dto.
```
michaellilltokiwa, line 1094: according to https://www.wortbedeutung.info/dto `dto.` is german. :-)

michaellilltokiwa (approved):

PR #1241, opened 2023-03-16
lib/list: remove `list_of_none_lazy`

maxteufel:
This is no longer necessary now that `list` and `infix :` take values directly.


fridis:
Thanks for fixing my broken comments!

PR #1257, opened 2023-03-21
lib/envir/vars: switch to `simpleEffect`

maxteufel:


fridis (changes requested):
I am happy to see some cleanup in the effects code, in particular I would like to get rid of the confusing `effectMode`. 

However, having the `Vars_Provider` is IMHO an important feature of the `envir.vars` effect since it permits to run code that accesses `envir.vars` without giving any access to the system's environment to that code, but diverting all accesses to a `Vars_Provider` that the caller controls. 

A solution would be to have either extend `simpleEffect` to support provider or have a new `lessSimpleEffect` that does (please chose a better name).

michaellilltokiwa (approved):

fridis (approved):

PR #1264, opened 2023-03-23
README: add code examples

### LINKED ISSUE #1125: Readme.md, add 1-3 fuzion code examples

maxteufel:
Closes #1125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Scorecard](https://api.securityscorecards.dev/projects/github.com/tokiwa-softwar
 ---
 
 <!--ts-->
+   * [Examples](#examples)
    * [Documentation](#documentation)
    * [Clone](#clone)
    * [Requirements](#requirements)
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
@maxteufel Good examples. I would move the examples below the table of contents and add mark there.
Also IMHO pre/post conditions should be show in the examples. 

What do you think of including the examples as pictures that have syntax highlighting?

michaellilltokiwa:
@maxteufel I think you're right. Text is probably better. But we could include a link to some flang.dev-page where one can play with the examples immediately. 

Regarding the first example with the pre-/post- conditions. I don't think this is how contracts should be used. Basically this would mean that _wrong_ user input violates the contract. In my understanding a contract should enforce correct usage of code by programmers not by the users of the program: https://en.wikipedia.org/wiki/Design_by_contract

fridis (approved):

PR #1267, opened 2023-03-23
parser: add support for multiline strings

michaellilltokiwa:
- multiline strings are denoted by fat quotations """
- the start of the multiline string is in the first line after the fat quotation marks
- the start of the multiline string defines the indentation required for the rest of the string.
- trailing whitespace are disallowed in multiline strings
- line breaks are normalized to LF

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -27,6 +27,7 @@ Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
 package dev.flang.parser;
 
 import java.nio.file.Path;
+import java.util.Optional;
 
 import dev.flang.ast.*;
 
@@ -2131,7 +2132,7 @@ simpleterm  : bracketTerm
           default          :
             if (isStartedString(current()))
               {
-                result = stringTerm(null);
+                result = stringTerm(null, Optional.empty());
               }
             else
               {
@@ -2175,19 +2176,23 @@ stringTermB : '}any chars&quot;'
             | '}any chars{' block stringTermB
             ;
   */
-  Expr stringTerm(Expr leftString)
+  Expr stringTerm(Expr leftString, Optional<Integer> multiLineIndentation)
   {
     return relaxLineAndSpaceLimit(() -> {
         Expr result = leftString;
         var t = current();
         if (isString(t))
           {
-            var str = new StrConst(posObject(), string());
+            var ps = string(multiLineIndentation);
+            var str = new StrConst(posObject(), ps._v0);
             result = concatString(posObject(), leftString, str);
             next();
             if (isPartialString(t))
               {
-                result = stringTerm(concatString(posObject(), result, block()));
+                var old = minIndent(-1);
+                var b = block();
+                minIndent(old);
+                result = stringTerm(concatString(posObject(), result, b), ps._v1);
               }
           }
         else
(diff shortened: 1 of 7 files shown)


tests/strings_multiline_negative/strings_multiline_negative.fz
```diff
@@ -0,0 +1,45 @@
...
+# -----------------------------------------------------------------------
+
+strings_multiline_negative is
+
+  x := 1
+
+  a := """"""    //  1. should flag an error: Expected multiline string to start in first line following fat quotation
+  b := """ """   //  2. should flag an error: Expected multiline string to start in first line following fat quotation
+
+                 //  3. should NYI flag NYI an NYI error: (error in line 32) Expected multiline string to start in first line following fat quotation
```
michaellilltokiwa, line 31: This negative test case does not work since I can not put the comment on the line where the error occurs.

tests/strings_multiline_negative/strings_multiline_negative.fz
```diff
@@ -0,0 +1,45 @@
...
+
+  d := """
+  $x
+ d              //  4. should flag an error: Found codepoint at less indentation than expected in multiline string.
+  """
+
+  e := """
+  hello
+  """
+  //   ^  5. should NYI flag NYI an NYI error: (error in line 43) Illegal trailing whitespace in multiline string.
```
michaellilltokiwa, line 45: This negative test case does not work since I can not put the comment on the line where the error occurs.

src/dev/flang/parser/Parser.java
```diff
@@ -2175,19 +2176,23 @@ Expr term()
...
           {
-            var str = new StrConst(posObject(), string());
+            var ps = string(multiLineIndentation);
+            var str = new StrConst(posObject(), ps._v0);
             result = concatString(posObject(), leftString, str);
             next();
             if (isPartialString(t))
               {
-                result = stringTerm(concatString(posObject(), result, block()));
+                var old = minIndent(-1);
```
michaellilltokiwa, line 2192: This means that the code in braces inside of a multiline string is allowed to not be indented at all.
Maybe this is not what we want?

src/dev/flang/parser/Parser.java
```diff
@@ -2175,19 +2176,23 @@ Expr term()
             | '}any chars{' block stringTermB
```
michaellilltokiwa, line 2176: I did not update the grammar, yet.

src/dev/flang/parser/Parser.java
fridis: I think this is fine, let's play around with it. The alternative would be to use the indentation level before the string or of the string itself.

tests/strings_multiline/multiline_strings_test.fz
```diff
@@ -0,0 +1,72 @@
...
+  # nested multiline string using braces and dollar interpolation
+  say """
+    hell😀 {
+if true
+  """
+{"""
+  $world"""}"""
+else
+  "?"
+    }"""
```
fridis, line 72: I would like to see some more tests, e.g., starting a string with `"""` but ending with `"` and vice, versa, using four or more quotation marks `""""`, etc.

fridis (commented):
Very nice work. My small comments can be addressed in separate PRs. I will ask @maxteufel to review as well.

fridis (approved):

PR #1268, opened 2023-03-24
ast/Type: remove pos from Type

michaellilltokiwa:
- added pos to dev.flang.ast.Generic
- added pos() to AbstractType uses either position from generic or the feature of the type


src/dev/flang/ast/AbstractType.java
```diff
@@ -1442,6 +1440,28 @@ public AbstractType checkConstraints(SourcePosition pos)
...
+   *
+   * 1) internally defined types like universe, t_Error,
+   *    fuzion.sys.internal_array, unit etc.
+   *    result in SourcePosition.notAvailable
+   * 2) type is generic argument: some_feat(A type) is
+   *                                     ---^
+   * 3) type belongs to feature : some_feat is
+   *                           ---^
+   */
+  public SourcePosition pos()
```
michaellilltokiwa, line 1454: maybe we should name this `posOfDef()` or similar to not have it confused with other `pos()`

src/dev/flang/ast/Type.java
```diff
@@ -184,9 +176,9 @@ public String name()
    *
    * @param o
    */
-  public Type(SourcePosition pos, String n, List<AbstractType> g, AbstractType o)
+  public Type(String n, List<AbstractType> g, AbstractType o)
```
fridis, line 179: IMHO a `Type` created by the parser should be the only type that has the original source code position. If a type in the source code is wrong (e.g., a spelling mistake), we want to point to that source code position. 
For all instances of AbstractType that are created later, there should not be any need for the souce code position.

tests/generics_negative/generics_negative.fz
```diff
@@ -57,7 +57,7 @@ generics_negative is
 
   opengenerics9 (A type, B type, C type...) : opengenerics6 C A A A is // 7. should flag an error: open generic cannot be used as generic argument
 
-  opengenerics10 (A type, B type, C type...) is  // NYI: an error is flagged here in addition to 8. below. Apparently, 8. is checked again after Types were intern()ed.
+  opengenerics10 (A type, B type, C type...) is
```
michaellilltokiwa, line 60: This error is not flagged anymore due to the changes concerning `ensureNotOpen()`

PR #1288, opened 2023-03-29
air: fix `require-condition3 failed: dev.flang.air.Clazz:<init>:325`

michaellilltokiwa:
Clazz.findOuter() did not cover the case where clazz is choice. In this case outer ref is null and thus can not be used for following to find the the clazz we are looking for.

### DIFF
--- a/src/dev/flang/air/Clazz.java
+++ b/src/dev/flang/air/Clazz.java
@@ -2113,11 +2113,16 @@ public class Clazz extends ANY implements Comparable<Clazz>
      */
     var res = this;
     var i = feature();
-    while (i != o && i.outerRef() != null)
+    while (i != o)
       {
-        res = res.lookup(i.outerRef(), pos).resultClazz();
+        res =  i.hasOuterRef() ? res.lookup(i.outerRef(), pos).resultClazz()
+                               : res._outer;
         i = i.outer();
       }
+
+    if (CHECKS) check
+      (i == o);
+
     return res;
   }
 


michaellilltokiwa:
I found this because the tests in PR #898 and #1223 are failing because of this defect.

src/dev/flang/air/Clazz.java
```diff
@@ -2111,13 +2111,17 @@ Clazz findOuter(AbstractFeature o, HasSourcePosition pos)
...
+    var i = isChoice() ? _outer.feature() : feature();
     while (i != o && i.outerRef() != null)
       {
         res = res.lookup(i.outerRef(), pos).resultClazz();
         i = i.outer();
       }
+
+    if (POSTCONDITIONS) ensure
+      (i == o);
+
```
fridis, line 2124: This should be a `check()` since `i` is an internal local variable. Also, this condition implies that 
```
  while (i != o && i.outerRef() != null)
```
can be simplified to
```
  while (i != o)
```
A good postcondition would be
```
  if (POSTCONDITIONS) ensure
    (res != null); 
```

fridis (changes requested):
Good job finding this!
My suggestion is to make the code a bit more general and robust, see detailed comments.

fridis (approved):

fridis:
@michaellilltokiwa , can you try this version:

```
    var res = this;
    var i = feature();
    while (i != o)
      {
        res =  i.hasOuterRef() ? res.lookup(i.outerRef(), pos).resultClazz()
                               : res._outer;
        i = i.outer();
      }

    if (CHECKS) check
      (i == o);

    return res;
```

PR #1299, opened 2023-03-30
parser: allow dismissing newlines in multiline strings

### LINKED ISSUE #1277: In multiline strings, allow dismissal of line breaks by ending line with `\`
this:
```
say """
   hello \
   world"""
```
should print:
```
hello world
```

maxteufel:
The escape character followed by a line feed now causes the newline to be dismissed. This makes it possible to write
```
say """
   hello \
   world"""
```
to print
```
hello world
```

Closes #1277.

### DIFF
--- a/src/dev/flang/parser/Lexer.java
+++ b/src/dev/flang/parser/Lexer.java
@@ -2518,7 +2518,14 @@ PIPE        : "|"
                     }
                   if (c < 0)
                     {
-                      Errors.unknownEscapedChar(sourcePos(), p, escapeChars);
+                      if (p == (int) '\n')
+                        {
+                          // codepoint is skipped
+                        }
+                      else
+                        {
+                          Errors.unknownEscapedChar(sourcePos(), p, escapeChars);
+                        }
                     }
                   escaped = false;
                 }
--- a/src/dev/flang/parser/Lexer.java
+++ b/src/dev/flang/parser/Lexer.java
@@ -2389,7 +2389,7 @@ PIPE        : "|"
      * If this is changed, https://flang.dev/tutorial/string_constants
      * must be changed as well.
      */
-    char[][] escapeChars = new char[][] {
+    int[][] escapeChars = new int[][] {
         { 'b', '\b'  },  // BS 0x08
         { 't', '\t'  },  // HT 0x09
         { 'n', '\n'  },  // LF 0x0a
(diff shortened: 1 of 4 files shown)


src/dev/flang/parser/Lexer.java
```diff
@@ -2518,7 +2518,14 @@ else if (
                     }
                   if (c < 0)
                     {
-                      Errors.unknownEscapedChar(sourcePos(), p, escapeChars);
+                      if (p == (int) '\n')
+                        {
+                          // codepoint is skipped
+                        }
+                      else
```
fridis, line 2525: Would be very nice to add `'\n'` to `escapedChars` and handle this in the loop just before this code. 
For that to work, you will need to change the type of `escapedChars` to `init[][]` such that this could be `-1`.
The reason is that chances are much lower that this escape sequence gets overlooked in documentation etc. 

My suggestion for the code, another example where Fuzion's loops with `until` and `else` would fit well:
```
                  var i = 0;
                  while (i < escapeChars.length && p != (int) escapeChars[i][0])
                    {
                      i++;
                    }
                  if (i < escapeChars.length)
                    {
                      c = (int) escapeChars[i][1];
                    }
                  else
                    {
                      Errors.unknownEscapedChar(sourcePos(), p, escapeChars);
                    }
```

Soon, I would like to start adding a Fuzion reference manual. Things like the table of escape sequenced should then be created directly from the sources using asciidoc or similar markup.

src/dev/flang/parser/Lexer.java
```diff
@@ -2518,7 +2518,14 @@ else if (
                     }
                   if (c < 0)
                     {
-                      Errors.unknownEscapedChar(sourcePos(), p, escapeChars);
+                      if (p == (int) '\n')
```
michaellilltokiwa, line 2521: You could put \\n in in the `escapeChars` array, but additionally I would think that this escaping should only work for multiline strings, not single line strings? Furthermore we need to support `\r\n`.

fridis (changes requested):
Very nice. 
Just a suggestion to avoid extra logic in favor of extending the `escapedChars` table.

src/dev/flang/parser/Lexer.java
fridis: I think this could be allowed for normal (non-multiline) strings. 

Letting `\` at EOL skip \r\n would be nice. But I am not happy about the added complexity...

PR #1315, opened 2023-04-04
parser: remove escape chars that do not need to be escaped

maxteufel:
Suggested by @michaellilltokiwa.


michaellilltokiwa (approved):

fridis:
I would like to keep `\}` just to keep the symmetry, e.g., `"\{a+b\} results in {a+b}"`, even if this is not needed.

PR #1319, opened 2023-04-04
naming improvements in Lexer

michaellilltokiwa:
rename posObject -> lexerSourcePos,
remove lexer.posObject(arg), duplicate of sourcefile.sourcePos(arg)

While implementing multiline strings I frequently mixed up the methods  `sourcePos` and `posObject` in SourceFile and Lexer. So this is an attempt to lessen future confusion.


fridis (changes requested):
Yes, these are confusing. `SourceFile.sourcePos()` is the current position in the source file we are reading from, while `Lexer.posObject()` is the start position of the current token. I suggest going a bit futher and also rename `Lexer._pos` and `Lexer.pos()`, maybe as follows: 
`Lexer._pos` -> `Lexer._tokenPos`
`Lexer.pos()` -> `Lexer.tokenPos()`
`Lexer.posObject()` -> `Lexer.tokenSourcePos()`
Does this make sense?

fridis (approved):

PR #1325, opened 2023-04-05
remove handling of `fun` keyword

maxteufel:


fridis (approved):
always nice to see code being removed!

PR #1330, opened 2023-04-06
lib/exit: introduce a `delayed_exit` effect

maxteufel:
This effect allows code to set an exit code, but postpone the exit until all code has been executed. This is intended for use in tests.

### DIFF
--- a/lib/exit.fz
+++ b/lib/exit.fz
@@ -91,8 +91,6 @@ exit_type is
 # delayed_exit -- effect that runs some code that can set an exit code, but
 # only exits after all the code has been run
 #
-# this is intended for use in tests
-#
 delayed_exit (_ unit) : effect effectMode.plain is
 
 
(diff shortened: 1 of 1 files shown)


lib/exit.fz
```diff
@@ -86,3 +86,49 @@ exit_type is
     #
     exit(code i32) =>
       fuzion.std.exit code
+
+
+# delayed_exit -- effect that runs some code that can set an exit code, but
+# only exits after all the code has been run
+#
+# this is intended for use in tests
```
michaellilltokiwa, line 94: I think this need to be the only intended usage.

lib/exit.fz
```diff
@@ -86,3 +86,49 @@ exit_type is
...
+
+
+  # mutable value of the current exit code set
+  #
+  exit_code := mut 0
+
+
+  # update the exit code to the given value
+  #
+  set0(code i32) =>
```
michaellilltokiwa, line 106: I think the common case and what probably most would expect is that the first none zero exit code is the one that is latter used at exit?

fridis (changes requested):
I did not really think about this enough and gave too little instructions, so this solution is different to what I had in mind. 

I think it would be better to add `set_exit_code` to `exit` and `Exit_Handler` and add `exit` without argument  in addition to `exit(code i32)` to perform an exit using the previously set code. 

Then, code using this, would not need to be executed within `delayed_exit.go`, but could just use `exit.env.set_exit_code` and finally perform `exit.env.exit`. 

Once fuzion supports something like Java's `ShutDownHook`, the `exit.env.exit` could be performed by a low-priority hook executed during shutdown.

PR #1332, opened 2023-04-06
fz: allow specifying -1 as maximum warning/error count for no limit

### LINKED ISSUE #1329: -XmaxErrors/Warnings=0 should remove limits
Right now one error/warning is printed out.

maxteufel:
If given a maximum error or warning count of -1, fz will display all the errors or warnings. The value -1 was chosen because while 0 would work for errors, 0 can be used to disable all warnings.

Closes #1329.

### DIFF
--- a/src/dev/flang/util/Errors.java
+++ b/src/dev/flang/util/Errors.java
@@ -257,7 +257,7 @@ public class Errors extends ANY
       {
         _errors_.add(e);
         print(pos, errorMessage(msg), detail);
-        if (count() >= MAX_ERROR_MESSAGES)
+        if (count() >= MAX_ERROR_MESSAGES && MAX_ERROR_MESSAGES != -1)
           {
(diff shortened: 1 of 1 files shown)


michaellilltokiwa (approved):

src/dev/flang/util/Errors.java
```diff
@@ -454,7 +454,7 @@ public static void showAndExit(boolean warningStatistics)
   {
     if (count() > 0)
       {
-        if (count() >= MAX_ERROR_MESSAGES)
+        if (count() >= MAX_ERROR_MESSAGES && MAX_ERROR_MESSAGES != -1)
```
fridis, line 457: Can you check if this code can be moved to before the call of `showAndExit` in `error` to avoid this second `if` check?

fridis (approved):

PR #1343, opened 2023-04-11
lib: remove orphaned comment, remove `spit`

michaellilltokiwa:
I think we should avoid cusswords in the stdlib even though there has been some tradition for that in the software community...


fridis (approved):
I have the impression that `spit` sounds worse to German ears than to native English speaker's ears, but not sure. We have `yak` instead.

PR #1350, opened 2023-04-13
ast: calls, add isArtificial() method

michaellilltokiwa:
this indicates if a call is present explicitly in the source code or if it is the result of syntax sugar and/or has been created internally by the compiler.

### DIFF
--- a/src/dev/flang/ast/AbstractCall.java
+++ b/src/dev/flang/ast/AbstractCall.java
@@ -94,6 +94,13 @@ public abstract class AbstractCall extends Expr
 
 
   /**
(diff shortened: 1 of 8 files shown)


fridis (changes requested):
Good to have Calls created by the parser in a separate class. 

I do not understand the need for the isArtificial flag and I would like to keep the .fum file as simple as possible, i.e., I prefer not to add it unless there is a convincing reason to do so. 
So either convince me or remove this flag :-).

michaellilltokiwa:
Superseded by #1363

PR #1357, opened 2023-04-17
call: propagateExpectedType, remove wrapInLazy workaround

michaellilltokiwa:
might be not needed anymore


michaellilltokiwa:
Too bad, seems its still necessary

PR #1359, opened 2023-04-17
lib: remove usage of the `some` feature

maxteufel:


michaellilltokiwa (approved):

michaellilltokiwa:
I think some was added when nesting of choices did not yet work. Seems unnecessary now.

fridis (changes requested):
I am happy to see the uses of `some` removed. But there are still cases where one would need some, e.g., `option nil` does not work, has to be `option (some nil)`. Not sure if there are other possible uses yet.

fridis (approved):

PR #1372, opened 2023-04-19
lib/io/file: convert all inner features to use `simpleEffect`

maxteufel:


michaellilltokiwa (approved):

lib/io/file/delete.fz
```diff
@@ -25,59 +25,42 @@
 
 # delete -- effect wrapping the file delete operation
 #
-private delete(
-    dop Delete_Operation,
-    redef r effect_mode.val,
-    _ unit
-  ) : deletes, effect r is
+private delete0(dop Delete_Operation) : simpleEffect is
```
michaellilltokiwa, line 28: probably better to use an additional arg `_ unit` since the name of effects are user visible.

PR #1377, opened 2023-04-20
issue 1055: fix ensure-condition failures in `findInheritanceChain`

michaellilltokiwa:
- must not refer to - `.this.type` - none existant instance in type feautre
- must not call feature needing instance via `.type.<feat>`

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -1764,6 +1764,20 @@ public class AstErrors extends ANY
           code(new_code_ref) + "to return a reference.");
   }
 
+  public static void illegalUseOfThisTypeInTypeFeature(DotType d, AbstractFeature outer)
+  {
+    error(d.pos(),
+          "Illegal use of " + code(".this.type") + " inside of type feature " + s(outer) + ".",
+          "To solve this, remove the offending code.");
+  }
+
+  public static void featureRequiringInstanceCalledOnDotType(Call c, AbstractFeature cf)
+  {
+    error(c.pos(),
+          "Called feature " + s(cf) + " needs an instance, but was called like a type feature.",
+          "To solve this, remove the offending code.");
+  }
+
 
 }
 
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -663,7 +663,15 @@ public class Call extends AbstractCall
                   }
                 if (fo != null)
                   {
-                    _calledFeature = fo._feature;
+                    if (_target != null && _target.isDotTypeCall() && needsInstance(fo))
+                      {
+                        AstErrors.featureRequiringInstanceCalledOnDotType(this, fo._feature);
+                        _calledFeature = Types.f_ERROR;
+                      }
+                    else
+                      {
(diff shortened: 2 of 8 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -1764,6 +1764,20 @@ public static void routineCannotReturnItself(AbstractFeature f)
           code(new_code_ref) + "to return a reference.");
   }
 
+  public static void illegalUseOfThisTypeInTypeFeature(DotType d, AbstractFeature outer)
+  {
+    error(d.pos(),
+          "Illegal use of " + code(".this.type") + " inside of type feature " + s(outer) + ".",
+          "To solve this, remove the offending code.");
```
michaellilltokiwa, line 1771: Needs better description.

src/dev/flang/ast/AstErrors.java
```diff
@@ -1764,6 +1764,20 @@ public static void routineCannotReturnItself(AbstractFeature f)
...
+  }
+
+  public static void featureRequiringInstanceCalledOnDotType(Call c, AbstractFeature cf)
+  {
+    error(c.pos(),
+          "Called feature " + s(cf) + " needs an instance, but was called like a type feature.",
+          "To solve this, remove the offending code.");
```
michaellilltokiwa, line 1778: Needs better description.

src/dev/flang/ast/DotType.java
```diff
@@ -108,14 +108,37 @@ public Expr visit(FeatureVisitor v, AbstractFeature outer)
...
+  /**
+   * @return Are we within a type feature?
+   */
+  private boolean withinTypeFeature(AbstractFeature f)
+  {
+    return f == null
+      ? false
+      : f.isTypeFeature()
+      ? true
+      : withinTypeFeature(f.outer());
```
fridis, line 141: A `? :` should never be used if one of the branches is `true` or `false`. The following is equivalent
```
  return f != null && (f.isTypeFeature() || whthinTypeFeature());
```

src/dev/flang/ast/Call.java
```diff
@@ -717,6 +725,17 @@ else if (_calledFeature == null) // nothing found, so flag error
   }
 
 
+  /**
+   * Does fo require an instance to be called?
```
fridis, line 729: I needed a while to understand what this means, the comment just repeats the method name. This basically checks that neither fo._feature nor fo_outer are type features.

fridis (changes requested):
I think the problem is somewhere else: When resolving the call to `T.type.frac`, the target  of the call is `T.type`, which should be of type `Type`, so `frac` should not be found and result in an error `Could not find called feature`, just as `T.frac` does: 
```
/home/fridi/fuzion/work/ex1055.fz:5:11: error 1: Could not find called feature
    say T.fract
----------^
Feature not found: 'fract' (no arguments)
Target feature: 'float.type'
In call: 'c.this.T.fract'
```
So, the question is: why does `Call.loadCalledFeature` find `frac`? Probably it has the wrong target type, so why is the target type of `T.type` wrong?

PR #1383, opened 2023-04-24
fe: execute universe's code when single field declaration given to `fz`

### LINKED ISSUE #1380: `Main feature must not be a field` is a bit confusing
if the code presented to `fz` contains several fields, it is fine

```
 > echo "a := 3; b := 4" | ./build/bin/fz -
```
If it contains only one field, we get this error:
```
 > echo "a := 3" | ./build/bin/fz -

-:1:1: error 1: Main feature must not be a field
a := 3
^
Main feature must be a non-abstract non-intrinsic routine
To solve this, use a non-abstract, non-intrinsic, non-generic routine as the main feature of your application.

one error.
```
The reason is that if the code presented to `fz` is a single feature declaration, that feature's code will be executed. Otherwise, the universe's code will be executed. Now, if there is a single feature that is a field, there is no code, so we get this error. 
Would de better to change the semantics such that universe's code will also be executed if the code presented to `fz` is a single field declaration.

maxteufel:
Fixes #1380.

### DIFF
--- a/src/dev/flang/fe/SourceModule.java
+++ b/src/dev/flang/fe/SourceModule.java
@@ -164,7 +164,7 @@ public class SourceModule extends Module implements SrcModule, MirModule
    * If source comes from stdin or an explicit input file, parse this and
    * extract the main feature.  Otherwise, return the default main.
    *
-   * @return the main feature found or null if none
+   * @return the main feature found or _defaultMain if none
    */
   String parseMain()
   {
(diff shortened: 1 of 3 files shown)


src/dev/flang/fe/SourceModule.java
```diff
@@ -181,7 +181,15 @@ String parseMain()
                 f.legalPartOfUniverse();  // suppress FeErrors.initialValueNotAllowed
                 if (stmnts.size() == 1)
                   {
-                    res =  f.featureName().baseName();
+                    if (f.kind() == Feature.Kind.Field)
+                      {
+                        // execute the universe's code in this case, to avoid FeErrors.mainFeatureMustNotBeField
+                        res = null;
```
fridis, line 187: would be better to keep `res` unchanged in this case, i.e., return `_defaultMain`. The comment `@return...` should also be fixed to reflect that this returns `_defaultMain` (even if, as far as I can see, this should be `null`.

src/dev/flang/fe/SourceModule.java
```diff
@@ -181,7 +181,15 @@ String parseMain()
                 f.legalPartOfUniverse();  // suppress FeErrors.initialValueNotAllowed
                 if (stmnts.size() == 1)
                   {
-                    res =  f.featureName().baseName();
+                    if (f.kind() == Feature.Kind.Field)
```
fridis, line 184: better use `f.isField()`.

fridis (changes requested):
Just minor comments to simplify the code and improve maintainability.

fridis (approved):

PR #1388, opened 2023-04-25
parser: make `infix :` right-associative

maxteufel:
This makes it easier to create lists using `list.infix :` because it is now possible to declare `a list i32 := 1:2:3:nil` instead of `a list i32 := 1:(2:(3:nil))`.

The big problem with this change is that `infix :` is also used for boolean implication. In this case right-associativity is unwanted.

### DIFF
--- a/src/dev/flang/parser/OpExpr.java
+++ b/src/dev/flang/parser/OpExpr.java
@@ -458,7 +458,7 @@ public class OpExpr extends ANY
   /**
    * Initial characters of operators that bind to the right.
    */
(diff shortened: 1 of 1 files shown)


fridis:
Alternatively, the syntax sugar performed in ParsedCall could check that for an `infix_right`, if the target is also an `infix_right` operand with same precedence and not in parentheses, then change `a.f(b).g(c)` into `a.f(b.g(c))`. This could use some of the infrastructure used for chained booleans.

PR #1395, opened 2023-04-27
add tool to measure coverage

michaellilltokiwa:
this makes it easier to see which error message are not hit by the tests.
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,12 @@ CLASS_FILES_MISC_LOGO         = $(CLASSES_DIR)/dev/flang/misc/logo/__marker_for_
 JFREE_SVG_URL = https://repo1.maven.org/maven2/org/jfree/org.jfree.svg/5.0.1/org.jfree.svg-5.0.1.jar
 JARS_JFREE_SVG_JAR = $(BUILD_DIR)/jars/org.jfree.svg-5.0.1.jar
 
+# coverage tool
+JARS_JACOCO_AGENT_URL = https://repo1.maven.org/maven2/org/jacoco/org.jacoco.agent/0.8.10/org.jacoco.agent-0.8.10.jar
+JARS_JACOCO_AGENT = $(BUILD_DIR)/jars/org.jacoco.agent-0.8.10.jar
+JARS_JACOCO_CLI_URL = https://repo1.maven.org/maven2/org/jacoco/org.jacoco.cli/0.8.10/org.jacoco.cli-0.8.10-nodeps.jar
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1079,3 +1085,18 @@ show_release_notes:
...
+$(JARS_JACOCO_AGENT):
+	mkdir -p $(@D)
+	curl $(JARS_JACOCO_AGENT_URL) --output $@
+	unzip $(JARS_JACOCO_AGENT) jacocoagent.jar -d $(BUILD_DIR)/jars
+
+$(JARS_JACOCO_CLI):
+	mkdir -p $(@D)
+	curl $(JARS_JACOCO_CLI_URL) --output $@
+
+coverage: $(FZ_INT) $(FZ_C) $(MOD_TERMINAL) $(MOD_JAVA_BASE) $(BUILD_DIR)/tests $(JARS_JACOCO_AGENT) $(JARS_JACOCO_CLI)
```
fridis, line 1098: Please add a comment to explain what this does: running the tests both interpreted and with the C backend to collect coverage date using jacoco and then create html to show the results in a browser. 

In a later step, we might want to enable this via an option or env var given to `fz` such that we can easily collect coverage data.

Makefile
```diff
@@ -1079,3 +1085,18 @@ show_release_notes:
...
+	(FUZION_JAVA_OPTIONS="-javaagent:../../jars/jacocoagent.jar=destfile=../../jacoco.exec" $(FZ_SRC)/bin/run_tests.sh $(BUILD_DIR) int) || true
```
fridis, line 1099: The use of relative paths `../../jars/..` here is maybe a little brittle. Does `$(BUILD_DIR)/jars` work better?

fridis (changes requested):
Does it make sense to use maven to get these jars? I do not have good memories of using maven, I was always confused about where code came from, but this might be my fault not knowing maven well enough.

Makefile
michaellilltokiwa: @fridis `$(BUILD_DIR)/jars` is a relative path as well, I think? Maybe we could do somethink like `$(realpath $(BUILD_DIR)/jars)`. But is this less brittle? I don't know.

michaellilltokiwa:
> Does it make sense to use maven to get these jars? I do not have good memories of using maven, I was always confused about where code came from, but this might be my fault not knowing maven well enough.

I don't know. An advantage we get by using maven is support by tooling to see vulnerable / outdated dependencies. Downside is one more dependency: `maven`.

PR #1402, opened 2023-05-02
switch tests from negative to simple tests

michaellilltokiwa:
#1408 should be merged first


fridis (approved):
Can you go through all the results and create issues for unexpected errors, I think I have seen some.

PR #1406, opened 2023-05-02
lib/outcome: change `val`, `err` to use try-effect instead of panic

michaellilltokiwa:
This PR is a suggestion.
Similar changes could be done for `Sequence.first`, `Sequence.last` or even `array. index[]`.

### DIFF
--- a/lib/outcome.fz
+++ b/lib/outcome.fz
@@ -77,15 +77,13 @@ is
 
   # value of an outcome that is known to contain a value
   #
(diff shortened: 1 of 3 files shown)


fridis:
My suggestion is to create new issues:

- integrate pre/postcondition and try effect
- make try effect extensible to allow different exception kinds similar to Java's try-catch
- make panic such a specialized try effect (as in Java's Error or ThreadDeath throwables). 

and close this PR.

PR #1411, opened 2023-05-03
remove keyword `redefine`

michaellilltokiwa:
Personally, I like `redefine` more than `redef` but I think I'm alone here?


fridis:
I like short keywords :-).

fridis (changes requested):
Can you solve the merge conflic.

PR #1422, opened 2023-05-10
be/c: Intrinsics: access static fields in a static way

maxteufel:


src/dev/flang/be/c/Intrinsics.java
```diff
@@ -888,7 +888,6 @@ public class Intrinsics extends ANY
 
   private static void put(String n, IntrinsicCode c) { _intrinsics_.put(n, c); }
   private static void put(String n1, String n2, IntrinsicCode c) { put(n1, c); put(n2, c); }
-  private static void put(String n1, String n2, String n3, IntrinsicCode c) { put(n1, c); put(n2, c); put(n3, c); }
```
michaellilltokiwa, line 891: We may use this in the future, I think it' reasonable to keep this around?

michaellilltokiwa (approved):

PR #1428, opened 2023-05-12
lib/io/file: add `using_file`,  `open_file`-effect

michaellilltokiwa:
Example when using `using_file` with type parameter:
```
ex is

  s :=
    io.file.using_file nil String "/tmp/test" io.file.mode.read ()->
      s2 :=
        io.file.using_file unit String "/tmp/test2" io.file.mode.read ()->
          (io.file.open_file nil).file_name
      s3 :=
        io.file.using_file nil String "/tmp/test3" io.file.mode.read ()->
          (io.file.open_file nil).file_name
      "{(io.file.open_file nil).file_name}, {s2.val}, {s3.val}"

  say s
```
```
/tmp/test, /tmp/test, /tmp/test3
```

### DIFF
--- a/lib/io/file/open.fz
+++ b/lib/io/file/open.fz
@@ -17,7 +17,7 @@
 #
 #  Tokiwa Software GmbH, Germany
 #
-#  Source code of Fuzion standard library feature open_file
+#  Source code of Fuzion standard library feature io.file.open
 #
 # -----------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


fridis (changes requested):
Nice, this reminds me of the way koka uses effects to access files avoiding the need to pass `fd` around. 

When naming features, we should avoid repeating what is clear from the outer features, i.e., `io.file.open_file` or `io.file.using_file` could just be `io.file.open` and `io.file.using` or `io.file.use`. This avoids redundancy and typing.

fridis (approved):

PR #1429, opened 2023-05-12
fz: add option `-no-backend`

michaellilltokiwa:
only frontend is created then shows errors and exits.

This could be used for syntax checking of idioms/examples etc.

### DIFF
--- a/src/dev/flang/tools/Fuzion.java
+++ b/src/dev/flang/tools/Fuzion.java
@@ -264,6 +264,19 @@ class Fuzion extends Tool
       }
     },
 
+    /**
+     * This backend does nothing except showing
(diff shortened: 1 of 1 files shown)


src/dev/flang/tools/Fuzion.java
```diff
@@ -264,6 +264,14 @@ void processFrontEnd(Fuzion f, FrontEnd fe)
       }
     },
 
+    noBackend("-no-backend")
```
fridis, line 267: please add a comment describing what this is good for.

fridis (changes requested):
Good idea!

Just a request for documentation.

fridis (approved):

PR #1440, opened 2023-05-15
lib: group complex, fraction and matrix feature in num outer feature

maxteufel:


tests/covariance/test_covariance.fz
```diff
@@ -28,54 +28,54 @@
 #
 test_covariance is
 
-  # a simple, numeric type providing special elements zero and one and operations
+  # a simple, num0eric type providing special elements zero and one and operations
```
michaellilltokiwa, line 31: num0eric

PR #1457, opened 2023-05-16
lib: Add NYI-comment to simple_effect.use

fridis:
There is currently no way to detect an `abort` that occured in a simple_effect. We should probably change this.

Also replaced var name in lambda from `unit` to `_`.

### DIFF
--- a/lib/effect.fz
+++ b/lib/effect.fz
@@ -153,6 +153,9 @@ is
 
   # install this simple_effect and run code
   #
+  # In case of an `abort`, `use` returns silently (NYI: maybe this should better
+  # return an oucome with the abort wrapped in an error?).
+  #
   use(code ()->unit) unit is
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
I was thinking about this as well. I came up with this.
Is this similar to what you had in mind?

```
less_simple_effect(R type) : simple_effect
is

  # mutuable field to hold result on abort
  #
  private aborted_with option R := nil


  abort_with(r R) =>
    set aborted_with := r
    abort

  # install this effect and run code that produces a result of
  # type T. On abort a result R is produced.
  #
  go_abortably(T type, f ()->T) R | T is
    fo () -> option T is ()->f()
    r := run fo (unit)->nil
    match r
      nil => aborted_with.get
      v T => v
```

PR #1468, opened 2023-05-17
lib: remove comments NYI: #167: 'ref' due to lack of 'like this'

michaellilltokiwa:
I am not sure what is desirable here? The way I have done it in this PR:
ps_set.add returns a `ref ps_set` because it implements abstract method add from `Set`
ps_map.add returns a value and can do that because `Map` does not have `add`.

### DIFF
--- a/examples/javaGraphics/gfx.fz
+++ b/examples/javaGraphics/gfx.fz
@@ -100,12 +100,12 @@ is
   sin240 := -866
 
 
-  G2D (g Java.java.awt.Graphics, tx, ty, rot0 i32) ref is  # NYI: #167: 'ref' due to lack of 'like this'
+  G2D (g Java.java.awt.Graphics, tx, ty, rot0 i32) is
     rot := if rot0 >= 0 rot0 % 360 else rot0 % 360 + 360
 
     rotate (d i32) => G2D g tx ty rot+d
 
-    threeTimes(d (G2D, Java.java.awt.Color) -> unit) unit is
+    threeTimes(d (ref G2D, Java.java.awt.Color) -> unit) unit is
       d  G2D.this              C1
       d (G2D.this.rotate -120) C2
       d (G2D.this.rotate -240) C3
--- a/lib/ps_map.fz
+++ b/lib/ps_map.fz
@@ -316,14 +316,14 @@ O(n log n).
   # both ps_map.this and other, it will be mapped to ps_map.this[k] or to other[k],
   # but it is undefined to which of these two.
   #
-  union (other ref ps_map PK V) ref ps_map PK V is  # NYI: #167: 'ref' due to lack of 'like this'
+  union (other ps_map PK V) ps_map PK V is
     if other.size < size
-      other ∪ ps_map.this
+      other ∪ (ps_map PK V data size fill)
     else
 
       # helper to add 'ps_map.this.data[l..r]' to 'a' and return the resulting map.
       #
-      add_all (a ref ps_map PK V, l i32, r i32) ref ps_map PK V is  # NYI: #167: 'ref' due to lack of 'like this'
+      add_all (a ps_map PK V, l i32, r i32) ps_map PK V is
         if l > r
           a
         else
@@ -335,7 +335,7 @@ O(n log n).
       # 'tail' elements at indices larger or equal to 'at+sz' added recursively as
       # well.
       #
-      add_all (a ref ps_map PK V, at i32, sz i32, tail i32) ref ps_map PK V is  # NYI: #167: 'ref' due to lack of 'like this'
+      add_all (a ps_map PK V, at i32, sz i32, tail i32) ps_map PK V is
         if sz = 0
           a
         else
@@ -355,7 +355,7 @@ O(n log n).
 
   # infix operator synonym for union of two ps_maps
   #
-  infix ∪ (other ref ps_map PK V) => union other  # NYI: #167: 'ref' due to lack of 'like this'
+  infix ∪ (other ps_map PK V) => union other
 
 
 # ps_maps -- unit type feature declaring features related to ps_map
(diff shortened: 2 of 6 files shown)


tests/reg_issue25_unbox_outerref/unboxOuterRef.fz
```diff
@@ -34,11 +34,11 @@ unboxOuterRef is
 
   y (i i32) is
     redef as_string => "y $i"
-    g (v y) is
+    g (v ref y) is
```
fridis, line 37: instead of `ref y`, my idea here was that this should be `y.this.type`. But then it can no longer be called as in `g (y i)`. 

Would maybe be good to have three cases here
```
  gy (v y) is...
  gry (v ref y) is...
  gyt (v y.this.type) is...
```
and then call all of these
```
  // gy y.this   -- incomptible types
  gy (y i) 
  gry y.this
  gry (y i)
  gyt y.this
  // gyt (y i) -- incompatible types
```

lib/ps_set.fz
```diff
@@ -36,7 +36,7 @@
 #
 ps_set
   (K type : has_total_order,
-   psm ref ps_map K unit,  # NYI: #167: 'ref' due to lack of 'like this'
+   psm ps_map K unit,
```
fridis, line 39: My original idea was to use `psm ps_set.this.type`, not sure if this works, though.

lib/ps_set.fz
```diff
@@ -60,7 +60,7 @@ is
 
   # add new element k to this set.
   #
-  add (k K) ref ps_set K is  # NYI: #167: 'ref' due to lack of 'like this'
+  add (k K) ref ps_set K is
```
fridis, line 63: The idea was to have 
```
  add (k K) ps_set.this.type is
```
but then the result in the `else` branch is incompatible. I am not sure if making this `fixed` works, but it probably should:
```
  fixed add (k K) ps_set.this.type is
```
This would prevent inheriting `add`, i.e., in an heir feature, `add` would be abstract unless explicitly implemented.

lib/ps_map.fz
```diff
@@ -316,14 +316,14 @@ O(n log n).
   # both ps_map.this and other, it will be mapped to ps_map.this[k] or to other[k],
   # but it is undefined to which of these two.
   #
-  union (other ref ps_map PK V) ref ps_map PK V is  # NYI: #167: 'ref' due to lack of 'like this'
+  union (other ps_map PK V) ps_map PK V is
     if other.size < size
-      other ∪ ps_map.this
+      other ∪ (ps_map PK V data size fill)
```
fridis, line 321: Ideally, `ps_map.this` should be usable when the argument type if `ps_map.this.type`.

examples/javaGraphics/gfx.fz
```diff
@@ -100,12 +100,12 @@ is
...
 
 
-  G2D (g Java.java.awt.Graphics, tx, ty, rot0 i32) ref is  # NYI: #167: 'ref' due to lack of 'like this'
+  G2D (g Java.java.awt.Graphics, tx, ty, rot0 i32) is
     rot := if rot0 >= 0 rot0 % 360 else rot0 % 360 + 360
 
     rotate (d i32) => G2D g tx ty rot+d
 
-    threeTimes(d (G2D, Java.java.awt.Color) -> unit) unit is
+    threeTimes(d (ref G2D, Java.java.awt.Color) -> unit) unit is
```
fridis, line 108: I would like to have this: 
```
 fixed rotate (d i32) G2D.this.type is G2D g tx ty rot+d
 threeTimes(d (G2D.this.type, Java.java.awt.Color) -> unit) unit is
```
But `fixed` does not work like this yet. I will create an issue for that.

fridis: Trying to add an issue, I realized this `fixed` actuall works already, see this example: 
```
test_fixed is

  A is
    fixed me A.this.type is A.this
    redef as_string => "A"

  B : A is
    fixed me => B.this
    redef as_string => "B"

  say A.me
  say B.me
```
that produces
```
> ./build/bin/fz test_fixed2.fz 
A
B
```
Can you try to change `rotate` to be `fixed` with result type `G2D.this.type` and change the first argument type of `threeTimes.d` to `G2d.this.type`?

fridis: I played aroudn with this a bit more to make sure that `B.me` is actually called, and it is:

```
test_fixed is

  A is
    fixed me A.this.type is A.this
    mer ref A is me
    redef as_string => "A"

  B(s String) : A is
    fixed me => (B "new")
    redef as_string => "B$s"

  say A.me
  say (B "orig").me

  boxed(r ref A) => say r.mer
  boxed A
  boxed (B "orig")
```
results in
```
 > ./build/bin/fz test_fixed2.fz 
A
Bnew
A
Bnew
```
Note that a call to `r.me` within `boxed` results in an error due to ambiguous result types.

fridis: I was wrong, `fixed` is not powerful enough for this yet, the following example fails

```
test_fixed is

  A is
    fixed me A.this.type is A
    redef as_string => "A"

  B(s String) : A is
    fixed me => (B "new")
    redef as_string => "B$s"

  say A.me
  say (B "orig").me
```
as follows
```
 > ./build/bin/fz test_fixed4.fz 

/home/fridi/fuzion/clean/fuzion/test_fixed4.fz:4:29: error 1: Incompatible types in assignment
    fixed me A.this.type is A
----------------------------^
assignment to field : 'test_fixed.A.me.result'
expected formal type: 'test_fixed.this.type.A.this.type'
actual type found   : 'test_fixed.this.type.A'
assignable to       : 'test_fixed.this.type.A'
for value assigned  : 
{
  box(deref(deref(me.this.#^test_fixed.A.me).#^test_fixed.A).A)
}

To solve this, you could change the type of the target 'test_fixed.A.me.result' to 'test_fixed.this.type.A' or convert the type of the assigned value to 'test_fixed.this.type.A.this.type'.

one error.
```

Issue #1477, opened 2023-05-22
Small example using `this.type` crashes interpreter and C backend


fridis:
This example
```
n is

  type.z n.this.type is abstract

  u unit is
    y := n.this.type
    q := y.z

i : n is
  fixed type.z => i

i.u
```
results in a crash, both in the interpreter as well as when using the C backend:
```
 > ./build/bin/fz test_covariance.fz

error 1: abstract feature n.type.z called on lvalue[instance[i.u]557041912@0(n.#type i)] of clazz n.#type i
Call stack:
i.u: /home/fridi/fuzion/clean/fuzion/test_covariance.fz:7:12:
    q := y.z
-----------^
universe#0: /home/fridi/fuzion/clean/fuzion/test_covariance.fz:12:3:
i.u
--^


*** fatal errors encountered, stopping.
 > ./build/bin/fz -c test_covariance.fz

/home/fridi/fuzion/clean/fuzion/test_covariance.fz:7:12: error 1: NYI: in i.u no targets for Call to (n.#type i).z target UNIT
    q := y.z
-----------^

one error.
```
Wrapping the code into a feature `t` avoids the problem:
```
t is

  n is

    type.z n.this.type is abstract

    u unit is
      y := n.this.type
      q := y.z

  i : n is
    fixed type.z => t.i

t.i.u
 > ./build/bin/fz test_covarianceA.fz
 > ./build/bin/fz -c test_covarianceA.fz
 > ./universe 
 > 
```

PR #1499, opened 2023-05-24
tests: add regression test for issue #1477

maxteufel:


fridis (approved):
This is somewhat covered by tests/covariance, which is why I did not add this. But it might help and does not hurt much to have this explicitly as well.

PR #1506, opened 2023-05-24
numeric: move features from plural features (numerics, i32s, …) to type features

maxteufel:
This should make it easier to eventually drop the type parameter from `numeric`.


lib/num/matrix.fz
```diff
@@ -27,81 +27,75 @@
...
-#
-matrices(M type : numeric M) : numerics (matrix M) is
 
-  # NYI: There is currently no easy way to get hold of an instance of the unit
-  # type corresponding to E.
+  # identity element for 'infix *'
   #
-  #   redef zero => matrix E.zero E.zero
-  #   redef one  => matrix E.one  E.zero
+  # note that this is only an identity element if we are working
```
michaellilltokiwa, line 97: ...only an identity element for square matrices.

lib/num/complex.fz
```diff
@@ -31,45 +31,45 @@
...
-# private:
-
-  # just for brevity
-  a => complex real imag
-
-# public:
 
   # enable generic features in ancestors
-  redef thiz => a
+  redef thiz => complex real imag
```
michaellilltokiwa, line 39: Can we get rid of `thiz`? (Word play not intended :-D)

michaellilltokiwa (changes requested):

fridis (approved):
looks good!

michaellilltokiwa (approved):

PR #1515, opened 2023-05-24
fe: Allow redefinition of feature with type parameters, fix #1387

### LINKED ISSUE #1387: Permit redefinition of feature with type parameters
There is no real reason to forbid redefinition of features with type parameters except that this needs to be supported by the implementation, in particular dynamic binding must be able to handle multiple different actual type parameters. 

Here is an example

```
x is
  q(T type) => say "x: $T"

y : x is
  redef q(T type) => say "y: $T"

x.q i32
y.q f64
```
that currently produces an error:
```
 > ./build/bin/fz redeftp.fz

/home/fridi/fuzion/work/redeftp.fz:5:3: error 1: Cannot redefine feature with type parameters
  redef q(T type) => say "y: $T"
--^
Feature that redefines existing feature: 'y'
original feature: 'x.q'
original feature defined in /home/fridi/fuzion/work/redeftp.fz:2:3:
To solve this, ask the Fuzion team to remove this restriction :-).

one error.
```

fridis:
This does allow the execution of the example given in #1387.  The example is added to a new test case.

However, the new test case also add three variants that currently do not work yet.  I will add separate issues for those to get fixed.  The main problem is types used in the signature that depend on type parameters of the redefined feature.

### DIFF
--- a/tests/redef_with_type_parameters/redef_test.fz
+++ b/tests/redef_with_type_parameters/redef_test.fz
@@ -75,8 +75,8 @@ one error.
 
   /* NYI: redefinition using types that use other type parameters does not work yet:
 
-  # thrid case, redefining feature x.q whose type parameter has a
-  # constraint depending same feature's type parameter.
+  # third case, redefining feature x.q whose type parameter has a
+  # constraint depending on the same feature's type parameter.
   #
   case3 =>
     x is
--- a/tests/redef_with_type_parameters/redef_test.fz
+++ b/tests/redef_with_type_parameters/redef_test.fz
@@ -1,0 +1,113 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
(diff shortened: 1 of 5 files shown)


tests/redef_with_type_parameters/redef_test.fz
```diff
@@ -0,0 +1,113 @@
...
+    y.q (list C)
+
+  case2 String "str"
+  case2 f32 3.14
+
+  */
+
+  /* NYI: redefinition using types that use other type parameters does not work yet:
+
+  # thrid case, redefining feature x.q whose type parameter has a
```
michaellilltokiwa, line 78: third

tests/redef_with_type_parameters/redef_test.fz
```diff
@@ -0,0 +1,113 @@
...
+  # constraint depending same feature's type parameter.
```
michaellilltokiwa, line 79: depending on

michaellilltokiwa (approved):

PR #1521, opened 2023-05-25
mutable_tree_map, Entry: inherit from Map

michaellilltokiwa:
When inheriting from Map I had to redef `K`,`V` but `K` has constraint `has_total_order` so it is not possible to use `redef`.
Now we have `KEY` and `V` which is not pretty...
What do you think? I see these options:
1) keep it like this
2) rename `V` as `VAL` or `VALUE`
3) use different names such as MTM_K, MTM_V

```
ex is

  m : mutate is

  m.go (()->
    map := container.mutable_tree_map m i32 String
    map.put 1 "Hello"
    map.put 2 "World"
    say (map.items.map_sequence (kv)->
      (k,v) := kv
      "$k = $v"
    )
  )
```
```
[1 = Hello,2 = World]
```

### DIFF
--- a/lib/container/Entry.fz
+++ b/lib/container/Entry.fz
@@ -24,17 +24,17 @@
 
 # Entry -- an entry of a mutable tree map
 #
-private Entry(LM type : mutate, K type : has_total_order, V type, key K, val V) ref is
+private Entry(LM type : mutate, KEY type : has_total_order, redef V type, key KEY, val V) ref : Map KEY V is
 
 
   # reference to the left subtree at this entry, or, if it is empty, nil
   #
-  private left := LM.env.new (option (Entry LM K V)) nil
+  private left := LM.env.new (option (Entry LM KEY V)) nil
 
 
   # reference to the right subtree at this entry, or if it is empty, nil
   #
-  private right := LM.env.new (option (Entry LM K V)) nil
+  private right := LM.env.new (option (Entry LM KEY V)) nil
 
 
   # height of the subtree whose root is this entry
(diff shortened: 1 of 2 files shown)


lib/container/mutable_tree_map.fz
```diff
@@ -34,13 +34,27 @@ mutable_tree_map(LM type : mutate, K type : has_total_order, V type) is
...
+  #
+  index [] (k KEY) option V is
+    get k
+
+
+  # get a sequence of all key/value pairs in this map
+  #
+  items Sequence (tuple KEY V) is
+    match root.get
+      e Entry LM KEY V => e.items
```
fridis, line 56: I think you can omit `LM KEY V` here.

lib/container/Entry.fz
```diff
@@ -45,7 +45,7 @@ private Entry(LM type : mutate, K type : has_total_order, V type, key K, val V)
   # get the value stored in this submap at key k, nil if k is not a key
   # in this submap
   #
-  private get(k K) option V is
+  private get(k KEY) option V is
     if k < key
       left.get ? nil => nil
                | e Entry => e.get k
```
fridis, line 51: Can this be written as
```
    left.get >>= (e -> e.get k)
```
?

fridis (approved):
Renaming is ok. Can you add an issues that we might consider permitting strengthening of constraints of inherited and redefined type parameters, I do not see any obvious reasons not to allow this. 

Two comments for minor improvements you can try in a new PR.

michaellilltokiwa:
closing in favor of  #1534

PR #1522, opened 2023-05-25
lib: add feature memoize

michaellilltokiwa:
Inspired by idiom 315: https://www.programming-idioms.org/idiom/315/memoization

Would be cool to be able to have this as a subfeature of Unary. So we could do `fn.memoize` to get a memoized version of a Unary.
But the `T` of memoize is constrained while Unary's `T` is not.

### DIFF
--- a/lib/memoize.fz
+++ b/lib/memoize.fz
@@ -50,7 +50,7 @@
 # NYI this should be map implementation
 # agnostic via map effect. currently using mutate and ordered_map.
 #
-# NYI constraint for T should be equatable
+# NYI constraint for T should be has_total_order, has_hashcode, or has_ordinal
 #
 memoize(LM type : mutate, T type : has_total_order, R type, f T->R) T->R is
   ref : Unary(R, T)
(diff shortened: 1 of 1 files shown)


lib/memoize.fz
```diff
@@ -0,0 +1,67 @@
...
+#   say (mem 2)
+#   say (mem 1)
+#   say (mem 3)
+# )
+# ```
+#
+# NYI this should be map implementation
+# agnostic via map effect. currently using mutate and ordered_map.
+#
+# NYI constraint for T should be equatable
```
fridis, line 53: I think `equatable` is not sufficient as a constraint. I can only imagine two reasonable constraints: `has_total_order` and `has_hashcode`, or maybe one third `has_ordinal` for something that maps to a number 0, 2, 3, etc. that can be used as an index in an array.

fridis (changes requested):
Very nice, we need this!
Soon, we might want to use tuples to memoize functions with more arguments. So far, I do not have a good idea for how do do this...
Just maybe change the comment asking for equatable constraint.

fridis (approved):

PR #1529, opened 2023-05-26
air: Remove redundant `_heirs.add(this)`

fridis:
The same is done in `registerAsHeir()` which is called for every clazz anyway.


src/dev/flang/air/Clazz.java
```diff
@@ -551,7 +551,6 @@ public Set<Clazz> heirs()
     if (_heirs == null)
       {
         _heirs = new TreeSet<>();
```
michaellilltokiwa, line 553: The comment for this method is potentially misleading now.

michaellilltokiwa (approved):

PR #1530, opened 2023-05-26
fe: replace `this` and `t` in `this.applyTypePars(t)`, fix #1527

### LINKED ISSUE #1527: applyTypePars confusion
There are two variants of `applyTypePars` (which was renamed from `actualType` by #1523 ) , one receiving a type, the other a feature and actuals. 
The confusion is that `a.applyTypePars(b)` replaces the type parameters in `b`, while `a.applyTypePars(f,actuals)` replaces the type parameters in `a`. It would be better to be consistent here and have something like `a.applyTypeParsFrom(b)` and `a.applyTypeParsFrom(f, actuals)` that both replace the type parameters of `a`.

fridis:
With this change, `AbstractTypes.applyPars(AbstractType)` and `AbstractTypes.applyPars(AbstractFeature, List<AbstractType>)` both modify the type given as `this`.

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -542,7 +542,7 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
 
 
   /**
-   * Replace generic types used in by this type by the actual types given in
+   * Replace generic types used by this type by the actual types given in
    * target.
    *
    * @param target a target type this is used in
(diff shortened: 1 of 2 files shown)


src/dev/flang/ast/AbstractType.java
```diff
@@ -542,33 +542,33 @@ public boolean dependsOnGenerics()
 
 
   /**
-   * Replace generic types used in given type t by the actual generic arguments
-   * given in this.
+   * Replace generic types used in by this type by the actual types given in
```
michaellilltokiwa, line 545: ...used in by this...

src/dev/flang/ast/AbstractType.java
```diff
@@ -578,35 +578,35 @@ public AbstractType applyTypePars(AbstractType t)
...
```
michaellilltokiwa, line 581: ...used in by this...

michaellilltokiwa (approved):

PR #1553, opened 2023-05-31
lib, array: move initialization to type feature

michaellilltokiwa:


fridis (commented):
It would be good to have this API in sync with that of `marray`, so moving the code to `type.new` is good. 
Neverthelesse, I think we should keep `array T count i->x` as convenience feature to create in array. In particular, we need this for type inference as in
```
  squares := array 100 i->i*i
```

PR #1554, opened 2023-05-31
lib: add io.buffered.reader effect

michaellilltokiwa:
The buffered reader is an effect that stores its buffer between subsequent calls to `read n`, `read_line`, etc.. When an error or end-of-file occurs an abort is raised and we jump back to where the effect was installed. This is done by using the try effect.


lib/io/buffered/reader.fz
```diff
@@ -0,0 +1,223 @@
...
+#
+# -----------------------------------------------------------------------
+
+
+# buffered.reader effect allows buffered reading
+# by using the given Read_Provider
+#
+# note: anything in the buffer when effect is uninstalled will be discarded.
+#
+reader(private rp Read_Provider, buf_size i32, buffer array u8 | io.end_of_file) : try is
```
fridis, line 30: I do not understand why this inherits from `try`. There is one place where `raise` is called, but two other places where `try.env` is used. So it looks to me as if two different types of `try` effects are mixed. Please explain. 

Without inheriting the try effect, my impression is that a buffered reader does not need to be an effect at all. But it does manage a buffer, so it contains mutable fields and a mutable array, so this should IMHO use a type parameter `LM type : mutate` to perform these mutations in a way that allows a local instance of mutate without adding one more effect.

fridis (approved):

PR #1566, opened 2023-06-02
lib: add array.type.new, marray shorthand

michaellilltokiwa:


fridis (commented):
`marray` should be removed now that we have `mutate.array`.

fridis (approved):

PR #1569, opened 2023-06-02
lib, Sequence: remove stream in some features, return list where possible

michaellilltokiwa:
- where `as_list` is used to use the corresponding list feature we can as well return `list` and not `Sequence`.


fridis (approved):
Good. 

In general, it is good to use `list` as a result type of routines, and `Sequence` as an argument type since `list` can be assigned to `Sequence` but now vice versa.

PR #1589, opened 2023-06-13
lib: remove `some`, wrapper for value types

michaellilltokiwa:
I don't think we need this (anymore). When nested choice types were broken this was used as a workaround.


michaellilltokiwa:
Max reminded me that we talked about this already:  #1359
But rethinking about the argument makes me wonder if option (some nil) or similar cases is actually useful.
My argument here is about readability/understandability. 
In case someone wants to use a choice type with none-disjoint types, I think it would be better if one created specific wrapper features instead of just using a generic `some`. 
> But there are still cases where one would need some, e.g., option nil does not work, has to be option (some nil). Not sure if there are other possible uses yet.

fridis:
I have create #1591 that explains one open problem with `option nil`.

Let's remove `some` for now.

fridis (approved):

Issue #1242, opened 2023-03-16
Thread safety, interpreter. `getChoiceTag`


michaellilltokiwa:
When running the test _sockets_tests_ in branch _lib/net_, I sometimes see this ensure condition failure:
```

[1;31merror 1[0m[1m: *** java.lang.Error: ensure-condition failed: dev.flang.air.Clazz:setRuntimeData:1473
Call stack:
fuzion.sys.net0.close: [32m$FUZION/lib/fuzion/sys/net0.fz:150:12:[39m
[34m    res := close0 sd
[33m-----------^[0m
(net.channel sockets_test).close: [32m$FUZION/lib/net/channel.fz:84:33:[39m
[34m          match fuzion.sys.net0.close d
[33m--------------------------------^[0m
net.channel sockets_test: [32m$FUZION/lib/net/channel.fz:97:15:[39m
[34m  (channel T).close
[33m--------------^[0m
net.client sockets_test: [32m$FUZION/lib/net/client.fz:42:7:[39m
[34m      channel T desc true
[33m------^[0m
sockets_test.client: [32m/home/not_synced/openvscode-server-fuzion/vscode-fuzion/fuzion-lsp-server/fuzion/tests/sockets/sockets_test.fz:90:15:[39m
[34m    match net.client sockets_test family protocol host port
[33m--------------^[0m
(sockets_test.#loop2.#loop1.ref #fun0).call: [32m/home/not_synced/openvscode-server-fuzion/vscode-fuzion/fuzion-lsp-server/fuzion/tests/sockets/sockets_test.fz:145:20:[39m
[34m        .map (i -> client f p (localhost f) port i)
[33m-------------------^[0m
(((list i32).map (array String)).ref @1074380833).head: [32m$FUZION/lib/list.fz:175:19:[39m
[34m          head => f c.head
[33m------------------^[0m
(list (array String)).flatMap String: [32m$FUZION/lib/list.fz:189:20:[39m
[34m        match (f c.head).as_list
[33m-------------------^[0m
(((list (array String)).flatMap String).ref @1074384435).tail: [32m$FUZION/lib/list.fz:195:41:[39m
[34m              tail => c2.tail ++ c.tail.flatMap B f
[33m----------------------------------------^[0m
(list String).tail: [32m$FUZION/lib/list.fz:91:29:[39m
[34m              | c Cons => c.tail
[33m----------------------------^[0m
[...]

PR #1607, opened 2023-06-15
be/interpreter: Fix race condition in caching of Callable instances

fridis:
The interpreter replaces inner Clazz instances in Clazz` runtime data by instances of Callable or a String "dyn".  When running multiple threads, this could result in a race resulting in a postcondition failure in Clazz.setRuntimeData.

This patch uses a Java monitor to avoid this race conditions.


michaellilltokiwa (approved):

michaellilltokiwa:
fixes #1242

michaellilltokiwa:
There are other places in the interpreter where `getRuntimeData` is used. Is it possible that this leads to problems as well?

michaellilltokiwa (approved):

Issue #1604, opened 2023-06-14
compare-and-swap does not work for choice types when using C backend


fridis:
This test
```
test_cas0 is

  c : choice large small is

  large(a,b,c,d u64).
  small(x u8).

  s := small 0x11
  l := large 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC
  a := concur.atomic c s

  o := a.compare_and_swap s l
  match o
    _ small => #  ok
    _ large => panic "***expected small***"
  o := a.compare_and_swap l s
  match o
    _ small => panic "***expected large***"
    _ large => #  ok
```
works with the interpreter backend
```
 > ./build//bin/fz test_cas0.fz
 >
```
but fails with the C backend
```
 > ./build//bin/fz -c test_cas0.fz
 > ./test_cas0
*** panic: ***expected large***
```
Here is an even more demanding test that tries to fill the stack with 0xAAAA and 0x5555 and then performs compare-and-swap with choice type values allocated on this stack.
```
test_cas is

  c : choice large small is

  large(a,b,c,d u64).
  small(x u8).


  fill(a,b,c,d,e,f u64, n i32) unit is
    if n > 0
      fill ~a ~b ~c ~d ~e ~f n-1

  s := small 0x11
  l := large 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC
  a := concur.atomic c s

  test(n i32) unit is
    if n > 0
      s1 := small 0x11
      l1 := large 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC 0xCCCC_CCCC_CCCC_CCCC
      o := a.compare_and_swap s1 l1
      match o
        _ small => #  ok
        _ large => panic "***expected small*** at $n"
      o := a.compare_and_swap l1 s1
      match o
        _ small => panic "***expected large*** at $n"
        _ large => #  ok

      test n-1

  N := 100

  say "setup:"
  (fill 0xAAAA_AAAA_AAAA_AAAA
        0xAAAA_AAAA_AAAA_AAAA
        0xAAAA_AAAA_AAAA_AAAA
[...]

PR #1618, opened 2023-06-19
be/c: use `__sync_val_compare_and_swap` builtin in cas intrinsic

maxteufel:


fridis (changes requested):
If I understand the [gcc spec ](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html) correctly, `__sync_val_compare_and_swap`  might work for aribrary types (except for choices as in #1604 or random data in pad bytes).

When you are using `__sync_val_compare_and_swap` , you should also change the implementation of `read0` and `write0` to use memory fences instead of the mutex. This is entering slippery grounds, I think we want the semantics to be similar to Java's semantics for volatile fields accesses defined in JSR 133. [Here](https://gee.cs.oswego.edu/dl/jmm/cookbook.html) is a practical guide how to implement this, essentially we need load fences executed after `read0` and store fences executed before `write0`, but maybe a bit more.

PR #1622, opened 2023-06-21
lib: remove the type parameter from `numeric` and friends

maxteufel:
This is the long-awaited pull request to remove the type parameter from `numeric` and friends (such as `wrapping_integer` and `has_interval`).

Unfortunately some features that were virtually unused in the standard library have caused problems in the process of this pull request and have been removed for now: they are `num.complex`, `num.fraction` and `Sequence.indexed (1 argument)`. The long-term intent is to add them back, but I did not want these features to block this pull request any longer.

One of more buggy parts of this pull request seems to be the `has_interval` feature. This is what causes the problem with `Sequence.indexed`, needs a workaround for `has_interval.infix .. .as_stream` (see separate commit) and requires a type parameter to be specified manually in the `ctrie` and `local_mutate` tests. I also believe it is causing the one remaining test failure.

The `reg_issue118` test is skipped for now. This is because the test includes an inheritance `a : i32 3 is`. This is no longer possible without implementing features marked as `fixed` in `i32` manually. I am not sure what to do about this test.

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -401,15 +401,15 @@ Sequence(T type) ref is
   # every element in the sequence
   #
   indexed list (tuple i32 T) is
-    # indexed 0
-    zip i32 (tuple i32 T) (0..) (a, b -> (b, a))
+    indexed 0
 
 
   # adds an index to every element
   # in the sequence starting at start_idx
   #
-  #indexed(I type : has_interval, start_idx I) list (tuple I T) is
-  #  zip (start_idx..) (a,b -> (b, a))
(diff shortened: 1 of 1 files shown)


michaellilltokiwa (approved):

fridis (commented):
Very nice and big step ahead!

A few requests and comments from me

- please create an issue for the failing type inference in `Sequence.indexed`
- another issue for the apparently redundant redefinitions in `float`
- would IMHO have been better to not remove and re-add `fraction` and `complex` in two patches, this loses git history

PR #1623, opened 2023-06-21
change default `_eraseInternalNamesInLib` to false

michaellilltokiwa:


src/dev/flang/ast/Feature.java
```diff
@@ -1554,12 +1554,13 @@ private void checkChoiceAndAddInternalFields(Resolution res)
         p.calledFeature().checkNoClosureAccesses(res, p.pos());
       }
 
-    _choiceTag = new Feature(res,
-                             _pos,
+    _choiceTag = new Feature(_pos,
                              Visi.PRIVATE,
+                             Consts.MODIFIER_FIXED,
```
fridis, line 1559: The original problem here is that a tag field gets added to `choice` and to any feature that inherits from `choice`. Using `fixed` only masks the error without removing the duplicate field. 
My suggestion: change `FuzionConstants.CHOICE_TAG_NAME` to be just `"tag"` and add `priv int tag := ?` to `choice.fz`. 
If that does not work, alternatively, you could add the `_choiceTag` feature here, but only if `this` is `choice` (i..e, `this == Types.f_choice`, but I am not sure if `Types.f_choice` is set at this point).

fridis (changes requested):
This will definitely facilitate our debugging. 

What effect does this have on the size of `base.fum`? I did erase these names because I had the impression that the names data was really excessive, but maybe this is no longer the case (with all the type features and other things being added now).

src/dev/flang/ast/Feature.java
michaellilltokiwa: > My suggestion: change FuzionConstants.CHOICE_TAG_NAME to be just "tag" and add priv int tag := ? to choice.fz.
Unfortunately there are two checks that make this hard. First that choices do not contain fields, second that choices contain only declarations.

PR #1644, opened 2023-06-26
lib: separate IEEE-like and mathematical floating point comparision

maxteufel:
Previously, f32 and f64 implemented type.equality and type.lteq by using the = and <= intrinsics. This is actually wrong because the = and <= intrinsics compare floating point numbers using IEEE rules, where NaN != NaN, and thus type.equality did not implement a proper equivalence relation.

This reintroduces the comparision infixes local to f32 and f64 that by default shadow the infixes provided by `lib/equals.fz`. If a proper equivalence relation or partial order is still desired, comparisions can still be done using `equals f32 f32.type.NaN f32.type.NaN` or `lteq f64 0.57721 f64.type.NaN`, for example.

The idea for this behavior is from [1].

[1]: https://forums.swift.org/t/comparable-and-floatingpoint-types/16886

### DIFF
--- a/lib/f32.fz
+++ b/lib/f32.fz
@@ -44,22 +44,8 @@ f32(val f32) : float is
 
   # comparision
   #
-  # These infixes compare floating point numbers according to the IEEE
-  # standards. They shadow the infixes provided globally by type.equality
-  # so that someone who expects floating point comparision operators to
-  # behave according to IEEE standards will not be surprised.
-  #
-  # If comparision based on a proper equivalence relation or partial order
-  # is desired, this can still be achieved by using the implementations of
-  # type.equality and type.lteq. These additionally define NaN = NaN and
-  # NaN <= x for all x in f32, respectively.
-  #
-  infix = (other f32) bool is intrinsic
-  infix != (other f32) bool is !(infix = other)
-  infix <= (other f32) bool is intrinsic
-  infix >= (other f32) bool is other <= val
-  infix > (other f32) bool is !(infix <= other)
-  infix < (other f32) bool is !(other <= val)
+  fixed infix = (other f32) bool is intrinsic
+  fixed infix <= (other f32) bool is intrinsic
 
 
   # conversion
@@ -165,18 +151,6 @@ f32(val f32) : float is
   # identity element for 'infix *'
   #
   fixed type.one  f32 is 1
-
-
-  # equality
-  #
-  fixed type.equality(a, b f32) bool is
-    a.infix = b || (is_NaN a && is_NaN b)
-
-
-  # total order
-  #
-  fixed type.lteq(a, b f32) bool is
-    a.infix <= b || is_NaN a
 
 
   fixed type.bytes => 4
--- a/lib/f32.fz
+++ b/lib/f32.fz
@@ -42,6 +42,26 @@ f32(val f32) : float is
   fixed infix ** (other f32) f32 is intrinsic
 
 
+  # comparision
+  #
+  # These infixes compare floating point numbers according to the IEEE
+  # standards. They shadow the infixes provided globally by type.equality
+  # so that someone who expects floating point comparision operators to
+  # behave according to IEEE standards will not be surprised.
+  #
+  # If comparision based on a proper equivalence relation or partial order
+  # is desired, this can still be achieved by using the implementations of
+  # type.equality and type.lteq. These additionally define NaN = NaN and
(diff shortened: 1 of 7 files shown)


michaellilltokiwa (approved):

lib/f32.fz
```diff
@@ -149,12 +169,14 @@ f32(val f32) : float is
 
   # equality
```
fridis, line 170: Can you please add a comment that this provides an equivalence relation in the mathematical sense and therefore breaks the IEEE-semantics. `infix =` or `infix !=`  should be used for IEEEE semantics.

lib/f32.fz
```diff
@@ -149,12 +169,14 @@ f32(val f32) : float is
...
   #
-  fixed type.equality(a, b f32) bool is intrinsic
+  fixed type.equality(a, b f32) bool is
+    a.infix = b || (is_NaN a && is_NaN b)
 
 
   # total order
```
fridis, line 176: Can you please add a comment that this provides a total order in the mathematical sense and therefore breaks the IEEE-semantics. `infix <=` etc.  should be used for IEEEE semantics.

lib/f32.fz
```diff
@@ -149,12 +169,14 @@ f32(val f32) : float is
 
   # equality
   #
-  fixed type.equality(a, b f32) bool is intrinsic
+  fixed type.equality(a, b f32) bool is
```
fridis, line 172: I think this can now be moved to `float`.

lib/f32.fz
```diff
@@ -149,12 +169,14 @@ f32(val f32) : float is
...
+    a.infix = b || (is_NaN a && is_NaN b)
 
 
   # total order
   #
-  fixed type.lteq(a, b f32) bool is intrinsic
+  fixed type.lteq(a, b f32) bool is
```
fridis, line 178: I think this can now be moved to `float`.

lib/f64.fz
```diff
@@ -162,12 +182,14 @@ f64(val f64) : float is
...
-  fixed type.equality(a, b f64) bool is intrinsic
+  fixed type.equality(a, b f64) bool is
+    a.infix = b || (is_NaN a && is_NaN b)
 
 
   # total order
   #
-  fixed type.lteq(a, b f64) bool is intrinsic
+  fixed type.lteq(a, b f64) bool is
+    a.infix <= b || is_NaN a
```
fridis, line 192: dto, see comments for `f32`

fridis (changes requested):
nice. 

Just a request to add more comments and to move code to `float` if possible.

lib/float.fz
```diff
@@ -239,3 +254,23 @@ float : numeric is
...
     # 1/2*ln((1+x)/(1-x))
     log ((one + val)/(one - val)) / two
+
+
+  # equality
+  #
+  # This provides an equivalence relation in the mathematical
+  # sense and therefore breaks the IEEE semantics. infix = should
+  # be used for IEEE semantics.
+  #
```
fridis, line 264: maybe would be good to document that `NaN = NaN` here.

lib/float.fz
```diff
@@ -239,3 +254,23 @@ float : numeric is
...
+  type.equality(a, b float.this.type) bool is
+    a.infix = b || (is_NaN a && is_NaN b)
+
+
+  # total order
+  #
+  # This provides a total order in the mathematical sense and
+  # therefore breaks the IEEE semantics. infix <= should be
+  # used for IEEE semantics.
+  #
```
fridis, line 274: maybe would be good to document that `NaN` is less than any other float value.

lib/float.fz
```diff
@@ -55,6 +55,21 @@ float : numeric is
   redef infix ** (other float.this.type) float.this.type is abstract
 
 
+  # comparision
+  #
+  # This provides comparision operators using IEEE semantics.
+  #
```
fridis, line 61: maybe would be good to document that `NaN` compared to anything yields false. 

I have just realized that this implementation of `>` and `<` is wrong since, e.g., `NaN > 0` and `3.14 < NaN` would both result in `true`.

fridis (approved):
Just found a bug in `<` and `>`. 

Also some more documentation would be good (see detailed comments)

fridis (changes requested):
sorry, accidental approve in my previous review.

fridis (approved):

PR #1651, opened 2023-06-26
lib: concur.threads, add `join`

michaellilltokiwa:
I implemented this mainly because of the ctrie tests.


fridis (approved):
We will need an effect that groups thread operation like `spawn` and `join`, but this is independent of this PR:

PR #1662, opened 2023-06-28
parser/ast: change visibilities according to design

michaellilltokiwa:
for an in depth description see: https://flang.dev/design/visibility


michaellilltokiwa:
> This does not actually make Fuzion enforce visibility, right?

Nothing is checked right now regarding visibility. This just replaces the keyword used.

michaellilltokiwa:
superseded by #1714

Issue #1680, opened 2023-07-03
windows, SIGSEGV in malloc when counting larger lisz


michaellilltokiwa:
The example:
```
ex is

  testfile := "testfile" # this file is ~65kb

  f =>
    io.file.open

  page_size := 1000

  say (io.file.use String testfile io.file.mode.read ()->
    f
      .read
      .bind String (x -> String.type.from_bytes (x.drop page_size))
  )

```
gdb shows this:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffae409e639 in ntdll!RtlProtectHeap () from /c/Windows/SYSTEM32/ntdll.dll
(gdb) bt
#0  0x00007ffae409e639 in ntdll!RtlProtectHeap () from /c/Windows/SYSTEM32/ntdll.dll
#1  0x00007ffae409e37b in ntdll!RtlProtectHeap () from /c/Windows/SYSTEM32/ntdll.dll
#2  0x00007ffae40bdbfa in ntdll!RtlAllocateHeap () from /c/Windows/SYSTEM32/ntdll.dll
#3  0x00007ffae40bb44d in ntdll!RtlAllocateHeap () from /c/Windows/SYSTEM32/ntdll.dll
#4  0x00007ffae4188a58 in ntdll!RtlRegisterSecureMemoryCacheCallback () from /c/Windows/SYSTEM32/ntdll.dll
#5  0x00007ffae40bd255 in ntdll!RtlAllocateHeap () from /c/Windows/SYSTEM32/ntdll.dll
#6  0x00007ffae40bb44d in ntdll!RtlAllocateHeap () from /c/Windows/SYSTEM32/ntdll.dll
#7  0x00007ffae3c79d40 in msvcrt!malloc () from /c/Windows/System32/msvcrt.dll
#8  0x00007ff7b2ef2ea5 in fzC__L3294array_w_u__cons__tail ()
#9  0x00007ff7b2ef8bb3 in fzC_list_w_u8__1count ()
#10 0x00007ff7b2ef8be6 in fzC_list_w_u8__1count ()
#11 0x00007ff7b2ef8be6 in fzC_list_w_u8__1count ()
#12 0x00007ff7b2ef8be6 in fzC_list_w_u8__1count ()
#13 0x00007ff7b2ef8be6 in fzC_list_w_u8__1count ()
#14 0x00007ff7b2ef8be6 in fzC_list_w_u8__1count ()
[...]

PR #1682, opened 2023-07-04
dfa: refine escape analysis for direct tail calls

michaellilltokiwa:
- if the instances of original value and call match, call escapes only if it is not calling itself

This fixes issue #1680 but there is probably a better way to this? Also I'm not at all sure if this is correct in all cases...


michaellilltokiwa:
Tests are showing this is not correct.

PR #1688, opened 2023-07-05
Add intrinsics to read the content of a directory

maxteufel:
~~It still needs to be checked whether this works on Windows.~~ It does.

### DIFF
--- a/lib/fuzion/sys.fz
+++ b/lib/fuzion/sys.fz
@@ -35,9 +35,3 @@ module sys is
   #
   module c_string (s String) =>
     (s + codepoint 0).utf8.as_array.internal_array.data
-
-
-  # Check if a given integer would be cast to the NULL pointer
-  # in the C backend, or if it is -1 in the Java backend.
-  #
-  module is_null_pointer (ptr i64) i64 => intrinsic
--- a/lib/fuzion/sys.fz
+++ b/lib/fuzion/sys.fz
@@ -35,3 +35,9 @@ module sys is
   #
   module c_string (s String) =>
     (s + codepoint 0).utf8.as_array.internal_array.data
+
+
+  # Check if a given integer would be cast to the NULL pointer
+  # in the C backend, or if it is -1 in the Java backend.
+  #
+  module is_null_pointer (ptr i64) i64 => intrinsic
--- a/src/dev/flang/be/c/C.java
+++ b/src/dev/flang/be/c/C.java
@@ -908,8 +908,6 @@ public class C extends ANY
    */
   private void createCode(CFile cf, COptions _options) throws IOException
   {
-    cf.print("#define _GNU_SOURCE\n"); // needs to be first to have an effect
-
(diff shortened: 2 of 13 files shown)


michaellilltokiwa:
@maxteufel I think it would be good to remove intrinsic `is_null_pointer` and either let opendir return e.g. 0 on error or pass an array for more a complex result like it is done e.g. in `bind0`.

fridis:
> The problem with returning 0 on error is that there is no guarantee by C that this is not a valid pointer, pointers can also be negative so I can't use -1 either.

We can safely assume that 0 and -1 are invalid C pointers. The result of malloc is never odd, always aligned to at least 4 bytes, most likely even 8 bytes.

src/dev/flang/be/c/C.java
```diff
@@ -930,7 +932,8 @@ private void createCode(CFile cf, COptions _options) throws IOException
        "#include <errno.h>\n"+
        // defines _O_BINARY
        "#include <fcntl.h>\n"+
-       "#include <stdatomic.h>\n");
+       "#include <stdatomic.h>\n"+
+       "#include <dirent.h>\n");
```
michaellilltokiwa, line 936: Since `dirent` is POSIX the code should be put in `posix.c`. Since recently there are only C-standard headers included here.

lib/fuzion/sys.fz
```diff
@@ -35,3 +35,9 @@ module sys is
   #
   module c_string (s String) =>
     (s + codepoint 0).utf8.as_array.internal_array.data
+
+
+  # Check if a given integer would be cast to the NULL pointer
+  # in the C backend, or if it is -1 in the Java backend.
+  #
+  module is_null_pointer (ptr i64) i64 => intrinsic
```
michaellilltokiwa, line 43: As discussed I would like to ideally see this removed or for it to have another name since this name only makes sense for the c backend.

michaellilltokiwa (changes requested):
Would be cool to also add a simple effect wrapping this and a test.

src/dev/flang/be/c/C.java
michaellilltokiwa: @maxteufel Yes, of course you have to pull out the relevant code and put it in the `posix.c/win.c`

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -732,6 +732,18 @@ else if (count > 0)
     return stacktrace.toString();
   }
 
+  public static byte[] fuzion_sys_fileio_read_dir(long fd)
+  {
+    var i = (Iterator<Path>)_openStreams_.get(fd);
```
michaellilltokiwa, line 737: Call to `Runtime.unsafeIntrinsic();` is missing I think?

michaellilltokiwa (approved):
Looks nice but how did you test it when all of the features are invisible outside of base-lib?

Issue #1508, opened 2023-05-24
ClassCastException in interpreter


michaellilltokiwa:
```
ex =>

  _ := ["hello"]
    .map_sequence (x -> x.utf8.as_array.internal_array.data)
    .as_array
```
yields
```
error 1: *** java.lang.ClassCastException: class dev.flang.be.interpreter.ArrayData cannot be cast to class dev.flang.be.interpreter.ValueWithClazz (dev.flang.be.interpreter.ArrayData and dev.flang.be.interpreter.ValueWithClazz are in unnamed module of loader 'app')
Call stack:
(option Any).postfix ??: $FUZION/lib/option.fz:43:17:
  postfix ?? => exists
----------------^
(option Any).get: $FUZION/lib/option.fz:110:26:
      safety: option.this??
-------------------------^
((list Any).ref as_stream).next: $FUZION/lib/list.fz:410:23:
      res := cur.head.get
----------------------^
(((list Any).as_array).ref @1073751817).call: $FUZION/lib/Sequence.fz:119:24:
    array T count i->s.next
-----------------------^
(array Any).@1073858074: $FUZION/lib/array.fz:230:20:
    internal[x] := init x
-------------------^
array Any: $FUZION/lib/array.fz:229:3:
  for x in indices do
--^
(list Any).as_array: $FUZION/lib/Sequence.fz:119:5:
    array T count i->s.next
----^
ex: /home/sam/playground/test.fz:5:6:
    .as_array
-----^



error 2: java.lang.ClassCastException: class dev.flang.be.interpreter.ArrayData cannot be cast to class dev.flang.be.interpreter.ValueWithClazz (dev.flang.be.interpreter.ArrayData and dev.flang.be.interpreter.ValueWithClazz are in unnamed module of loader 'app')
	at dev.flang.be.interpreter.Interpreter.execute(Interpreter.java:432)
	at dev.flang.be.interpreter.Interpreter.execute(Interpreter.java:382)
	at dev.flang.be.interpreter.Interpreter.executeArgs(Interpreter.java:607)
```

michaellilltokiwa:
We are mapping pointers here, crash is okay and not possible to trigger with correct visibility.

PR #1708, opened 2023-07-12
lib/process: add intriniscs/effect for processes

michaellilltokiwa:
- starting, reading stdout/err, writing stdin, waiting, pipeing


michaellilltokiwa:
blocked by #1508, also buffered_reader should be merge first.

PR #1719, opened 2023-07-13
lib: rename `wrapping_integer` as `num.wrap_around`

maxteufel:
This was my suggestion with respect to moving the `numeric` hierarchy to `num` (this could be extended to `integer` -> `num.integer`, `numeric` -> `num.base`, `float` -> `num.float`, and finally, `num` -> `numeric`). But now that I see it, I'm not sure whether I like it. What do you think, @fridis @michaellilltokiwa?


fridis (approved):
good name.

PR #1720, opened 2023-07-13
parser: prohibit usage of the `set` keyword by default

maxteufel:
Limit the usage of the keyword to the standard library and to the existing tests that use it. Due to its nature involving mutation, it should not be used by any other code.

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -84,6 +84,13 @@ public class Parser extends Lexer
   /*----------------------------  variables  ----------------------------*/
 
 
+  /**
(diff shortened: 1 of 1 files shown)


src/dev/flang/parser/Parser.java
```diff
@@ -84,6 +84,9 @@ static enum EnvOrType
   /*----------------------------  variables  ----------------------------*/
 
 
+  public static boolean ENABLE_SET_KEYWORD = false;
```
michaellilltokiwa, line 87: Needs a comment

michaellilltokiwa (approved):

PR #1722, opened 2023-07-13
fe, tests: Type inference for formal arguments

fridis:
NOTE: This is experimental, would like to gain experience and know if this is useful or does harm.

With this patch, formal arguments types in a feature declaration can be omitted provided that the currently compiled module has a call to the formal argument's outer feature that provides an actual argument that permits type inference.

As an example, here is the first example from https://flang.dev/tutorial/feature_decl_fields
```
    Point(x, y i32) is

    show (Point 3 4 )
    show (Point 5 12)

    show(p Point) => say "{p.x}, {p.y}"
```
Type inference permits omitting all formal argument types in this example, so we can omit `i32` and `Point` types for `x`, `y` and `p` and write
```
    Point(x, y) is

    show (Point 3 4 )
    show (Point 5 12)

    show(p) => say "{p.x}, {p.y}"
```
This looks like magic, but it is in fact just a removal of redundant information, the language is still fully statically typed.  One caveat is that the documentation function of types in the source code gets lost, in particular since the call to a feature may be far away from its declaration, even in a separate file, but always in the same module. Documentation tools like `fz -docs` operate after the front end is done, so they should have full type information.

This, however, also creates new possibilities for errors: First, there might be several calls to the same feature with incompatible actual argument types. Second, there might be no call at all so the argument type is unknown. Both cases result in new errors.

A new negative test `typeinference_for_formal_args_negative` checks these error cases.

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -680,6 +680,7 @@ public class Call extends AbstractCall
    */
   void loadCalledFeature(Resolution res, AbstractFeature thiz)
   {
+    AbstractFeature targetFeature = null;
     if (PRECONDITIONS) require
       (thiz.state() == Feature.State.RESOLVING_INHERITANCE
        ? thiz.outer().state().atLeast(Feature.State.RESOLVED_DECLARATIONS)
@@ -693,7 +694,7 @@ public class Call extends AbstractCall
         var actualsResolved = false;
         if (_name != Errors.ERROR_STRING)    // If call parsing failed, don't even try
           {
-            var targetFeature = targetFeature(res, thiz);
+            targetFeature = targetFeature(res, thiz);
             if (CHECKS) check
               (Errors.count() > 0 || targetFeature != null && targetFeature != Types.f_ERROR);
             if (targetFeature != null && targetFeature != Types.f_ERROR)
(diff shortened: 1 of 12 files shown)


src/dev/flang/ast/Call.java
```diff
@@ -764,7 +766,7 @@ else if (_calledFeature == null) // nothing found, so flag error
       }
 
     if (POSTCONDITIONS) ensure
-      (Errors.count() > 0 || calledFeature() != Types.f_ERROR,
+                          (Errors.count() > 0 || calledFeature() != Types.f_ERROR || targetFeature == Types.resolved.f_void,
```
michaellilltokiwa, line 769: accidental indentation?

src/dev/flang/ast/Feature.java
```diff
@@ -603,7 +598,7 @@ public Feature(SourcePosition pos,
     this._visibility = v;
     this._modifiers  = m;
     this._returnType = r;
-    this._posOfReturnType = r == NoType.INSTANCE || r.isConstructorType() ? pos : r.functionReturnType().pos2BeRemoved();
+    this._posOfReturnType =  r == NoType.INSTANCE || r.isConstructorType() ? pos : r.functionReturnType().pos2BeRemoved();
```
michaellilltokiwa, line 601: accidental space?

src/dev/flang/ast/Impl.java
```diff
@@ -86,14 +88,24 @@ public Expr initialValue()
...
+   */
+  final List<AbstractFeature> _outerOfInitialValues;
 
 
   public enum Kind
   {
     FieldInit,    // a field with initialization syntactic sugar
     FieldDef,     // a field with implicit type
-    FieldActual,  // an argument field with type defined by actual argument
+    FieldActual, // an argument field with type defined by actual argument
```
michaellilltokiwa, line 108: indentation of comment

src/dev/flang/ast/Impl.java
```diff
@@ -109,7 +121,7 @@ public String toString(){
         {
           case FieldInit         : yield "field initialization";
           case FieldDef          : yield "field definition";
-          case FieldActual       : yield "actual argument";
+          case FieldActual      : yield "actual argument";
```
michaellilltokiwa, line 124: indentation, space missing

michaellilltokiwa (approved):

PR #1745, opened 2023-07-18
add first batch of visibilty rules and simple test

michaellilltokiwa:
All the weird/tricky cases regarding visibility seem to be related to inheritance.
Examples:

1) lets say we inherit from feature in another file that inherits from feature in current file.
should the inherited private features be visible?

2) We have this feature and return its supertype `Any` somewhere. Currently calling `as_string` results in `mod` even though the redefinition feature might not even be visible.
```
private mod : Any is
  private redef as_string is
   "mod"
```

3) what to do with cases like this?:
```
module feat : Sequence i_am_private is
```

4) I'm all but certain to have thought of all possible cases...

---

Reusing names for fields (see commented out `say b.get_field`) is not yet working. crashes:
```
error 2: java.lang.Error: require-condition3 failed: dev.flang.be.interpreter.Layout:offset:246
	at dev.flang.util.ANY.require(ANY.java:113)
	at dev.flang.be.interpreter.Layout.offset(Layout.java:246)
```
```
error 1: java.lang.Error: Value.setField for 'b.field' called on class UNIT (class dev.flang.fuir.analysis.dfa.Value$2), expected class dev.flang.fuir.analysis.dfa.Instance
```


src/dev/flang/ast/AbstractType.java
```diff
@@ -1562,6 +1564,64 @@ static boolean checkActualTypePars(SourcePosition pos, AbstractFeature called, L
...
+              {
+                at.usedTypes(result);
+              }
+          }
+      }
+  }
+
+
+  /**
+   * This type and any of its generic types that are less visible than `v`.
```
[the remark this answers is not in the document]
michaellilltokiwa: I meant to write any of its generics.

src/dev/flang/parser/Parser.java
```diff
@@ -416,9 +416,6 @@ else if (s instanceof Feature f)
           {
             if (first)
               {
-                if (CHECKS) check
```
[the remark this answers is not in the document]
michaellilltokiwa: because _visibilty is private and visibility() does not return unspecified anymore but turns unspecified into private

src/dev/flang/fe/Module.java
```diff
@@ -313,7 +368,7 @@ public SortedMap<FeatureName, AbstractFeature> declaredOrInheritedFeatures(Abstr
                     // * same as previous, but there is some syntax for 'D' to
                     //   chose 'a.[B].f' or 'a.[C].f'.
                     //
-                    if (existing != null)
+                    if (existing != null && f != existing)
```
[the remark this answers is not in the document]
michaellilltokiwa: yes, but it does not work for features yet, since renaming is not yet supported.

src/dev/flang/ast/AbstractType.java
```diff
@@ -1562,6 +1564,64 @@ static boolean checkActualTypePars(SourcePosition pos, AbstractFeature called, L
...
+          {
+            at.usedTypes(result);
+          }
+        if (isChoice())
+          {
+            for (AbstractType at : this.choiceGenerics())
+              {
+                at.usedTypes(result);
+              }
+          }
```
fridis, line 1601: I think this should not be needed, but it depends what you want. The current implementation for a type `t` in
```
  t : u a b is
```
woulbe be just `t` while with
```
  t : choice a b is
```
would be `t`, `a`, `b`.

I think you should only check the visibility of the direct type parameter, for inherited type parameters, this should be checked at the declaration of the type `t`: if `t` inherits from some `u` that is at least as visible as `t`, then all the type parameter of `u` should be as visible as `t`. So your do not need to check the inheritance at a use of `t`, also no need to check the choice generics.

src/dev/flang/ast/AbstractType.java
```diff
@@ -1562,6 +1564,64 @@ static boolean checkActualTypePars(SourcePosition pos, AbstractFeature called, L
...
+   * example: Cons i32 (list i32) => { Cons, i32, list }
+   *
+   * @return
+   */
+  private Set<AbstractType> usedTypes()
+  {
+    var result = new TreeSet<AbstractType>();
+    usedTypes(result);
+    return result;
+  }
```
fridis, line 1580: generally, I prefer traversing on demand. There is a method `applyToGenericsAndOuter` that visits all used types (which is maybe too much), but you could add a variant `applyToXYZ`.
Here, I think all you need to check is the direct type parameters (see comments below), which are directly available with `generics()´.

src/dev/flang/ast/AbstractType.java
```diff
@@ -1562,6 +1564,64 @@ static boolean checkActualTypePars(SourcePosition pos, AbstractFeature called, L
...
+   * @param v
+   * @return
+   */
+  public Set<AbstractType> lessVisibleThan(Visi v)
+  {
+    if (PRECONDITIONS) require
+      (!v.definesTypeVisibility());
+
+    return isGenericArgument()
+      ? genericArgument().constraint().lessVisibleThan(v)
```
fridis, line 1618: The visibiliy of type parameter should probably not depend on the constraint, but on its underlying feature, i.e., `genericArgument().typeParameter()`?

src/dev/flang/ast/AbstractType.java
```diff
@@ -1562,6 +1564,64 @@ static boolean checkActualTypePars(SourcePosition pos, AbstractFeature called, L
...
+  }
+
+
+  /**
+   * This type and any of its generics that are less visible than `v`.
+   *
+   * @param v
+   * @return
+   */
+  public Set<AbstractType> lessVisibleThan(Visi v)
```
fridis, line 1612: I find the term _less visible_ not very intuitive, how about _has more restrictive visibility than_? Also for the error messages.

src/dev/flang/ast/Visi.java
```diff
@@ -99,14 +99,65 @@ public static Visi from(int ordinal)
...
+  }
+
+
+  /**
+   * @return The visibilty for types encoded in this.
+   * PRIV => PRIV, PRIVMOD => MOD, PRIVPUB => PUB, etc.
+   */
+  public Visi typeVisibility()
+  {
+    switch (this)
```
fridis, line 134: use the new Java syntax: 
```
  return switch (this) {
    case UNSPECIFIED, PRIV -> Visi.PRIV; 
    case ...
```

src/dev/flang/fe/Module.java
```diff
@@ -269,6 +268,62 @@ else if (existing == f && !existing.generics().equals(f.generics()) ||
...
+   * @param af
+   * @return
+   */
+  protected boolean typeVisible(SourceFile usedIn, AbstractFeature af)
+  {
+    var m = (af instanceof LibraryFeature lf) ? lf._libModule : this;
+    var definedIn = af.pos()._sourceFile;
+    var v = af.visibility();
+
+    // NYI check if generics/constraints visible?
```
fridis, line 285: This should not be checked here, but at the declaration of `af`.

src/dev/flang/fe/Module.java
```diff
@@ -269,6 +268,62 @@ else if (existing == f && !existing.generics().equals(f.generics()) ||
...
+   * Is type defined by feature `af` visible in file `usedIn`?
+   * If `af` does not define a type, result is false.
+   *
+   * @param usedIn
+   * @param af
+   * @return
+   */
+  protected boolean typeVisible(SourceFile usedIn, AbstractFeature af)
+  {
+    var m = (af instanceof LibraryFeature lf) ? lf._libModule : this;
```
fridis, line 281: Avoid `instanceof`, instead please add a method `libModule()` or similar to `AbstractFeature' that returns `null` or the source module and that is redefined in `LibraryFeature`.

src/dev/flang/fe/SourceModule.java
```diff
@@ -969,13 +973,16 @@ public AbstractFeature lookupFeature(AbstractFeature outer, FeatureName name, Ab
    * outer's outer (i.e., use is unqualified), false to search in outer only
    * (i.e., use is qualified with outer).
    *
+   * @param sf the source file where type is used
```
fridis, line 976: `where feature is used`, not `type`.

When can this be `null`, what does that mean?

src/dev/flang/fe/SourceModule.java
```diff
@@ -1283,6 +1294,103 @@ else if ((t1.isChoice()
...
+  }
+
+
+  /**
+   * Check that type of outer feature of `f` isn't less visible than `f` itself.
+   * @param f
+   */
+  private void checkOuterVisibility(Feature f)
+  {
+    if (f.outer() != null && f.outer().visibility().typeVisibility().ordinal() < f.visibility().ordinal())
```
fridis, line 1311: I think we will have to drop this restriction, as discussed, a private feature should be able to define a public `as_String`.

fridis (changes requested):
Very nice. Particularly like that you did a big effort to check many special cases of visibility. 

Just some minor comments suggestions from me.

src/dev/flang/fe/Module.java
michaellilltokiwa: but then ast needs to dependend on fe?

src/dev/flang/ast/AbstractType.java
michaellilltokiwa: Maybe I misunderstand but this should be okay (`Sequence` being public):
```
private a : Sequence some_private_type is
```
while this is not
```
public a : Sequence some_private_type is
```
So I think it is in general not sufficient to just check direct type parameters?

src/dev/flang/fe/SourceModule.java
michaellilltokiwa: I dropped this check.

src/dev/flang/fe/SourceModule.java
michaellilltokiwa: I removed this parameter, filtering the result afterwards instead.

src/dev/flang/fe/Module.java
michaellilltokiwa: I removed the comment.

src/dev/flang/ast/Visi.java
michaellilltokiwa: updated

src/dev/flang/ast/AbstractType.java
michaellilltokiwa: used `moreRestrictiveVisibility`

src/dev/flang/ast/AbstractType.java
michaellilltokiwa: now, done as suggested

src/dev/flang/ast/AbstractType.java
michaellilltokiwa: I now added an abstract method `abstract void usedFeatures(Set<AbstractFeature> s)` to `AbstractType` which collects all the features the type uses/needs into a set. These can then be checked for sufficient visibility.

michaellilltokiwa:
#1776 should be merge first

src/dev/flang/fe/SourceModule.java
```diff
@@ -1110,7 +1119,10 @@ public FeatureAndOuter lookupType(SourcePosition pos,
...
         var type_fs = new List<AbstractFeature>();
         var nontype_fs = new List<AbstractFeature>();
-        var fs = lookup(outer, name, null, traverseOuter);
+        var fs = lookup(outer, name, null, traverseOuter)
+                   .stream()
+                   .filter(fo -> typeVisible(pos._sourceFile, fo._feature))
+                   .collect(Collectors.toList());
         for (var fo : fs)
           {
             var f = fo._feature;
```
fridis, line 1128: Same here, maybe add an `if (typeVisible(...))` in the loop instead of using the stream.

src/dev/flang/ast/AbstractType.java
```diff
@@ -1680,6 +1689,38 @@ AbstractType clone(AbstractFeature originalOuterFeature)
...
+   */
+  protected abstract void usedFeatures(Set<AbstractFeature> s);
+
+
+  /**
+   * @param v
+   *
+   * @return this type and any of its generics that have more restrictive visibility than `v`.
+   */
+  public Set<AbstractFeature> moreRestrictiveVisibility(Visi v)
```
fridis, line 1706: This is ok for now, but think it might be better to use a visitor similar to `applyToGenericsAndOuter` eventually.

src/dev/flang/ast/ResolvedNormalType.java
```diff
@@ -646,6 +651,31 @@ AbstractFeature originalOuterFeature(AbstractFeature currentOuter)
...
 
+
+  /**
+   * traverse a resolved type collecting all features this type uses.
+   *
+   * @param s the features that have already been found
+   */
+  protected void usedFeatures(Set<AbstractFeature> s)
+  {
+    var f = featureOfType();
```
fridis, line 662: This currently does not touch the outer features. This means that for a type like `(x T).y U` the visibility of `x` and `T` will be ignored, which is probably wrong. 

But this can be fixed later.

fridis (changes requested):
Looks good. 

Just some minor restructuring requests and a NYI that can be removed.

fridis (approved):

PR #1752, opened 2023-07-20
be/c: add method for calling reasonably simple C functions

maxteufel:
Reasonably simple means that the C functions only take integer, floating-point and string arguments, and also only return a result of one of these types.

The DFA is adjusted to automatically handle these calls, as well.

### DIFF
--- a/src/dev/flang/be/c/Intrinsics.java
+++ b/src/dev/flang/be/c/Intrinsics.java
@@ -1199,9 +1199,10 @@ public class Intrinsics extends ANY
 
     var in = c._fuir.clazzIntrinsicName(cl);
     var cg = _intrinsics_.get(in);
+    var result = CStmnt.EMPTY;
     if (cg != null)
       {
-        return cg.get(c, cl, outer, in);
+        result = cg.get(c, cl, outer, in);
       }
     else
       {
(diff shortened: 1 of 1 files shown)


src/dev/flang/be/c/Intrinsics.java
```diff
@@ -1213,13 +1213,55 @@ CStmnt code(C c, int cl)
...
+                         c._fuir.clazzIs(ac, FUIR.SpecialClazzes.c_u32) ||
+                         c._fuir.clazzIs(ac, FUIR.SpecialClazzes.c_u64) ||
+                         c._fuir.clazzIs(ac, FUIR.SpecialClazzes.c_f32) ||
+                         c._fuir.clazzIs(ac, FUIR.SpecialClazzes.c_f64))
+                  {
+                    args.add(new CIdent("arg" + i));
+                  }
+                else
+                  {
+                    {
```
fridis, line 1241: unmotivated `{`

src/dev/flang/be/c/Intrinsics.java
```diff
@@ -1213,13 +1213,55 @@ CStmnt code(C c, int cl)
...
+              {
+                var str = new CIdent("str");
+                var res = new CIdent("res");
+                return CStmnt.seq(CExpr.decl("char*", str, CExpr.call(c._fuir.clazzBaseName(cl), args)),
+                                  c.constString(str, CExpr.call("strlen", new List<>(str)), res),
+                                  res.castTo(c._types.clazz(rc)).ret());
+              }
+
+            return CStmnt.seq(CExpr.call(c._fuir.clazzBaseName(cl), args).ret());
           }
```
fridis, line 1265: These nested returns at different nesting levels make the code very hard to read. Please restructure this such that there is only one singe return at the end.

fridis (changes requested):
Nice.  Just some code style issues to fix. 

For a separate PR: Can you check what would be a good syntax for these, just using `intrinsic` is IMHO not nice, we would need something like `extern "C"`, `native`, etc.

PR #1760, opened 2023-07-23
fe: cleanup of type resolution, fix #1084

### LINKED ISSUE #1084: discussion: remove pos from (Abstract)Type
I think the position of a Type is too ambiguous to give satisfying information. IMHO a type can have anything from zero to a lot of positions. 

In many cases what we actually want is the position of the feature defining the type. Then there is types which are generic parameters, so I think we should add `pos` to `Generic`.

As can be seen in #1082 there is 'only' two failing tests if the type always reports the position of the feature if available and otherwise returns that no position is available.

fridis:
Replaces class `Type` by a hierachy of classes: First, `UnresolvedType` instances like `ParsedType` and `BuiltInType` are created. Unresolved types have a source code position.

During `findGenerics` and type resolution, these unresolved types are are replaced instances of `ResolvedType`, in particular `ResolvedParametricType` for type parameters and `ResolvedNormalType` for types based on a routine or a choice feature.  Resolved types are interned, they have a new feature `declarationPos` for the location the type's underlying feature is declared. they no longer have information on the source code location they came from (with some exceptions like yResolvedNomarlType.unresolvedGenerics` that provides the original type parameters when available for error messages).

`AbstractType.pos2BeRemoved` was removed.

`AstErrors.outerTypeMayNotBeRefType` was removed since `ref` types had been removed before.

The complex visiting for Type resolution using `FeatureVisitor.actionBefore` and `Type.visit2` has been cleaned up.

`AbstractType.visit` no longer recursively visits all the type parameters of that type, but performs the provided action only on that type itself. So the action will have to traferse the type if it needs to.x


src/dev/flang/ast/AbstractType.java
```diff
@@ -1554,14 +1651,36 @@ static boolean checkActualTypePars(SourcePosition pos, AbstractFeature called, L
...
+   *
+   * This is used for type features that use types from the original feature,
+   * but needs to replace generics by the type feature's generics.
+   *
+   * @param originalOuterFeature the original feature, which is not a type
+   * feature.
+   */
+  AbstractType clone(AbstractFeature originalOuterFeature)
+  {
+    //throw new Error("clone(originalOuterFeature) not supported for "+getClass());
```
michaellilltokiwa, line 1680: comment was used for debugging probably?

src/dev/flang/ast/AbstractType.java
```diff
@@ -1409,14 +1467,44 @@ public boolean containsThisType()
...
+  /**
+   * For a normal type, this is the list of actual type parameters given to the type.
+   */
   public abstract List<AbstractType> generics();
+
+
+  /**
+   * For a normal type, this is the list of the unresolved version of actual
+   * type parameters given to the type, as far as they are available. They are
+   * not avaible, e.g., when the type was inferred or was loaded from a module
```
michaellilltokiwa, line 1502: available

src/dev/flang/fe/NormalType.java
```diff
@@ -143,10 +149,17 @@ public AbstractType applyTypePars(List<AbstractType> g2, AbstractType o2)
     if (PRECONDITIONS) require
       (!isGenericArgument());
 
-    return new NormalType(_libModule, _at, _pos, _feature, _valRefOrThis, g2, o2);
+    return new NormalType(_libModule, _at, _feature, _valRefOrThis, g2, o2);
   }
 
 
+  /**
+   * For a resulved normal type, return the underyling feature.
```
michaellilltokiwa, line 157: resolved

michaellilltokiwa: underlying

src/dev/flang/fe/GenericType.java
```diff
@@ -108,6 +112,13 @@ public AbstractType visit(FeatureVisitor v, AbstractFeature outerfeat)
   }
 
 
+  /**
+   * For a resulved normal type, return the underyling feature.
```
michaellilltokiwa, line 116: resolved

michaellilltokiwa (approved):

PR #1766, opened 2023-07-25
add badges of testruns to readme

michaellilltokiwa:
also changed to run the tests weekly on apple

image

on windows, mmap test fails because required stack size is larger than windows default and `list.concat` can not be tail call optimized. this can be reproduced on linux easily by decreasing stack size e.g. `ulimit -s 512`
on apple, floating point test fails probably due to a rounding error.


michaellilltokiwa:
@maxteufel Yes, I agree and added the linux badge.

PR #1767, opened 2023-07-25
fix NPE when building bad base.fum

michaellilltokiwa:
- code cleanups
- fixes in comments

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -1412,7 +1412,7 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
 
 
   /**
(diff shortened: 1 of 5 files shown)


michaellilltokiwa:
One example would be:
```diff
diff --git a/lib/Any.fz b/lib/Any.fz
index 2e0819f7..bbcaec7d 100644
--- a/lib/Any.fz
+++ b/lib/Any.fz
@@ -25,7 +25,7 @@
 
 # Any -- parent feature of all features that do not have an explicit parent
 #
-public Any ref is
+public Ay ref is
 
   public hash_code i32 is intrinsic
```
In general if any feature from Types.resolved is broken different than Types.resolved expects.

PR #1772, opened 2023-07-26
be/c: add method for calling reasonably simple C functions

maxteufel:
Reasonably simple means that the C functions only take integer, floating-point and string arguments, and also only return a result of one of these types.

The DFA is adjusted to automatically handle these calls, as well.


fridis (approved):
Looks good.

PR #1773, opened 2023-07-26
be/c: use a single return statement in `Intrinsics.code`

maxteufel:
This matches more closely the preferred code style for the Fuzion codebase which discourages the use of returns in branches and encourages a single return at the end of each function.


fridis (approved):
That's good, but in this case I would have had no problem with the original code since it is just nested if's that all had a `return` as the last statement, so pretty much the same way Fuzion returns a result :-).

PR #1774, opened 2023-07-26
fe: Support free types, fix #1656

### LINKED ISSUE #1656: Syntax sugar for implicit type parameters?
Declaring a feature with type parameters looks something like

```
max(T type : has_total_order, a, b T) T is
  if a <= b then b else a
```

which is even longer (looking at the signature declaration only) than the corresponding Java code

```java
<T extends Comparable> T max(T a, T b)
{
  if (a.compareTo(b) <= 0) return b; else return a;
}
```
How about declaring type parameters implicitly as follows
```
max(a, b T : has_total_order) T is
  if a <= b then b else a
```
or allowing anonymous type parameters when they are not needed
```
max(a, b : has_total_order) =>
  if a <= b then b else a
```
Would this be helpful? Or would this be confusing? Will users understand the difference between this and
```
max(a, b i64) =>
  if a <= b then b else a
```
?

fridis:
Within formal argument lists, type parameters can now be added inline. I call these _free types_, the basic syntax is `T : Constraint` as in
```
  a(v T : Any) => ...
```
which is equivalent to
```
  a(T type : Any, v T) => ...
```
The constraint can be dropped, i.e., any unqualified type name without actual type parameters that is not found will be treated as a free type as in
```
  a(v T) => ...
```
Free types can be shared by several arguments
```
  a(v, w T) => ...
  a(V T, w T) => ...
```
which means that the types must all be the same. Also, free types can be anonymous using `_`:
```
  a(v, w _) => ...
```
which means that the type cannot be referenced explicitly, but there is only one type parameter and `v` and `w` have to be the same type.

In contrast, several anonymous free types are distinct, i.e.,
```
  a(v _, w _) => ...
```
has two different anonymous type parameters, one for `v` and one for `w`.

Anonymous type parameters and type inference for arguments as in
```
  a(v, w) => ...
```
is different, the latter infers exactly one static type for v and one for w, while the former creates a type polymorphic feature.

Free types may be used as type parameters within other types, e.g, as a choice
```
  g(v A | (B : numeric)) => ...
```
or as Sequence
```
  x(s Sequence T) => ...
```
Next to supporting free types, this PR adds new errors trying to limit confusion in case of a typo in a type.  For safety, free types must not mask existing types.

This patch adds positve and negetive tests `free_types` and `free_types_negative` with a number of examples.


fridis:
> The thing I found most surprising when playing around was that when specifying explicit types is that you also have to specify the free types (after the non-free types), like so:

I thought about forbidding mixing explicit and free type parameters, but decided to allow it for now. Let's see if this is useful or too confusing.

Also, unrelated to this patch, it could be possible on a call to infer some of the type parameters while requiring some to be given explicitly. The problem is to find a good rule or syntax, e.g. if you want to infer `u` in `zip t u l a,b->x` and you write `zip t l a,b->x`, how could you distinguish this from the case you want to infer `t`? Maybe `zip t _ l a,b->x` and `zip _ u l a,b->x` would be a solution?

fridis:
> ```
> ex_free2 is
>   a(x, y T, z U) is
>   a(x, y, z V) is
> ```
> 
> 
>     
>   
> 
> throws a duplicate feature error, even though semantically it should not be different.

I have added a NYI for this problem which is only in the source module for now.

fridis:
> To elaborate on what I expected when specifying types explicitly with free types used: I expected that you could never specify a free type explicitly, and that those would always be inferred.
> 
> So that with `zip(V type, b Sequence U, f (T, U) -> V) => as_list.zip0 b.as_list f` you would explicitly specify the type like this: `zip (tuple I T) (start_idx..) (a,b -> (b, a))`. However, I see that this might cause ambiguities with feature overloading.

This was not my idea, I only see free types as a short way to specify type parameters.

michaellilltokiwa (approved):

PR #1785, opened 2023-07-31
unify indentation of `=>` in base-lib and tests

michaellilltokiwa:


michaellilltokiwa:
No, I did not do that yet.

fridis (approved):
I remember that there was some issue that requires `=>` to be indented in an unintuitive way. Has anything with respect to this been fixed?

PR #1809, opened 2023-08-02
ast: use `Expr` instead of `Stmnt`

michaellilltokiwa:
This PR does not remove `Stmnt` just yet. Could be done in a successive PR.


michaellilltokiwa:
superseded by #1812

PR #1812, opened 2023-08-03
ast: use `Expr` instead of `Stmnt`

michaellilltokiwa:


fridis (approved):
Looks good. In a next patch, it would be great to get rid of `producesResults`, since everything produces a result, but sometimes it is of type `unit` or even `void`.

PR #1814, opened 2023-08-03
improve backward type propagation for arrays

### LINKED ISSUE #169: language improvement: backward type propagation for []
the type of `[]` is `array<void>`, it would be nicer to use the type of the target if available, i.e., 
```
a array<i32> := []                        # should result in array<i32>
b Sequence<i32> := []                     # should result in array<i32>
c Sequence<u8> := []                      # should result in array<u8>
d Sequence<f32> := [0,1,2]                # should result in array<f32>
d Sequence<Sequence<i16>> := [[0,1],[2]]  # should result in array<array<i16>>
```

michaellilltokiwa:
- fixes #169

### DIFF
--- a/src/dev/flang/ast/InlineArray.java
+++ b/src/dev/flang/ast/InlineArray.java
@@ -181,6 +181,7 @@ public class InlineArray extends ExprWithPos
     if (PRECONDITIONS) require
       (t != null);
 
+    // NYI see issue: #1817
     if (Types.resolved.f_array.inheritsFrom(t.featureOfType()) &&
         t.generics().size() == 1)
       {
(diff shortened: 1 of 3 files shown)


src/dev/flang/ast/InlineArray.java
```diff
@@ -181,7 +181,7 @@ private AbstractType elementType(AbstractType t)
     if (PRECONDITIONS) require
       (t != null);
 
-    if (t.featureOfType() == Types.resolved.f_array &&
+    if (Types.resolved.f_array.inheritsFrom(t.featureOfType()) &&
         t.generics().size() == 1)
```
fridis, line 185: Since `Sequence` is the only parent of `array` with one argument, this works, but it is a bit brittle since changes to the parents of `array` may break this code (say, we make `array` inherit from `has_size i32`).

For now, this code is goo, but please add a `NYI` comment here and refer to #1817 such that this can be improved once a more powerful mechanism is in place.

fridis (changes requested):
Suprisingly small patch with a big effect!

This can be used, but add a comment to revisit this once #1817 is fixed.

fridis (approved):

PR #1818, opened 2023-08-04
constraintAssignableFrom, check if generics are assignable

### LINKED ISSUE #1054: No error even though constraint not satisfied
This small example should produce an error since `f32` does not satisfy constraint `float f64`:
```
ex =>
  c(T type : float f64) is
  c f32
```

michaellilltokiwa:
- fixes #1054


fridis (approved):
Nice. 

Would be good to have issues, best with examples, for the remaining NYI-comments.

PR #1833, opened 2023-08-09
fe: infer result types from the result type of redefined features

maxteufel:


src/dev/flang/ast/Impl.java
```diff
@@ -522,13 +525,42 @@ AbstractType inferredType(Resolution res, AbstractFeature f)
...
       case FieldDef    -> _initialValue.typeIfKnown();
-      case RoutineDef  -> _code.typeIfKnown();
+      case RoutineDef  -> typeFromRedefinedFeatures(f);
       case FieldActual -> typeFromInitialValues(res, f, false);
       default -> throw new Error("missing case "+_kind);
       };
   }
 
 
+  AbstractType typeFromRedefinedFeatures(AbstractFeature f)
```
fridis, line 535: This should be named `typeFromRedefinedOrCode` or similar, in most cases `f.redefines()` will be empty and the type will be taken from the code.

Please add a comment to this. You may need to add `res` as an argument.

src/dev/flang/ast/Impl.java
```diff
@@ -522,13 +525,42 @@ AbstractType inferredType(Resolution res, AbstractFeature f)
...
+  {
+    AbstractType result;
+
+    if (PRECONDITIONS) require
+      (_kind == Kind.RoutineDef);
+
+    var rf = f.redefines()
+              .stream()
+              .filter(x -> x.state().atLeast(State.RESOLVED_TYPES))
```
fridis, line 544: Instead of filtering what happens to have been resolved already (which depends on whether we are lucky or not), maybe explicitly resolve `x` via a call to `res.resolveTypes(x)`.

src/dev/flang/ast/Impl.java
```diff
@@ -522,13 +525,42 @@ AbstractType inferredType(Resolution res, AbstractFeature f)
...
+              .map(x -> x.resultType())
+              .filter(x -> !x.isGenericArgument())
+              .filter(x -> x.generics().size() == 0)
+              .filter(x -> !x.isThisType())
```
fridis, line 548: Instead of these restrictions, it should be possible to use `AbstractFeature.handDownNonOpen` to get the type after it was inherited. Admittedly,  `handDownNonOpen` could have a nicer name `handDownType` or `inheritedType` or similar. 
The remaining restriction before using `handDownNonOpen` is that `!x.isOpenGeneric()`. Would be great if you could try to create an example that uses an open generic result type and redefines that feature to check what makes sense here (in a separate issue / PR).

src/dev/flang/ast/Impl.java
```diff
@@ -522,13 +525,42 @@ AbstractType inferredType(Resolution res, AbstractFeature f)
...
+              .collect(Collectors.toList());
+
+    if (rf.size() == 1)
```
fridis, line 551: In case there are several redefined features, your should IMHO still use the type from the redefined feature, at least as long as all are equal (`compareTo()==0`). 
Best would be to have `AbstractType.intersection` (similar to `AbstractType.union`) that returns a type that can be assigned to both (for ref types, the more specific type, not the more general type returned by `union`, and `void ∩ T == void`, `ERROR ∩ T = T`, basically everything the other way around than in `union` without support for boxing (`asRef`)).

fridis (changes requested):
Nice you found a solution to this. 

I have added a number of comments and request to make this more generic, hope this helps your understanding of the front end and to improve this patch.

PR #1836, opened 2023-08-10
fe: explicit error message when using a choice type as constraint

maxteufel:


src/dev/flang/ast/AstErrors.java
```diff
@@ -1621,7 +1621,15 @@ static void incompatibleActualGeneric(SourcePosition pos, Generic f, AbstractTyp
         error(pos,
               "Incompatible type parameter",
               "formal type parameter " + s(f) + " with constraint " + s(f.constraint()) + "\n"+
-              "actual type parameter " + s(g) + "\n");
+              "actual type parameter " + s(g) + "\n" +
+              (f.constraint().isChoice() ? "A choice type cannot be used as a constraint.\n" +
+                                           "To solve this, you could change the constraint to a newly created dummy feature and make " +
+                                           f.constraint().choiceGenerics()
```
michaellilltokiwa, line 1627: I think this then is a choice type contradicting the previous claim that a one can not use a choice type as constraint?

michaellilltokiwa (approved):

PR #1839, opened 2023-08-15
mark `numeric.from_u32` as abstract

michaellilltokiwa:
implement type.zero/type.one via `from_u32`

### DIFF
--- a/lib/f32.fz
+++ b/lib/f32.fz
@@ -155,14 +155,11 @@ public f32(val f32) : float is
   # type features:
 
 
-  # identity element for 'infix +'
+  # the value corresponding to v in whatever
+  # numeric implementation we have,
+  # maximum in case of overflow
   #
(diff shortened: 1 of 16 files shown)


fridis (commented):
@maxteufel , what do you think of this change from a mathematical point of view. The basic question is if a `numeric`, which basically defines a mathematical field, should define zero/one or should provide a mapping from all _small_ positive integers. I think 0/1 may be more natural. 

Also, the original code that implements `from_u32` using `zero` and `one` is a challenge for the compilers that I expect them to solve some time.

michaellilltokiwa:
@fridis, @maxteufel marked this as draft, since I didn't know/think `if v = 0 zero else (from_u32 v-1) +^ one` could solve this,
so it should probably stay as it is.

michaellilltokiwa:
closing, since it will stay as is.

PR #1852, opened 2023-08-15
makefile: mark tests/include as phony

michaellilltokiwa:
For the `lib` rebuild on change is done ensured via `$(shell find $(FZ_SRC_LIB) -name "*.fz")`.
No similar thing is done for tests/include currently which is a a bit inconvenient.
Since copying tests and include takes next to no time marking as phony is good enough, I suppose.


michaellilltokiwa:
superseded by #1855

PR #1864, opened 2023-08-17
be/c: remove redundant cast in `accessField`

michaellilltokiwa:


fridis (approved):
Very nice, but how did you find this???

PR #1866, opened 2023-08-17
`checkTypes` use `isDirectlyAssignable`

michaellilltokiwa:
instead of `isAssignable`

### DIFF
--- a/src/dev/flang/ast/InlineArray.java
+++ b/src/dev/flang/ast/InlineArray.java
@@ -277,7 +277,7 @@ public class InlineArray extends ExprWithPos
 
     for (var e : _elements)
       {
-        if (!elementType.isAssignableFrom(e.type()))
+        if (!elementType.isDirectlyAssignableFrom(e.type()))
           {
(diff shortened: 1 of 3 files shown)


src/dev/flang/ast/InlineArray.java
```diff
@@ -277,7 +277,7 @@ public void checkTypes()
 
     for (var e : _elements)
       {
-        if (!elementType.isAssignableFrom(e.type()))
+        if (!elementType.isDirectlyAssignableFrom(e.type()))
```
fridis, line 280: I think this would have the effect that code like
```
  x array (option i32) := [nil, 42, nil, 4711]
```
would no longer work or does it?

michaellilltokiwa: It should still work. (I just tested it.) 
Assumption is that checkTypes happens after boxing.

fridis: Yes, did not check this well. During checkTypes, isAssignableFrom should no longer be used.

fridis (approved):

Issue #1902, opened 2023-08-24
be/jvm: choice kind `voidlike`/`unitlike` does not work when using its outer ref


michaellilltokiwa:
here `t` is passed as an outer reference to `as_num`. But `t` has type `void`. But arguments may not be of type `void` in java. Thus classloader complains.
```
  t: choice a b is
    as_num =>
      match t.this
        a a => "a"
        b b => "b"
```
As a workaround I disabled choice-implkinds `voidlike`, `unitlike` for now:
```patch
diff --git a/src/dev/flang/be/jvm/Choices.java b/src/dev/flang/be/jvm/Choices.java
index daa3ba4d..d9b9f587 100644
--- a/src/dev/flang/be/jvm/Choices.java
+++ b/src/dev/flang/be/jvm/Choices.java
@@ -157,8 +157,8 @@ public class Choices extends ANY implements ClassFileConstants
           }
       }
 
-    if      (nonVoid == 0                               ) { return ImplKind.voidlike; }
-    else if (nonVoid == units && units == 1             ) { return ImplKind.unitlike; }
+    if      (nonVoid == 0                               ) { return ImplKind.general; }
+    else if (nonVoid == units && units == 1             ) { return ImplKind.general; }
     else if (nonVoid == units && units == 2             ) { return ImplKind.boollike; }
     else if (nonVoid == units                           ) { return ImplKind.intlike;  }
     else if (nonVoid == 2     && units == 1 && refs == 1) { return ImplKind.nullable; }
```

michaellilltokiwa:
The source of the problem is `clazzChoice` which returns void if element is never instantiated.

```
  public int clazzChoice(int cl, int i)
  {
    if (PRECONDITIONS) require
      (i >= 0 && i < clazzNumChoices(cl));

    var cc = clazz(cl);
    var cg = cc.choiceGenerics().get(i);
    var res = cg.isRef()          ||
              cg.isInstantiated()    ? cg
                                     : Clazzes.c_void.get();
    return id(res);
  }
```

PR #1904, opened 2023-08-24
be/jvm: implement intrinsics, `sys.net.***`

michaellilltokiwa:


michaellilltokiwa:
#1902 needs to be fixed first

fridis (approved):

PR #1917, opened 2023-08-28
lib: implement treiber-stack

michaellilltokiwa:
https://en.wikipedia.org/wiki/Treiber_stack


michaellilltokiwa:
@maxteufel maybe you're right and it is better to put this into a module.
@fridis what is your opinion? do we need a concurrent stack in the standard library?

fridis (changes requested):

michaellilltokiwa:
@fridis I added a new module `lock_free` and moved the stack to this module.

fridis (approved):
Fine to have this in a separate module. We will need some thought on structuring the modules, but this might be easier if we have some contents to put into that structure:-).

PR #1922, opened 2023-08-29
be/jvm: implement the `concur.util.loadFence` and `concur.util.storeFence` intrinsics

maxteufel:


michaellilltokiwa (approved):

fridis (approved):
I had hoped we could map the fences to the [internal fences in OpenJDK](https://github.com/openjdk/jdk/blob/3c743cfea00692d0b938cb1cbde936084eecf369/src/java.base/share/classes/jdk/internal/misc/Unsafe.java#L3398), but this does not seem to be possible any more...

So ok for this patch, we might not be able to do any better.

PR #1925, opened 2023-08-29
remove `unbox` from IR and AST.

### LINKED ISSUE #1435: Remove `Unbox` from IR
`Unbox` is the identity operation in the C backend and it is unclear if it is needed in the interpreter backend.  Maybe it can be removed from FUIR, AST, etc.?

michaellilltokiwa:
fixes #1435


fridis (approved):
good.

PR #1927, opened 2023-08-30
be/jvm/runtime: throw an Error in case of a contract failure

maxteufel:
Without any extra work, this also provides a stack trace, which due to the name mangling also contains a readable stack trace for the Fuzion part of the code.


michaellilltokiwa:
Very nice!

michaellilltokiwa (approved):

fridis:
I do not agree with this. Yes, we should use the name mangling in the stack trace in case of contract failure, but throwing an `Error`  is reserved for signalling that there is a serious bug in our implementation, while a contract failure is an error in the code we execute.  Mixing this up will cause additional problems in the future.

PR #1936, opened 2023-09-01
check that conditions do not use effects

michaellilltokiwa:


fridis (commented):
This make sense, but please hold this back, I have a number of ideas of how this should be done, none of which has ripened enough.

michaellilltokiwa:
closing as suggested, since this will be done differently

PR #1944, opened 2023-09-03
fe: fix #1943: Using type parameter as outer type

### LINKED ISSUE #1943: Using type parameter as outer type does not work
Using type parameter as outer type results in crash shown by this small example
```
x is
  y is
    redef as_string => "y in {x.this}"

z : x is

test(T type : x,
     v T.y) =>
  say "T is $T v is $v"

test x x.y
test z z.y
```
which produces a require-condition failure, or a runtime crash if checks are disabled:
```
 > PRECONDITIONS=true POSTCONDITIONS=true ./build/bin/fz bug.fz

error 1: java.lang.Error: require-condition1 failed: dev.flang.ast.Generic:constraint:119
	at dev.flang.util.ANY.require(ANY.java:78)
	at dev.flang.ast.Generic.constraint(Generic.java:119)
	at dev.flang.ast.UnresolvedType.resolveFeature(UnresolvedType.java:725)
	at dev.flang.ast.UnresolvedType.resolve(UnresolvedType.java:692)
	at dev.flang.ast.FunctionReturnType.resolveArgumentType(FunctionReturnType.java:142)
...
```

fridis:
I was unsure when I wrote this code if using a type parameter as an outer type actually makes sense, but it does as shown in the examples of the regression test.

The bug that is patch fixes was that the constraint type of the type parameter was not resolved yet and hence not usable.

### DIFF
--- a/src/dev/flang/ast/UnresolvedType.java
+++ b/src/dev/flang/ast/UnresolvedType.java
@@ -722,7 +722,8 @@ public abstract class UnresolvedType extends AbstractType implements HasSourcePo
         if (o != null && !isThisType() && !o.isThisType())
           {
             o = o.resolve(res, of);
-            var ot = o.isGenericArgument() ? o.genericArgument().constraint() : o;  // NYI: check: do we really want the constraint here?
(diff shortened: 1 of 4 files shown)


fridis:
Merging since I need this tomorrow. Please review anyway.

PR #1950, opened 2023-09-03
parser: disallow line break after `->` in function return type, fix #1945

### LINKED ISSUE #1945: Declaration and call syntax ambiguous for result types that are function types  and lambdas as arguments
Here is a rather lengthy example using declarations and calls, only the very last call to `f` shows the bug
```
ambiguous_declaration is

  # ------- declarations -------

  # field declarations
  a ()->i32 := ()->42
  b () -> i32 := () -> 42
  c () (() ->         # we (currently) need explicit arg list `()` to help parser
        i32  ) := () -> 42
  d () ->
    i32 := () -> 42   # Should be illegal, cannot spread type over several lines!"

  say a.call
  say b.call
  say c.call
  say d.call

  # function declarations
  i ()->i32 is ()->42
  j () -> i32 is () -> 42
  k () (() ->         # we (currently) need explicit arg list `()` to help parser
        i32  ) is () -> 42
  l () ->
    i32 is () -> 42   # Should be illegal, cannot spread type over several lines!"

  say i.call
  say j.call
  say k.call
  say l.call

  # ------- calls -------

  # function to be called
  f(p ()->i32) =>
    say "in f: {p.call}"

  # calls
  f ()->42
  f ()->
    42
  f (() ->
    42)
  f (() ->
    v := 42
    v)
  f ()->
    {
      v := 42
      v
    }

  f ()->       # should be legal
    v := 42
    v
```
the last call produces the following errors
```
 > ./build/bin/fz ambiguous_declaration.fz 

/home/fridi/fuzion/work/ambiguous_declaration.fz:54:5: error 1: Inconsistent indentation
    v
----^
Indentation reference point is /home/fridi/fuzion/work/ambiguous_declaration.fz:6:3:
  a ()->i32 := ()->42
--^
While parsing: exprs, parse stack: exprs, block, implRout, implFldOrRout, routOrField, feature, expr, exprs, unit


/home/fridi/fuzion/work/ambiguous_declaration.fz:53:5: error 2: Type not found
    v := 42
----^
Type 'v' was not found, no corresponding feature nor formal type parameter exists
Type that was not found: 'v'
in feature: 'ambiguous_declaration.f'
To solve this, check the spelling of the type you have used.


/home/fridi/fuzion/work/ambiguous_declaration.fz:52:3: error 3: Incompatible types in assignment
  f ()->       # should be legal
--^
assignment to field : 'ambiguous_declaration.f'
expected formal type: 'Function **error**'
actual type found   : 'ref i32'
assignable to       : 'Any',
                      'num.ref wrap_around',
                      'property.ref equatable',
                      'property.ref hashable',
                      'property.ref orderable',
                      'property.ref partially_orderable',
                      'ref has_interval',
                      'ref i32',
                      'ref integer',
                      'ref numeric'
for value assigned  : 'box(42)'
To solve this, you could change the type of the target 'ambiguous_declaration.f' to 'ref i32' or convert the type of the assigned value to 'Function **error**'.

3 errors.
```
A solution for this would be to require types that are not in parentheses to not be split over several lines. That would make the declarations `d` and `l` in the example illegal, but make sure that the last call is not parsed as a declaration.

fridis:
This is an somewhat ugly special handling but the best solution that I could find so far to avoid the ambiguity between a feature declaration with a function result type and a call with a lambda argument.

Now, it is no longer legal to write declarations such as
```
  f () ->
    i32 := ()->42
```
one must use either
```
  f (() ->
     i32  ) := ()->42
```
or
```
  f () -> i32 := ()->42
```
while a call
```
  f ()->
    x := 42
```
can not be written as
```
  f ()->x:=42
```
This patch adds a positive and a negative regression test for #1945.


fridis:
Merging this since I need it for examples in my TyDe talk. Please review anyway.

PR #1960, opened 2023-09-06
Expr as string

### LINKED ISSUE #1670: `Expr.toString()` should not be used in comiler error messages
Currently, `Expr.toString()` attempts to re-construct the source code, but this is doomed to fail since loads of information (where to place parentheses, white space, etc.) is lost and added (e.g., following outer references to find the correct outer instance). This results in ugly and strange error messages.  In particular, if the expression was generated artificially for syntax sugar like lambdas or lazy arguments, this gets fully confusing. 

Instead, wherever an expression is presented in an error message, it should best be taken from the source code directly. One way to do this would be to record the start and the end position of an expression and extract the exact source code when creating the error. It might make sense to not replace the `SourcePosition` given by `Expr.pos()`, by a `SourceSection` that might be stored by the user of the expression.

Here is an example:

```
expr is

  a is
    b is
      c =>
        s String := ((( xyz + 123 )))

  xyz := 456

  a.b.c
```
that produces

```
 > ./build/bin/fz expr.fz 

/home/fridi/fuzion/clean/fuzion.1/expr.fz:6:9: error 1: Incompatible types in assignment
        s String := ((( xyz + 123 )))
--------^
assignment to field : 'expr.a.b.c.s'
expected formal type: 'String'
actual type found   : 'ref i32'
assignable to       : 'Any',
                      'ref equatable',
                      'ref has_hash',
                      'ref has_interval',
                      'ref has_partial_order',
                      'ref has_total_order',
                      'ref i32',
                      'ref integer',
                      'ref numeric',
                      'ref wrapping_integer'
for value assigned  : 
{
  {
    {
      box(deref(deref(deref(c.this.#^expr.a.b.c).#^expr.a.b).#^expr.a).xyz.infix + 123)
    }
  }
}

To solve this, you could change the type of the target 'expr.a.b.c.s' to 'ref i32' or convert the type of the assigned value to 'String'.

one error.
```
In particular, the `for value assigned` section is bad, should maybe be
```
for value assigned  : ((( xyz + 123 )))
```
or even
```
for value assigned  : /home/fridi/fuzion/clean/fuzion.1/expr.fz:6:9: error 1: Incompatible types in assignment
        s String := ((( xyz + 123 )))
--------------------^^^^^^^^^^^^^^^^^
```
or using even nicer ASCII-graphics with ANSI colors to highlight the code.

michaellilltokiwa:
In error messages, use source text of expressions not a string representation of the expression like it was done before.
fixes #1670 

Blocks do not have a source position themselves anymore but return the range from beginning of first to end of last expression.

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -830,7 +830,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
         var typeFeature = new Feature(p, visibility().typeVisibility(), 0, NoType.INSTANCE, new List<>(name), typeArgs,
                                       inh,
                                       Contract.EMPTY_CONTRACT,
-                                      new Impl(p, new Block(p, new List<>()), Impl.Kind.Routine));
+                                      new Impl(p, new Block(new List<>()), Impl.Kind.Routine));
         typeFeature._typeFeatureOrigin = this;
         res._module.addTypeFeature(outerType, typeFeature);
         result = typeFeature;
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -123,11 +123,7 @@ public class AstErrors extends ANY
   }
   static String s(Expr e)
   {
-    return expr(e.toString());
-  }
-  static String sle(List<Expr> e)
-  {
-    return expr(e.toString());
+    return expr(e.pos().sourceText());
   }
   static String s(AbstractAssign a)
   {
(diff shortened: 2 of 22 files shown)


fridis (approved):
This patch looks like magic, simplifying the code and enhancing the output. 

One idea for a future enhancement: It might be good to store with each Expr the whole source range of the Expr, maybe the Parser could do this after parsing an expr when it is returned. To illustrate what I want: The old cold produced output like
```
In call: 'neg.this.x.prefix -'
In call: 'd tuple'
```
with your patch, these become
```
In call: '-'
In call: 'd tuple'
```
what I would like to see is
```
In call: '-x'
In call: 'd ()'
```
But this can be done next.

PR #1969, opened 2023-09-11
ast: implement Comparable for Generic

michaellilltokiwa:
also improve readability of AbstractType.compareToIgnoreOuter


fridis (approved):
nice.

PR #1970, opened 2023-09-11
lib/io: use `fuzion.sys.Pointer` for references to arrays' internal data

maxteufel:
This will avoid accidentally supplying values other than the internal data structure of an array to these features, and thus make it a little bit safer to work on the standard library.


michaellilltokiwa (approved):

fridis (commented):
good.

PR #1972, opened 2023-09-12
be/c, dfa: clean up handling of native features

maxteufel:
This also changes the semantics of the native features. Native features that used to return `String` will no longer work, they now have to have the more precise result type `Const_String`. Likewise, features that previously accepted an `Any` argument need to be updated to accept an argument of type `fuzion.sys.Pointer`.


michaellilltokiwa (approved):

fridis (commented):
nice.

fridis (commented):
nice.

PR #1976, opened 2023-09-13
be/jvm: create JAR files instead of a collection of class files

maxteufel:
A first look into generating JAR files from the JVM backend. @fridis, I want to know what you think about this...

### DIFF
--- a/src/dev/flang/be/jvm/JVM.java
+++ b/src/dev/flang/be/jvm/JVM.java
@@ -40,6 +40,7 @@ import dev.flang.util.Errors;
 import dev.flang.util.List;
 import dev.flang.util.Pair;
 
+import java.io.FileOutputStream;
 import java.io.IOException;
 
 import java.nio.charset.StandardCharsets;
@@ -47,6 +48,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 
 import java.util.ArrayList;
+import java.util.jar.Attributes;
+import java.util.jar.JarEntry;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
 import java.util.stream.Stream;
 
 
@@ -483,7 +488,7 @@ should be avoided as much as possible.
         jvm._runner.runMain(applicationArgs);
       }
     },
-    SAVE {
+    SAVE_CLASSES {
       boolean condition(JVM jvm)
       {
         return jvm._options._saveClasses;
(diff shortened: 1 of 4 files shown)


src/dev/flang/be/jvm/JVM.java
```diff
@@ -490,18 +491,50 @@ boolean condition(JVM jvm)
...
-            catch (IOException io)
-              {
-                Errors.error("JVM backend I/O error",
-                             "While creating directory '" + PATH_FOR_CLASSES + "', received I/O error '" + io + "'");
+                jvm._jos.putNextEntry(new JarEntry(d));
+                jvm._jos.write(Files.readAllBytes(jvm._options.fuzionHome()
+                                                              .resolve("classes")
+                                                              .resolve(d)
+                                                              .normalize()
+                                                              .toAbsolutePath()));
```
fridis, line 530: This is ok for now, but eventually I would like to to have a `build/jars/fzrt.jar` created by Makefile that contains these classes and add a dependency in the generated `.jar` file instead of duplicating the classes.

src/dev/flang/be/jvm/JVM.java
```diff
@@ -600,6 +645,8 @@ void finish(JVM jvm)
 
   Runner _runner;
 
+  JarOutputStream _jos;
+
```
fridis, line 649: Please add a comment what this is (same for my code where these comments are missing, but in a separeate patch).

src/dev/flang/be/jvm/classfile/ClassFile.java
```diff
@@ -942,17 +944,12 @@ public Path classFile()
...
    */
-  public void write(Path dir) throws IOException
+  public void write(JarOutputStream jos) throws IOException
   {
-    var fp = dir.resolve(classFile());
-
-    if (_verbose) // NYI: Use Options.verbosePrintln?
-      {
-        System.out.println(" + " + fp);
-      }
```
fridis, line 954: This output should be there for the `.jar` file.

src/dev/flang/be/jvm/JVM.java
```diff
@@ -510,28 +543,40 @@ void compile(JVM jvm, int cl)
...
-                             "While creating class '" + ci.classFile() + "' in '" + PATH_FOR_CLASSES + "', received I/O error '" + io + "'");
+                             "While creating class '" + ci.classFile() + "' in JAR, received I/O error '" + io + "'");
               }
           }
       }
+      void finish(JVM jvm)
+      {
+        try
+          {
+            jvm._jos.close();
```
fridis, line 572: Would be great to also create a shell script with the base name of the main clazz that performs `java -jar <baseName>.jar`. Same should be done for `-classes` (in a speparate patch).

fridis (changes requested):
This looks good. 
A number of minor suggestions and requests for additional PRs:
I would like to keep the `-classes` backend to create classes and have a `-jar` backend to create a jar file. Can you change the patch and split `SAVE` into two phases `SAVE_CLASSES` and `SAVE_JAR` and add the jar option to JVMOptions and Fuzion.java?
Also, please add the verbose output of ` + xyz.jar`.

src/dev/flang/be/jvm/JVM.java
fridis: Will add a PR for `_numLocals` documentation, that is really very cryptic.

src/dev/flang/be/jvm/JVM.java
fridis: No, I did not know binfmt_misc so far. Seems useful, needs to be set up by the user. I would like to provide an easy and ideally uniform way to run the created code, independent of the backend (C/jar/classes/...).

fridis (approved):

PR #1977, opened 2023-09-13
add pom.xml

michaellilltokiwa:
In my tests compiling with:
`mvn -T 1C compile`: ~7s
`make build/bin/fz` : ~18s

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 build
 fuzion_generated_clazzes
+target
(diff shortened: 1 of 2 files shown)


fridis:
This is an impressive speedup, but I do not understand where it comes from and for me .pom files are just gibberish that I do not understand, so I need help here. 
My experience with mvn so far was that it downloads tons of packages from all around the world and I completely lose control of what is happening. 
So the only blocking factor for this PR is my lack of understanding what mvn does and how to control it.

michaellilltokiwa:
I could/should have explained that `-T 1C` means one thread per cpu core. So this uses all available cpu-cores when possible. This might be one reason why it is so much faster but this is just a guess since I also do not understand the maven magic.

michaellilltokiwa:
@fridis @maxteufel I'm not sure what to do with this? My idea would be to keep `make` as the main build tool, but optionally have make use mvn for building java. But I'm not sure if it's worth the hassle of maintaining... What do you think?

michaellilltokiwa:
@tokiwa-software/developers  Okay, let's stick with `make` then.

PR #2007, opened 2023-09-21
be/jvm: Expr: consistently use `JavaType` as an argument

maxteufel:
Previously some methods, such as `getstatic`, accepted a `JavaType` argument, while other methods, such as `putstatic`, accepted type information by passing a `String`. Change the latter to take a `JavaType` argument, to make the usage more consitent.


fridis (approved):
good.

PR #2009, opened 2023-09-21
air/me: mark clazzes/feature as instantiated/used/called programmatically

michaellilltokiwa:
- manually marking clazzes as called/instantiated
- marking i8 and other features as used
- move visit(s) to `AbstractConstant`


fridis (approved):
Nice. 
One detail that should be fixed, maybe in a separate PR: the precondition should not be part of the code of a feature like the postcondition is. Instead. the precondition is part of a call to a feature. This becomes important, e.g., when there is a precondition of an abstract feature: This is precondition will be executed on a call to that abstract feature, while the call will end up in a redefinition that might have a different (weaker) precondition (while Fuzion currently does not enforce that preconditions are weakened and postconditions are strengthened in redefinitions).

PR #2013, opened 2023-09-22
array constants with Fridi's fixes

fridis:


fridis:
This is part of #2021 now.

PR #2018, opened 2023-09-22
Const arrays in jvm backend

fridis:


fridis:
replaced by PR #2012.

PR #2020, opened 2023-09-25
be/jvm: implement `fuzion.java.Java_Object.is_null` intrinsic

maxteufel:
depends on #2001


fridis (approved):
good. 

It would be better to mark `fuzion.java.object`/`object_ref` as used only if they the corresponding intrinsic is used, but this can be done in a separate step.

Issue #2046, opened 2023-10-03
Framework for Function composition


fridis:
Languages like BQN or Uiua provide powerful operators for function composition using single character operators that results in extremely dense code. Unfortunately, the code is also very cryptic and practically incomprehensible for outsiders and surely not in the spirit for Fuzion. 

Nevertheless, the basic concepts of functional composition seems useful and it might make sense to provide more features for function composition than just `Unary.compose`.
What we need is 

- find the set of composition functions we consider useful and needed
- write examples that use these and compare named functions an infix / prefix / postfix operators to decide which is most useful.
- check if type inference is powerful enough or needs to be enhanced
- check if any other syntax sugar might be useful, e.g., `f+g` as short for `x->(f x)+(g x)` or `f [1,2,3]` for `[1,2,3].as_list.map f`.

Goal should be 

- readability and clarity
- compactness without being cryptic
- avoiding accidental errors

Here is a collection of resources on composition

- [combinarorylogic.com](https://combinatorylogic.com/table.html)
- Talk at Strange Loop 2016 by Amar Shah: [Point-Free or Die: Tacit Programming in Haskell and Beyond](https://www.youtube.com/watch?v=seVSlKazsNk)
- Talk at CppNorth 2023 by Conor Hoekstra: [Function Composition in Programming Languages - Conor Hoekstra - CppNorth 2023](https://www.youtube.com/watch?v=JELcdZLre3s)
- Paper by Conor Hoekstra: [Combinatory Logic and Combinators in Array Languages](https://web.archive.org/web/20220617020347id_/https://dl.acm.org/doi/pdf/10.1145/3520306.3534504)
- [BQN tutorial](https://mlochbaum.github.io/BQN/tutorial/index.html)

PR #2047, opened 2023-10-03
lib: Add collection of features for function composition, #2046

fridis:
This is not intended to be useful at this point, it is just a first try to collect basic composition features that appear in literature and other languages.

To make this useful, we need to write example code using these and play around with it. Maybe we need to move these into Unary.fz, maybe add Binary.fz, maybe make some operators or add operator aliases, maybe add features or operators to the universe, etc. Maybe we want these in a separate module?

### DIFF
--- a/lib/composition.fz
+++ b/lib/composition.fz
@@ -1,0 +1,142 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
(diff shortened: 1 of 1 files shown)


lib/composition.fz
```diff
@@ -0,0 +1,142 @@
...
+
+  # I
+  # identity
+  # BQN: ⊣⊢
+  # Haskell id
+  id(T type) T -> T is x->x
+
+  # K
+  # Elementary Cancellator
+  # Kestrel
```
michaellilltokiwa, line 56: Bird:

michaellilltokiwa:
Interesting but imho hard to review without seeing these used in some examples.

PR #2049, opened 2023-10-04
be/jvm: Minor fixes in design of JVM backend

fridis:


fridis:
Forgot to publish this PR earlier. Just minor fixes in the text. Will need to review the design to match the actual implementation of the JVM backend once it is stable.

PR #2067, opened 2023-10-09
support for turning calls into compile time constants

michaellilltokiwa:
```
ex is

  a := (u32 65, codepoint 97)
  say a.values.0
  say a.values.1
```
`a` instantiation is compiled to:
```c
    fzCur->fzF_0_a = (fzT_2tuple_w_u32_w_codepoint){.fzF_0_values = 65U,.fzF_1_values = (fzT_1codepoint){.fzF_0_val = 97U}};
```


src/dev/flang/ast/AbstractCall.java
```diff
@@ -125,6 +128,80 @@ public boolean isDynamic()
...
+        throw new UnsupportedOperationException("Unimplemented method 'visit'");
+      }
+    };
+  }
+
+
+  @Override
+  public boolean isCompileTimeConst()
+  {
+    var result =
```
michaellilltokiwa, line 190: Some of these may be redundant or could be relaxed.

src/dev/flang/fuir/FUIR.java
```diff
@@ -2410,6 +2433,52 @@ public boolean hasPrecondition(int cl)
...
+  public boolean callCanBeConstant(int cl, int c, int ix)
+  {
+    if (PRECONDITIONS) require
+      (_codeIds.get(c).get(ix) instanceof AbstractCall);
+
+    var cc = accessedClazz  (cl, c, ix);
+
+    var e = _codeIds.get(c).get(ix);
+    return ((AbstractCall)e).isCompileTimeConst() && !dependsOnOuterRef(cc)
+      // NYI can we allow none universe targets in some cases?
```
michaellilltokiwa, line 2448: If call has a target that is not universe, what can we do about it?

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -1755,6 +1750,53 @@ Value newConstString(byte[] utf8Bytes, Context context)
...
+              var bytes = b.slice(offset, n);
+              offset += n;
+              var arg = newValueConst(fr, context, bytes);
+              args.add(arg);
+              result.setField(this, f, arg);
+            }
+
+          // register calls for constant creation even though
+          // not every backend actually performs these calls.
+          newCall(constCl, false, _universe, args, null /* new environment */, context);
```
michaellilltokiwa, line 1792: This needed for the JVM backend otherwise `new0(constCl)` will not work because it has not been generated

src/dev/flang/ast/AbstractFeature.java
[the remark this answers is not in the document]
michaellilltokiwa: This should better be in another PR. I believe `inherits()` to be never empty, so this ternary is useless...

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -498,12 +498,7 @@ public Pair<Val, Unit> constData(int constCl, byte[] d)
              c_array_f32  ,
              c_array_f64  -> newConstArray(constCl, d, _call);
         case c_Const_String -> newConstString(d, _call);
-        default ->
-        {
-          Errors.error("Unsupported constant in DFA analysis.",
-                       "DFA cannot handle constant of clazz '" + _fuir.clazzAsString(constCl) + "' ");
-          yield null;
-        }
+        default -> newValueConst(constCl, _call, ByteBuffer.wrap(d));
```
fridis, line 501: I would keep the error case for `_fuir.clazzIsChoice(constCl)`.

src/dev/flang/be/jvm/CodeGen.java
```diff
@@ -972,11 +972,50 @@ Pair<Expr, Expr> createConstant(int constCl, byte[] d)
...
+  private Expr valueConst(int constCl, ByteBuffer b)
+  {
+    return switch (_fuir.getSpecialId(constCl))
+      {
+        // we reached the "bottom" of this value const
+      case c_i8, c_i16, c_i32, c_i64, c_u8, c_u16, c_u32, c_u64, c_f32, c_f64 ->
         {
-          Errors.error("Unsupported constant in JVM backend.",
-                       "Backend cannot handle constant of clazz '" + _fuir.clazzAsString(constCl) + "' ");
-          yield null;
```
fridis, line 979: I would keep the error case for `_fuir.clazzIsChoice(constCl)`.

src/dev/flang/ast/AbstractConstant.java
```diff
@@ -75,6 +75,25 @@ public Expr visit(FeatureVisitor v, AbstractFeature outer)
   }
 
 
+  /**
+   * Is this a compile-time constant?
+   */
+  @Override
+  public boolean isCompileTimeConst()
+  {
+    return !type().isRef(); // String is ref but NYI
```
fridis, line 84: I do not understand this NYI comment.

fridis (changes requested):
Very nice and clean patch compared with the big impact this has!

One case I am unsure of is this:
```
  side_effect is say "this does something!"
  point(x,y i32) : side_effect is
    as_string => "$x,$y"
  c := point 3 4
  say c
```
Does this create a constant out of `point 3 4`? If so, this might be a bug.

Some more minor comments in the code.

michaellilltokiwa:
> Very nice and clean patch compared with the big impact this has!
> 
> One case I am unsure of is this:
> 
> ```
>   side_effect is say "this does something!"
>   point(x,y i32) : side_effect is
>     as_string => "$x,$y"
>   c := point 3 4
>   say c
> ```
> 
> Does this create a constant out of `point 3 4`? If so, this might be a bug.
> 
> Some more minor comments in the code.

I now check if the inherited features code contains only declarations.

fridis (approved):

PR #2087, opened 2023-10-16
ast: move typeIfKnown from Block to AbstractBlock

michaellilltokiwa:
Also `AbstractBlock.typeIfKown` always returned `unit` which is kind of odd...


fridis (approved):
good.

PR #2089, opened 2023-10-17
be/jvm: introduce a null type that is compatible to all types

maxteufel:


src/dev/flang/be/jvm/classfile/ClassFileConstants.java
```diff
@@ -604,6 +604,14 @@ public boolean sameAs(ClassType other)
   static ClassType JAVA_LANG_OBJECT = new ClassType("java/lang/Object");
   static ClassType JAVA_LANG_STRING = new ClassType("java/lang/String");
 
+  static ClassType NULL_TYPE = new ClassType("java/lang/Object") {
+    @Override
+    public boolean sameAs(ClassType other)
+    {
+      return true;
+    }
```
fridis, line 612: `sameAs` should be symmetric. Better add a special case in `CodeGen.drop`: `|| jt == NULL_TYPE`.

fridis (changes requested):
Looks good. Just a suggestion to keep `sameAs` symmetric.

michaellilltokiwa (approved):

fridis (approved):

PR #2092, opened 2023-10-17
workaround for failing test _stdin_ in main branch

michaellilltokiwa:
tests/stdin currently fails for interpreter and jvm backend. This PR works around the problem until it is fixed.


michaellilltokiwa:
Will create an issue, I didn't do it yet because I don't understand why it fails yet.

PR #2094, opened 2023-10-18
fix #2093 always recreate array constants

### LINKED ISSUE #2093: array constants don't always work correctly
the bug is reproducible in interpreter and jvm backend when reverting #2092 and running test _stdin_

michaellilltokiwa:


michaellilltokiwa:
#2093 has been fixed differently

PR #2099, opened 2023-10-18
jvm: add checks to trap accidental modifications of immutable array data

fridis:
This turns the precondition failure in `test/stdin` into a nicer error saying `error 1: Attempt to modify immutable array`.

The way this works is that all internal array data that are put into an instance of an (immutable) array will be marked as frozen using a weak map of all frozen array data pointers. On modification, it will then be checked that the modified array is not in that map.

These checks are currently enabled only if CHECKS is set. In case debugging is disabled, eventually no code should be generated for this.

### DIFF
--- a/src/dev/flang/be/jvm/Intrinsix.java
+++ b/src/dev/flang/be/jvm/Intrinsix.java
@@ -447,20 +447,54 @@ public class Intrinsix extends ANY implements ClassFileConstants
             }
           else
             {
-              var arrAndIndex = args.get(0)
-                .andThen(Expr.checkcast(jt.array()))
-                .andThen(args.get(1));
               if (in.equals("fuzion.sys.internal_array.get"))
                 {
-                  val = arrAndIndex
+                  val = args.get(0)
+                    .andThen(Expr.checkcast(jt.array()))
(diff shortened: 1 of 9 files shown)


src/dev/flang/be/jvm/Intrinsix.java
```diff
@@ -447,24 +447,58 @@ else if (jvm._fuir.clazzIsChoice(rc) && jt instanceof ClassFileConstants.AType)
...
+                    {
+                      check_frozen = Expr.DUP
+                        .andThen(Expr.invokeStatic(Names.RUNTIME_CLASS,
+                                                   "ensure_not_frozen",
+                                                   "(" + (JAVA_LANG_OBJECT.descriptor()) + ")V",
+                                                   ClassFileConstants.PrimitiveType.type_void));
+                    }
+                  code = args.get(0)
+                    .andThen(Expr.checkcast(jt.array()))
+                    .andThen(check_frozen)
```
michaellilltokiwa, line 470: code for check_frozen seems to be generated regardless checks?

michaellilltokiwa (approved):

PR #2102, opened 2023-10-19
move RefOrVal to AbstractType

michaellilltokiwa:


fridis (changes requested):
It is good to cleanup this part for the AST code, the current handling of `RefOrVal` is somewhat clumsy. 

However, `RevOrVal` should only be used for unresolved types until a module file is written. For types loaded form a module file, RefOrVal should no longer be used (it, however, is used in `dev.flang.fe.GenericType`, but this can probably be removed). There is no method declard in `AbstractType` or `AbstractFeature` that uses `RefOrVal` in its signature, and this is intentional because I want to get rid of this. 

My suggestion: Try to remove the uses of `RefOrVal` within `dev.flang.fe.GenericType`. I think only two values, `Boxed` and `LikeUnderlyingFeature` are used there, so coud be done with one `boolean`. Then, you should be able to make `RefOrVal` package private for `dev.flang.ast` and then, for convenience, it may be moved to `AbstractType.java` or maybe even be its own class.

fridis (approved):
very nice.

PR #2106, opened 2023-10-19
fix #2103, non-disjoint types in choice are reported twice

### LINKED ISSUE #2103: Non-disjoint types in choice are reported twice
This small example
```
c : choice TRUE TRUE is
```
correctly reports an error, but incorrectly reports it twice:
```
 > ../../bin/fz unit_like.fz

unit_like.fz:1:1: error 1: Actual type parameters to choice type must be disjoint types
c : choice TRUE TRUE is
^
The following types have overlapping values:
'TRUE'
'TRUE'


<built-in>: error 2: Actual type parameters to choice type must be disjoint types
The following types have overlapping values:
'TRUE'
'TRUE'

2 errors.
```

michaellilltokiwa:
fixes #2103
This `if` is a bit ugly IMHO... Maybe there is a better way?

### DIFF
--- a/src/dev/flang/ast/Feature.java
+++ b/src/dev/flang/ast/Feature.java
@@ -1678,7 +1678,12 @@ public class Feature extends AbstractFeature
             AstErrors.failedToInferResultType(this);
             _resultType = Types.t_ERROR;
           }
-        _resultType.checkChoice(_posOfReturnType);
+
+        if (!isTypeParameter())
+          {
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
fixes #2103

fridis:
I think the reason for this is that for type parameter, the result type is (ab-)used as the type constraint.

fridis (approved):

PR #2107, opened 2023-10-19
handle ambigous assignment via subtype

### LINKED ISSUE #1865: Choice for `String` and `Sequence u8` behaves strange and creates broken code in C backend
This example
```
test_rau is

  t : choice String (Sequence u8) nil unit is

  x(v t) =>
      match v
        s String => say "Str"
        t Sequence => say "Seq"
        nil => say "nil"
        unit => say "unit"

  x ""
  x "".utf8
  x nil
  x unit
```
results in double `Str`:
```
 > ./build/bin/fz test_rau.fz  
Str
Str
nil
unit
```
and not-compilable C code:
```
 > ./build/bin/fz -c test_rau.fz 
test_rau.c:18379:16: error: duplicate case value '884'
          case 884/* ref Const_String */:
               ^
test_rau.c:18361:16: note: previous case defined here
          case 884/* ref Const_String */:
               ^
1 error generated.

error 1: C backend: C compiler failed
C compiler call 'clang -Wall -Werror -Wno-gnu-empty-struct -Wno-unused-variable -Wno-unused-label -Wno-unused-function -Wno-infinite-recursion -Wno-unused-but-set-variable -O3 -lm -lpthread -o test_rau test_rau.c' failed with exit code '1'

one error.
```

michaellilltokiwa:
fixes #1865


fridis (changes requested):
Good catch! 
This is so important that I would like to see this added to a negative tests. Maybe `tests/choice_negative` could be extended to cover this case?
Also, would be good to have an issue for this.

fridis (approved):

Issue #957, opened 2023-01-17
ClassCastException when trying to option type with generic feature


fridis:
See this example: 
```
 > cat test_get_type.fz

gt0(T type, x T) => T
Any.gt => gt0 Any.this

say (option 32).gt
 > ./build/bin/fz test_get_type.fz

error 1: java.lang.ClassCastException: class dev.flang.fe.NormalType cannot be cast to class dev.flang.ast.Type (dev.flang.fe.NormalType and dev.flang.ast.Type are in unnamed module of loader 'app')
	at dev.flang.air.Clazz.determineActualGenerics(Clazz.java:2313)
	at dev.flang.air.Clazz.dependencies(Clazz.java:409)
	at dev.flang.air.Clazzes.create(Clazzes.java:388)
	at dev.flang.air.Clazzes.clazzWithSpecificOuter(Clazzes.java:1235)
	at dev.flang.air.Clazz.lookup(Clazz.java:1162)
	at dev.flang.air.Clazzes.findClazzes(Clazzes.java:887)
	at dev.flang.air.Clazz$1.action(Clazz.java:1417)
	at dev.flang.ast.Stmnt.visitStatements(Stmnt.java:72)
	at dev.flang.ast.AbstractCall.visitStatements(AbstractCall.java:123)
	at dev.flang.ast.AbstractBlock.visitStatements(AbstractBlock.java:101)
	at dev.flang.ast.AbstractAssign.visitStatements(AbstractAssign.java:140)
	at dev.flang.ast.AbstractBlock.visitStatements(AbstractBlock.java:101)
	at dev.flang.ast.AbstractFeature.visitStatements(AbstractFeature.java:1330)
	at dev.flang.air.Clazz.inspectCode(Clazz.java:1367)
	at dev.flang.air.Clazz.findAllClasses(Clazz.java:1411)
	at dev.flang.air.Clazzes.findAllClasses(Clazzes.java:455)
	at dev.flang.fuir.FUIR.<init>(FUIR.java:196)
	at dev.flang.opt.Optimizer.fuir(Optimizer.java:73)
	at dev.flang.tools.Fuzion$Backend.processFrontEnd(Fuzion.java:367)
	at dev.flang.tools.Fuzion.lambda$parseArgsForBackend$3(Fuzion.java:832)
	at dev.flang.tools.Tool.lambda$run$0(Tool.java:154)
	at dev.flang.util.Errors.runAndExit(Errors.java:704)
[...]

Issue #1561, opened 2023-06-01
Incompatible types, actual type found: long.feature.name.#anonymous0


michaellilltokiwa:
This example results in a very long and confusing error message:
```
ex =>
  read_line option String is

    (try ()->
      ref : String
        utf8 Sequence u8 is
          (Sequence u8).type.empty
    ).as_option
```
```
/home/sam/playground/test.fz:54:7: error 1: Incompatible types in assignment
    ).as_option
------^
assignment to field : 'ex.read_line.result'
expected formal type: 'option String'
actual type found   : 'option (((ex#0.this.type.read_line#0.this.type.value #fun0.this.type).call#0.this.type).#anonymous0)'
assignable to       : 'option (((ex#0.this.type.read_line#0.this.type.value #fun0.this.type).call#0.this.type).#anonymous0)'
for value assigned  : 
{
  {
    try ex.this.type.read_line.this.type.#fun0.this.type.call.this.type.#anonymous0 read_line.this.#fun0
  }.as_option
}

To solve this, you could change the type of the target 'ex.read_line.result' to 'option (((ex#0.this.type.read_line#0.this.type.value #fun0.this.type).call#0.this.type).#anonymous0)' or convert the type of the assigned value to 'option String'.

one error.
```

Issue #1559, opened 2023-06-01
ensure condition failure in Clazzes.clazz: thiz.isRef() == result._type.isRef()


michaellilltokiwa:
```
a option String is
  mutate.go ()->
    ref : String
      utf8 => (Sequence u8).type.empty
```
```
error 2: java.lang.Error: ensure-condition failed: dev.flang.air.Clazzes:clazz:1235
	at dev.flang.util.ANY.ensure(ANY.java:236)
	at dev.flang.air.Clazzes.clazz(Clazzes.java:1235)
	at dev.flang.air.Clazz.actualClazz(Clazz.java:678)
	at dev.flang.air.Clazz.determineActualGenerics(Clazz.java:2358)
	at dev.flang.air.Clazz.dependencies(Clazz.java:417)
	at dev.flang.air.Clazzes.create(Clazzes.java:380)
	at dev.flang.air.Clazzes.create(Clazzes.java:279)
	at dev.flang.air.Clazzes.clazz(Clazzes.java:1230)
	at dev.flang.air.Clazz.actualClazz(Clazz.java:678)
	at dev.flang.air.Clazz.determineResultClazz(Clazz.java:2259)
	at dev.flang.air.Clazz.dependencies(Clazz.java:419)
	at dev.flang.air.Clazzes.create(Clazzes.java:380)
	at dev.flang.air.Clazzes.clazzWithSpecificOuter(Clazzes.java:1263)
	at dev.flang.air.Clazz.lookup(Clazz.java:1305)
	at dev.flang.air.Clazz.lookup(Clazz.java:1076)
	at dev.flang.air.Clazz.lookup(Clazz.java:1049)
	at dev.flang.air.Clazz.actualFields(Clazz.java:2523)
	at dev.flang.air.Clazz.determineArgumentFields(Clazz.java:2316)
	at dev.flang.air.Clazz.dependencies(Clazz.java:416)
	at dev.flang.air.Clazzes.create(Clazzes.java:380)
	at dev.flang.air.Clazzes.clazzWithSpecificOuter(Clazzes.java:1263)
	at dev.flang.air.Clazz.lookup(Clazz.java:1305)
	at dev.flang.air.Clazzes.clazz(Clazzes.java:1123)
	at dev.flang.air.Clazzes.findClazzes(Clazzes.java:1033)
	at dev.flang.air.Clazz$1.action(Clazz.java:1578)
[...]

Issue #1216, opened 2023-03-13
Small example causes require condition failure in `Box.<init>`


fridis:
This
```
outer is
  f(minime outer.this.type) =>
  b ref : outer is
  f b
```
causes
```
 > PRECONDITIONS=true POSTCONDITIONS=true ./build/bin/fz box_require.fz

error 1: java.lang.Error: require-condition2 failed: dev.flang.ast.Box:<init>:80
	at dev.flang.util.ANY.require(ANY.java:94)
	at dev.flang.ast.Box.<init>(Box.java:80)
	at dev.flang.ast.Expr.box(Expr.java:390)
	at dev.flang.ast.Call.box(Call.java:2143)
	at dev.flang.ast.Feature$6.action(Feature.java:1705)
	at dev.flang.ast.Feature$6.action(Feature.java:1703)
	at dev.flang.ast.Call.visit(Call.java:1056)
	at dev.flang.ast.Call.visit(Call.java:46)
	at dev.flang.ast.Block.visit(Block.java:226)
	at dev.flang.ast.Block.visit(Block.java:40)
	at dev.flang.ast.Impl.visit(Impl.java:337)
	at dev.flang.ast.Feature.visit(Feature.java:1027)
	at dev.flang.ast.Feature.box(Feature.java:1703)
	at dev.flang.ast.Resolution.resolveOne(Resolution.java:390)
	at dev.flang.ast.Resolution.resolve(Resolution.java:338)
	at dev.flang.fe.SourceModule.createASTandResolve(SourceModule.java:264)
	at dev.flang.fe.FrontEnd.<init>(FrontEnd.java:178)
	at dev.flang.tools.Fuzion.lambda$parseArgsForBackend$3(Fuzion.java:830)
	at dev.flang.tools.Tool.lambda$run$0(Tool.java:154)
	at dev.flang.util.Errors.runAndExit(Errors.java:704)
	at dev.flang.tools.Tool.run(Tool.java:154)
	at dev.flang.tools.Fuzion.main(Fuzion.java:481)


*** fatal errors encountered, stopping.
```
but produces a proper error message with precondition checks disabled
```
 > ./build/bin/fz box_require.fz

/home/fridi/fuzion/clean/fuzion.1/box_require.fz:4:5: error 1: Incompatible types when passing argument in a call
  f b
----^
Actual type for argument #1 'minime' does not match expected type.
In call to          : 'outer.f'
expected formal type: 'outer.this.type'
[...]

PR #2119, opened 2023-10-23
fix issues, add regression tests: #957, #960, #1051, #1216, #1559, #1561

michaellilltokiwa:
fixes: #957, #960, #1051, #1216, #1559, #1561


tests/reg_issue960_npe_resolve_syntactic_sugar2/test_issue960.fz.expected_err
```diff
@@ -0,0 +1,16 @@
...
+
+--CURDIR--/test_issue960.fz:8:12: error 2: No type information can be inferred from a lambda expression
+    f := ()->x
+-----------^
+A lambda expression can only be used if assigned to a field or argument of type 'Function'
+with argument count of the lambda expression equal to the number of type parameters of the type.  The type of the
+assigned field must be given explicitly.
+To solve this, declare an explicit type for the target field, e.g., 'f (i32, i32) -> bool := x, y -> x > y'.
+
+2 errors.
```
fridis, line 16: This could be another enhancement request, or ma<be even two

- first, this example should produce only one error, not two
- second, it should actually be possible to infer the type of a nullary lambda expression as in this case, problematic are only the arguments.
Ok, for now, but can you add issues for these and maybe refer to them in the test?

fridis (changes requested):
Very nice to close all these issues!
Just a small change request (headers) and new issues to be opened to be fixed later.  
Please add our GPL-headers to the Fuzion sources you added for the tests.
Also a comment for new issues to be added.

fridis (approved):

PR #2124, opened 2023-10-24
improve inference of array type in if/else

### LINKED ISSUE #1899: inference array void
type inference should almost never infer `array void`.
example:
```
ex =>

  a =>
    if true
      []
    else
      [3]
```
```
/home/sam/playground/test.fz:4:5: error 1: Incompatible types in branches of if expression
    if true
----^
block returns value of type 'array void' at /home/sam/playground/test.fz:5:7:
      []
------^
block returns value of type 'array i32' at /home/sam/playground/test.fz:7:7:
      [3]
------^
```

michaellilltokiwa:
fixes #1899


fridis (commented):
I fear this might break some cases. Can you add the following to the added tests just to verify:
```
  say []
  say (type_of [])
  say (type_of ([],[42]))
```
Can you think of other strange cases?

michaellilltokiwa:
@fridis I would argue that it's actually a good thing that is breaks those cases since the programmer clearly did not supply enough information? `array void` when using syntax sugar `[]` is probably never what is intended?
```
/home/sam/playground/test.fz:3:8: error 1: Failed to infer actual type parameters
  say (type_of [])
-------^^^^^^^
In call to 'type_of', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'T'
Type inference failed for one type parameter 'T'


/home/sam/playground/test.fz:4:16: error 2: Failed to infer open type parameter type from actual argument.
  say (type_of ([],[42]))
---------------^
Type of second actual argument could not be inferred at /home/sam/playground/test.fz:4:17:
  say (type_of ([],[42]))
----------------^


/home/sam/playground/test.fz:2:7: error 3: Failed to infer type of expression.
  say []
------^
Expression with unknown type: ''


/home/sam/playground/test.fz:3:16: error 4: Failed to infer type of expression.
  say (type_of [])
---------------^
Expression with unknown type: ''

4 errors.
```

fridis:
You managed to convince me.

fridis (approved):

PR #2125, opened 2023-10-24
improve type inference, arg result type was inferenced to void instea…

### LINKED ISSUE #1918: type is inferenced to `void`, when it should be `i32`
```
test =>

  count(n) =>
    true

  (1..4)
    .map_sequence (_)->
      concur.thread.spawn ()->
        do
          p option i32 := 1
          while !(count p.get)
```
```
/home/sam/playground/test.fz:11:27: error 1: Incompatible types when passing argument in a call
          while !(count p.get)
--------------------------^^^
Actual type for argument #1 'n' does not match expected type.
In call to          : 'test.count'
expected formal type: 'void'
actual type found   : 'i32'
assignable to       : 'i32'
for value assigned  : 'deref(#loop0.this.#^test.#fun0.call.#fun1.call.#loop1.#loop0).p.get'
To solve this, you could change the type of the target 'n' to 'i32' or convert the type of the assigned value to 'void'.

one error.
```

michaellilltokiwa:
…d of i32

fixes #1918

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -1258,7 +1258,7 @@ public class Call extends AbstractCall
   private void resolveFormalArg(Resolution res, int argnum, AbstractFeature frml)
   {
     int cnt = 1;
-    var frmlT = frml.resultType();
+    var frmlT = frml.resultTypeRaw(res);
     if (CHECKS) check
       (frmlT == Types.intern(frmlT));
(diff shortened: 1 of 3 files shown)


fridis (approved):
Good. 
I think we also will need some clean-up in the naming and documentation of functions that produce a result type, it is not clear enough at what point which function should be used.

PR #2135, opened 2023-10-25
be/jvm: fix `thread.join0`

michaellilltokiwa:
previous to this change skipped test concur_simple did not work: 
```
RUN concur_simple.fz 0a1,26
> 
> error 1: java.lang.NullPointerException: Cannot invoke "java.lang.Thread.join()" because the return value of "dev.flang.be.interpreter.OpenResources.get(java.lang.Long)" is null
>       at dev.flang.be.jvm.runtime.Runtime.fuzion_sys_thread_join0(Runtime.java:717)
>       at fzC_concur__1thread__1spawned__join.fzRoutine(Unknown Source)
>       at fzC_concur_simple.fzRoutine(Unknown Source)
>       at fzC_universe.main(Unknown Source)
>       at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>       at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
>       at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>       at java.base/java.lang.reflect.Method.invoke(Method.java:568)
>       at dev.flang.be.jvm.Runner.runMain(Runner.java:120)
>       at dev.flang.be.jvm.JVM$CompilePhase$4.finish(JVM.java:493)
>       at dev.flang.be.jvm.JVM.lambda$createCode$0(JVM.java:799)
>       at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992)
>       at java.base/java.util.stream.ReferencePipeline$Head.forEachOrdered(ReferencePipeline.java:772)
>       at dev.flang.be.jvm.JVM.createCode(JVM.java:790)
>       at dev.flang.be.jvm.JVM.compile(JVM.java:775)
>       at dev.flang.tools.Fuzion$Backend$3.process(Fuzion.java:163)
>       at dev.flang.tools.Fuzion$Backend.processFrontEnd(Fuzion.java:444)
>       at dev.flang.tools.Fuzion.lambda$parseArgsForBackend$3(Fuzion.java:908)
>       at dev.flang.tools.Tool.lambda$run$0(Tool.java:145)
>       at dev.flang.util.Errors.runAndExit(Errors.java:790)
>       at dev.flang.tools.Tool.run(Tool.java:145)
>       at dev.flang.tools.Fuzion.main(Fuzion.java:557)
> 
> *** fatal errors encountered, stopping.

```

### DIFF
--- a/src/dev/flang/be/jvm/runtime/Intrinsics.java
+++ b/src/dev/flang/be/jvm/runtime/Intrinsics.java
@@ -811,6 +811,21 @@ public class Intrinsics extends ANY
     return System.getenv(Runtime.utf8ByteArrayDataToString((byte[])s)) != null;
   }
 
+  public static void fuzion_sys_thread_join0(long threadId)
+  {
+    try
+      {
+        Runtime._startedThreads_.get(threadId).join();
+        Runtime._startedThreads_.remove(threadId);
+      }
+    catch (InterruptedException e)
+      {
+        // NYI handle this exception
+        System.err.println("Joining of threads was interrupted: " + e);
+        System.exit(1);
+      }
+  };
+
 }
 
 /* end of file */
--- a/src/dev/flang/be/jvm/runtime/Intrinsics.java
+++ b/src/dev/flang/be/jvm/runtime/Intrinsics.java
@@ -813,18 +813,30 @@ public class Intrinsics extends ANY
 
   public static void fuzion_sys_thread_join0(long threadId)
   {
-    try
+    var thread = Runtime._startedThreads_.get(threadId);
+    var result = false;
+    do
       {
-        Runtime._startedThreads_.get(threadId).join();
-        Runtime._startedThreads_.remove(threadId);
+        try
+          {
+            thread.join();
+            result = true;
+          }
+        catch (InterruptedException e)
+          {
+
+          }
       }
-    catch (InterruptedException e)
-      {
-        // NYI handle this exception
-        System.err.println("Joining of threads was interrupted: " + e);
-        System.exit(1);
-      }
-  };
+    while (!result);
+
+    // NYI: comment, fridi:
+    // Furthermore, remove should probably not be called by join, but either by
+    // the Thread itself or by some cleanup mechanism that removes terminated
+    // threads, either when new threads are started or by a system thread that
+    // joins and removes threads that are about to terminate.
+    Runtime._startedThreads_.remove(threadId);
+  }
+
 
 }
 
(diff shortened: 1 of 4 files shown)


src/dev/flang/be/jvm/runtime/Intrinsics.java
```diff
@@ -811,6 +811,21 @@ public static boolean fuzion_sys_env_vars_has0(Object s)
...
+        Runtime._startedThreads_.get(threadId).join();
+        Runtime._startedThreads_.remove(threadId);
+      }
+    catch (InterruptedException e)
+      {
+        // NYI handle this exception
+        System.err.println("Joining of threads was interrupted: " + e);
+        System.exit(1);
+      }
+  };
```
fridis, line 827: Better put this into a loop that repeats `join` until it finished without exception. 
BTW, this code is not thread-safe: ` _startedThread_.get()`  could result in `null` causing a `NullPointerException`. Furthermore, ` remove` should probably not be called by `join`, but either by the Thread itself or by some cleanup mechanism that removes terminated threads, either when new threads are started or by a system thread that joins and removes threads that are about to terminate.

fridis (changes requested):
There are still a few issues with threads, no need to solve them all in this PR, but maybe create issues for those that I mentioned and that are not solved.

src/dev/flang/be/jvm/runtime/Intrinsics.java
michaellilltokiwa: @fridis ~~My understanding is that this does not need to be thread-safe since fuzion stdlib only lets you join a thread that the thread calling `join` spawned.~~ 

Seems like this is not the case (yet). But probably should be.

My question is:
Are there any use-cases where you want to join a thread from a different thread than the one that started it?

Also there is this comment on `join` in stdlib:
# NYI it should be made impossible to call this more than once

michaellilltokiwa: from `man pthread`:
> If multiple threads simultaneously try to join with the same thread, the results are undefined

fridis: I was not aware of this. In Java. `Thread.join()` can be called from any thread to any other thread and an arbitrary number of times. Should IMHO be the same in our base lib, so we eventually will need to have a more sophisticated handling here. For now, since we do not have any mechanism like Java's `wait`/`notify`, I should be enough to document that his needs more work.

michaellilltokiwa: I added issue #2161

fridis (approved):

PR #2148, opened 2023-10-26
ast/call: resolve syntactic sugar only for none inheritance calls

michaellilltokiwa:


src/dev/flang/ast/Call.java
```diff
@@ -2479,7 +2479,7 @@ Expr resolveSyntacticSugar(Resolution res, AbstractFeature outer)
   {
     Expr result = this;
     //    if (true) return result;
-    if (!Errors.any())
+    if (!Errors.any() && !isInheritanceCall())
```
fridis, line 2482: Please add a comment why this is not done for an inheritance call. 
I guess that for inherting from a boolean operator (`f : true.prefix !`) , you will get an error anyways since these are not constructors (better check:-), while for a numeric literal (`f : i32 42`) we do not want this replaced by a `NumLiteral`.

fridis (changes requested):
Looks reasonable, but IMHO needs some documentation.

fridis (approved):

PR #2149, opened 2023-10-26
enable jvm-backend

michaellilltokiwa:


tests/javaBase/skip_jvm
```diff
@@ -0,0 +1 @@
+NYI
```
fridis, line 1: Can you create an issue for this an reference it here.

tests/unicode/skip_jvm
```diff
@@ -0,0 +1 @@
+NYI results in code size of bytecode created for `character_encodings.unicode.data.lower_case_mappings` is 107131, the maximum allowed length is 65535
```
fridis, line 1: Please create an issue for this an reference it here.

fridis (changes requested):
Good, we can pull this once the sockets work again...
One request: I consider the JVM backend the most important one now, so I would prefer the order jvm, c, int.

fridis (approved):

PR #2150, opened 2023-10-26
be/jvm, fix "Attempt to modify immutable array" in test/sockets

michaellilltokiwa:


fridis (approved):
looks like my last-minute change to my patch destroyed it. Thanks for fixing this.

PR #2152, opened 2023-10-26
change visibility of `fuzion.sys` to module

### LINKED ISSUE #2097: `fuzion.sys.internal_array.index[]` and `set []` should not be public
Best would be if `fuzion.sys.internal_array` as well could be visible to the base module only.

michaellilltokiwa:
fixes #2097

### DIFF
--- a/src/dev/flang/fe/Module.java
+++ b/src/dev/flang/fe/Module.java
@@ -37,6 +37,7 @@ import dev.flang.mir.MIR;
 
 import dev.flang.util.ANY;
 import dev.flang.util.Errors;
+import dev.flang.util.FuzionConstants;
 import dev.flang.util.HasSourcePosition;
 import dev.flang.util.SourceFile;
 
@@ -132,6 +133,12 @@ public abstract class Module extends ANY
    * @param outer the declaring feature
    */
   public abstract SortedMap<FeatureName, AbstractFeature>declaredFeatures(AbstractFeature outer);
+
+
+  /**
+   * The name of this module, e.g. base
+   */
+  abstract String name();
 
 
   /**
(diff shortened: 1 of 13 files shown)


src/dev/flang/fe/Module.java
```diff
@@ -316,8 +318,9 @@ protected boolean featureVisible(SourceFile usedIn, AbstractFeature af)
     var definedIn = af.pos()._sourceFile;
     var v = af.visibility();
 
-    return  // built-in or generated features like #loop0
-            af.pos().isBuiltIn()
+    return  SourcePosition.builtIn._sourceFile == usedIn
```
fridis, line 321: This needs an explanation comment like the other cases. 
It would be better to not compare against `SourcePosition.builtIn._sourceFile`, but instead add a new static field `SourcePosition._builtInFile_`.
Then, I would still restrict this a bit more, e.g.,
```
 // generated inline code may access base library features
|| v.ordinal() >= Visi.MOD.ordinal() && m.name.equals("base") && usedIn == SourcePosition._builtInFile_
```
(and "base" should better be a `static final` field in `FrontEnd.java`).

src/dev/flang/fe/Module.java
```diff
@@ -298,7 +299,8 @@ protected boolean typeVisible(SourceFile usedIn, AbstractFeature af, boolean ign
     var v = af.visibility();
     var definesType = af.definesType() || ignoreDefinesType;
 
-    return definesType && (usedIn.sameAs(definedIn)
+    return SourcePosition.builtIn._sourceFile == usedIn
```
fridis, line 302: basically same coment as in line 321.

fridis (changes requested):
Nice. 
I would add some comments and make it a bit stricter to reduce the risk that generated code accidentally accesses private features or features that are local to a different module than the base module.

PR #2174, opened 2023-10-31
jvm: Enable support for option -unsafeIntrinsics=off

fridis:
Unlike the old interpreter backend where unsafe intrinsics are disabled by default, they are not enabled by default for the interpreter and the JVM backends, so care is needed when running code form an untrusted source.


michaellilltokiwa:
@fridis has conflicts now.

michaellilltokiwa (approved):

PR #2199, opened 2023-11-03
remove unused method `Clazz.depth`

michaellilltokiwa:


fridis (approved):
Nice to get rid of this old code that was used when handling of outer clazzes was still very unstable.

PR #2216, opened 2023-11-07
turn (some) inline arrays into compile-time constants

michaellilltokiwa:
- modifies `InlineArray` to generate code in syntax-sugar2-phase but not be replaced by, instead saving it in a field `_code`.
- extend `LibraryOut` with ability of writing `InlineArray`
- extend `LibraryFeature`/`LibraryModule` with ability of reading `InlineArray`
- be/interpreter, middle-end and IR use InlineArray.code()
- `FUIR.toStack` checks if `InlineArray` can be constant, if yes we turn the array into an `AbstractConstant` an put it on the stack otherwise `code` will be used for creating the `InlineArray`.
- dfa/be-c/be-jvm are extended to be able to handle `AbstractConstant` of array of none primitive types.


fridis (approved):
very nice, nothing to complain about!!

PR #2228, opened 2023-11-11
ast/tests: join findGenerics with resolve in UnresolvedType

fridis:
This is a major cleanup of the spaghetti style code that performs type resolution.  Methods `findGenerics` and `resolveFeature` are now both part for `UnresolvedType.resolve` and no longer needed in `AbstractType`.

Other methods like `AbstractType.resolveThisType`, `AbstractType.ensureNotOpen`, `ResolvedNomralType.resolve`, `ResolvedNormalType.resolveGenerics` and one varaint of `SourceModule.lookupType` could also be removed.

One important special handling remains after this patch: For type inference within match cases, a `clone()` of a partially resolved `ResolvedNormalType` is created that later gets its actual type parameters resolved.

Tests were changed as follows:

- Added `a` and `b` to the error case `49` in `tests/generics_negative` since that case appeared twice.

- Fixed undesired compilation errors in `tests/reg_issue309` since `Set` was moved to `container.Set`.

- Updated expected error output for `tests/visibility_modules_negative` sice the order of visibility checks for types has changed.

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -443,7 +443,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
                       {
                         if (res != null)
                           {
-                            af.visit(res.findGenerics);
+                            af.visit(res.resolveTypesOnly);
                           }
                         t = af.returnType().functionReturnType();
                       }
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -464,7 +464,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
   /**
    * Get instance of Generic that corresponds to this type parameter.
    *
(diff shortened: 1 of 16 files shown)


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -464,7 +464,7 @@ boolean hasOpenGenericsArgList(Resolution res)
   /**
    * Get instance of Generic that corresponds to this type parameter.
    *
-   * NYI: Since there is a 1-to-1 correspondance between type parameter features
+   * NYI: Since there is a 1-to-1 correspondent between type parameter features
```
michaellilltokiwa, line 467: should be correspondence, imho?

michaellilltokiwa (approved):

src/dev/flang/ast/AbstractFeature.java
fridis: leo.org does not agree, we might need a native speaker to clarify this.

PR #2243, opened 2023-11-16
air: replace Clazz.actualClazz by Clazz.handDown(), fix #2214

### LINKED ISSUE #2214: Type lookup confusion between outer and inherited type
This small example
```
  u : Unary unit i32 is
  _ := (a i32).m u

  a(T type) is
    m(B type, f T->B) : a B is
```
results in a check failure or in broken code when using the JVM, or —less severely so— the C backend:
```
 > PRECONDITIONS=true POSTCONDITIONS=true ./build/bin/fz primes6_broken_map_small.fz

error 1: java.lang.Error: check-condition failed: dev.flang.air.Clazzes:propagateExpectedClazz:736
	at dev.flang.util.ANY.check(ANY.java:383)
	at dev.flang.air.Clazzes.propagateExpectedClazz(Clazzes.java:736)
	at dev.flang.air.Clazzes.findClazzes(Clazzes.java:832)
	at dev.flang.air.Clazz$1.action(Clazz.java:1548)
...
```
The reason is that `T` in the type signature of `f` gets replaced by the actual type passed to `B`, i.e., `unit` and not `i32`.
This is the reason why `examples/primes6_broken_map` currently fails in Jenkins.

fridis:
The problem with `Clazz.actualClazz` was that it replaces all generics found in parents and in outer features.  This is a problem when a feature inherits from one of its outer features since there are two possible values of the actual generics and for the examplein #2214, the wrong one was used.

With this patch, `Clazzes.down` now hands down an inherted type along the inheritance change in each nesting level of features, so we have much better control how inherited types are replaced by the actual values provided in the inherts-call.

This patch also adds a test case with the code from #2214 plus the flang.dev example `primes6_broken_map.fz` that showed this problem since a long time.

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -1211,50 +1211,50 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
     return a;
   }
 
-  public static AbstractType handDownStatic(List<AbstractCall> inh, AbstractType a, AbstractFeature heir)
-  {
-    return handDownStatic(null, inh, new AbstractType[] { a }, heir)[0];
-  }
   public static List<AbstractType> handDownStatic(List<AbstractCall> inh, List<AbstractType> a, AbstractFeature heir)
   {
-    var ar1 = a.toArray(new AbstractType[a.size()]);
-    var ar2 = handDownStatic(null, inh, ar1, heir);
-    var al = new List<AbstractType>(ar2);
-    return al;
+    for (AbstractCall c : inh)
+      {
+        var f = c.calledFeature();
+        var actualTypes = c.actualTypeParameters();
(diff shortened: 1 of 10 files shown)


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -1205,33 +1205,51 @@ public AbstractType[] handDown(Resolution res, AbstractType[] a, AbstractFeature
...
-            for (AbstractCall c : heir.findInheritanceChain(outer()))
+            a = AbstractFeature.handDownInheritance(res, inh, a, heir);
+          }
+      }
+    return a;
+  }
+
+
+  /**
+   * Helper for handDown() to hadn down an array of types along a given inheritance chain.
```
michaellilltokiwa, line 1216: hadn -> hand

michaellilltokiwa (approved):

Issue #1835, opened 2023-08-10
be/c: unhandled dynamic target in `idiom82ex.fz`


maxteufel:
Running `idioms/src/idiom82ex.fz` throws this error:
```
*** testbin.c:11628: unhandled dynamic target 1513216160 in access of (ref Cons u8 (list u8)).head within (container.searchable_sequence u8).starts_with
```

It should be noted that `idiom82ex.fz` seems to be prone to producing this kind of error, and there is possibly some deeper kind of underlying issue that has not been fixed, because there have been two similar "fixed" issues already: #116, #177. Perhaps this is also related to #1750.

michaellilltokiwa:
example is probably not minimal yet but it is a start:
```
ex is
  ss(A type, from Sequence A)
  is

    starts_with bool is
      match from.as_list
        nil     => false
        c1 Cons => say c1.head; true

    tails list (list A) is
      ref : Cons (list A) (list (list A))
        head => from.as_list
        tail => (from.as_list ? nil    => nil
                              | c Cons => c.tail.tails)

  say ((ss "banana".utf8)
    .tails
    .filter (t -> (ss t).starts_with))
```

michaellilltokiwa:
an example exhibiting the problem that does not rely on `base.fum` (much)
```
ex is
  public list2(A type) : choice nil (Cons A (list2 A)) is

    tails2 list2 (list2 A) is
        ref : Cons (list2 A) (list2 (list2 A))
          head => list2.this
          tail => (list2.this ? nil    => nil
                              | c Cons => c.tail.tails2
                              )

    redef as_string =>
      match list2.this
        nil => ""
        c Cons => c.head.as_string + " " + c.tail.as_string


  e list2 i32 :=
    ref : Cons i32 (list2 i32)
      head  => 0
      tail =>
        x list2 i32 := nil
        x
  x list2 i32 :=
    ref : Cons i32 (list2 i32)
      head  => 0
      tail => e

  say x.tails2
```
interpreter:
```
0 0  0 
```
c-backend:
```
*** ex.c:19216: unhandled dynamic target 3455 in access of (ref Cons i32 (ex.list2 i32)).head within (ex.list2 i32).as_string
```

michaellilltokiwa:
slightly modified example:
```
/home/not_synced/fuzion (2023-08-18T12-28-57+02-00)$ cat ~/playground/test.fz 
public Cons2(B type) ref is
  public tail B is abstract

ex =>

[...]

PR #2260, opened 2023-11-22
be/c: don't deref heap cloned result of call

michaellilltokiwa:
fixes idiom82
fixes https://github.com/tokiwa-software/fuzion/issues/1835


fridis (approved):
great to finally understand and fix this!

PR #2296, opened 2023-12-04
fe: fix using `.this.type` inside of type feature

### LINKED ISSUE #2295: using `.this.type` in type feature leads to feature not found
```
ex is

  type.one => "1"

  type.two =>
    ex.this.type.one
```
this works:
```
ex is

  type.one => "1"

  type.two =>
    one
```

michaellilltokiwa:
fixes #2295


fridis (changes requested):
Nice. 
Just one change request not to silently ignore `t == null` if `propagateExpectedType`.

fridis (approved):

PR #2299, opened 2023-12-04
lib: Add `Sequence.combine`, `Sequence.pairs` and `list.combine_list`

fridis:
These seem useful for AdventOfCode-kind of problems, used `combine` on 4 Dec solution.

### DIFF
--- a/lib/list.fz
+++ b/lib/list.fz
@@ -457,6 +457,25 @@ public list(public A type) : choice nil (Cons A (list A)), Sequence A is
               | nil    => nil
 
 
+  # create a new list from the result of applying 'f' to the
+  # elements all combinations of elements of this list and
+  # all elements of 'b' iterating of 'b' repeatedly as follows
+  #
+  #   list.this[0]  , b[0]
+  #   list.this[0]  , b[1]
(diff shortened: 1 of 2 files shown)


lib/list.fz
```diff
@@ -457,6 +457,25 @@ public list(public A type) : choice nil (Cons A (list A)), Sequence A is
...
+  #   list.this[0]  , ...
+  #   list.this[0]  , b.last
+  #   list.this[1]  , b[0]
+  #   list.this[1]  , b[1]
+  #   list.this[1]  , ...
+  #   ...           , ...
+  #   list.this.last, b.last
+  #
+  public combine_list(U, V type, b Sequence U, f (A,U)->V) =>
+    flat_map V x->(b.map_sequence y->(f x y))
```
michaellilltokiwa, line 476: It is unclear to me why you use `map_sequence` here instead of `map`?

fridis: because `b` is of type `Sequence` and `Sequence.map` does not exists yet...
But maybe `b` should be of type `list U`, not sure.

michaellilltokiwa (approved):

PR #2317, opened 2023-12-06
lib: rename `map`, `flat_map`

### LINKED ISSUE #2286: Covariance for `Sequence.map`
There are too many features to perform a map of a `Sequence`: there is `Sequence.map_sequence` that produces a `Sequence` which is a `ref` type, while specific sequences define their own `map` like `array.map` and `list.map` that produce a value type. 
It would be good if the result type of `Sequence.map` would be `Sequence.this` such that it can be redefined by `array` and `list` to return the corresponding type.
Same holds for related features like `flat_map`.

michaellilltokiwa:
fixes #2286

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -84,7 +84,7 @@ public Sequence(public T type) ref is
   public count
   pre
     analysis: finite  # in practice, we do not always have this information
-  => (map_sequence i32 (_ -> 1)).fold i32.type.sum
+  => (map i32 (_ -> 1)).fold i32.type.sum
 
 
   # get the first element of this Sequence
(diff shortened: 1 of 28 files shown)


fridis (commented):
Very nice. A few minor changes that I would like to see:

- the result type of `Sequence.map` and `Sequence.flat_map` should be `Sequence T`, it durrently is `list T`. 
- it would be good if any children of `Sequence` that provide their own `map_to_<myself>` should also redefine `map` that calls `map_to_<myself>`. The reason is that the basic performance characteristics remain the same after a `map`.

fridis (approved):

PR #2337, opened 2023-12-08
lib: Redefine count and drop for array backed sequences

fridis:
The redefinition of `count` in `Mutable_Array` now has O(1) performance.

Similarly, redefine `Sequence.drop` in `array` and result of `array.slice`. This improves the performance of `drop n` from O(n) to O(1).

I did not add a redefinition of `drop` to `Mutable_Array`, I am not sure if `Mutable_Array` should be a `Sequence` at all.

### DIFF
--- a/lib/mutate.fz
+++ b/lib/mutate.fz
@@ -335,6 +335,6 @@ public Mutable_Array(public T type) ref : Sequence T is
   # count the number of elements in this Sequence.
   #
   # Since we know the exact length, this is redefined to achive O(1) performance instead
-  # of O(n) for couting the elements.
+  # of O(n) for counting the elements.
   #
   public redef count => length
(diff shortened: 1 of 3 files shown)


lib/mutate.fz
```diff
@@ -321,3 +324,17 @@ public Mutable_Array(public T type) ref : Sequence T is
...
+  # is this Sequence known to be array backed? If so, this means that operations
+  # like index[] are fast.
+  #
+  public redef is_array_backed => true
+
+
+  # count the number of elements in this Sequence.
+  #
+  # Since we know the exact length, this is redefined to achive O(1) performance instead
+  # of O(n) for couting the elements.
```
michaellilltokiwa, line 338: counting

fridis: fixed.

michaellilltokiwa (approved):

PR #2350, opened 2023-12-11
add utility `add_simple_test`

michaellilltokiwa:
Using `make add_simple_test` will create a new template test in tests.

### DIFF
--- a/lib/String.fz
+++ b/lib/String.fz
@@ -39,6 +39,9 @@ public String ref : property.equatable, property.hashable, property.orderable is
   # returns true if string is empty or contains whitespace only
   public is_blank => utf8 ∀ u -> u.is_ascii_white_space
 
+  # returns true if string contains whitespace
+  public contains_whitespace => utf8 ∃ u -> u.is_ascii_white_space
+
   # length of this string in bytes
   public byte_length => utf8.count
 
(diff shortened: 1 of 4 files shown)


lib/String.fz
```diff
@@ -39,6 +39,9 @@ public String ref : property.equatable, property.hashable, property.orderable is
   # returns true if string is empty or contains whitespace only
   public is_blank => utf8 ∀ u -> u.is_ascii_white_space
 
+  # returns true if string contains whitespace
+  public contains_whitespace => utf8 ∃ u -> u.is_ascii_white_space
+
```
fridis, line 44: I think this is unrelated to the PR

lib/io/file.fz
```diff
@@ -26,3 +26,10 @@
 # file -- a unit type that contains effects related to file I/O
 #
 public file is
+
+
+  # Does a file or directory at path exist?
+  #
+  public exists(path String) =>
+    (stat path false).ok
```
fridis, line 34: I think this is unrelated to the PR

fridis (changes requested):
Lookg good, just think there are some unrelated commits that ended up in this PR.

lib/String.fz
michaellilltokiwa: @fridis It is related. used in bin/add_simple_test.fz

lib/io/file.fz
michaellilltokiwa: @fridis also related

fridis (approved):

PR #2352, opened 2023-12-11
lib: cleanup `effect.fz`

michaellilltokiwa:
removed unused `return` and `use0`
also removed `effect_mode.abort` which seems useless?


fridis (approved):
I think that `return` is the common term used in literature for what we call `abort`. So maybe we should rename `abort` as `return` in a next step.

PR #2359, opened 2023-12-12
fix choice with none disjoint generics via type parameter

### LINKED ISSUE #1591: `option nil` does not work properly yet
Inspired by PR #1589, I wrote this test
```
test_option =>

  nil_option(T type) option T is nil
  val_option(T type, v T) option T is v

  a(T type, x option T) =>
    match x
      v T => "val: $v"
      nil => "nil!"

  b(T type, x option T) =>
    match x
      nil => "nil!"
      v T => "val: $v"

  chck (a (nil_option i32)) "nil!"
  chck (a (val_option 132)) "val: 132"
  chck (a (nil_option nil)) "nil!"
  chck (a (val_option nil)) "val: nil"

  chck (b (nil_option i32)) "nil!"
  chck (b (val_option 132)) "val: 132"
  chck (b (nil_option nil)) "nil!"
  chck (b (val_option nil)) "val: nil"

  chck(a, b String) =>
    if a = b
      say "PASSED: $a = $b"
    else
      say "** FAILED **: $a /= $b"
```
which produces three failures when interpreted
```
 > ./build/bin/fz  test_option.fz 
PASSED: nil! = nil!
PASSED: val: 132 = val: 132
** FAILED **: val: --nil-- /= nil!
** FAILED **: val: --nil-- /= val: nil
PASSED: nil! = nil!
PASSED: val: 132 = val: 132
PASSED: nil! = nil!
** FAILED **: nil! /= val: nil
```
and two failures with the C backend
```
 > ./build/bin/fz -c test_option.fz 
 > ./test_option
PASSED: nil! = nil!
PASSED: val: 132 = val: 132
PASSED: nil! = nil!
** FAILED **: nil! /= val: nil
PASSED: nil! = nil!
PASSED: val: 132 = val: 132
PASSED: nil! = nil!
** FAILED **: nil! /= val: nil
```

michaellilltokiwa:
fixes #1591

### DIFF
--- a/src/dev/flang/air/Clazz.java
+++ b/src/dev/flang/air/Clazz.java
@@ -1743,7 +1743,7 @@ public class Clazz extends ANY implements Comparable<Clazz>
         if (g._type.isDirectlyAssignableFrom(staticTypeOfValue))
           {
             if (CHECKS) check
-              (result < 0);
+              (Errors.any() || result < 0);
             result = index;
           }
         index++;
@@ -1995,6 +1995,10 @@ public class Clazz extends ANY implements Comparable<Clazz>
     if (isInstantiated() && _abstractCalled != null)
       {
         AirErrors.abstractFeatureNotImplemented(feature(), _abstractCalled, _instantiationPos);
+      }
+    if (isInstantiated() && isChoice())
+      {
+        this._type.checkChoiceDisjoint(_instantiationPos.pos());
       }
   }
 
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -942,7 +942,21 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
           {
             AstErrors.refToChoice(pos);
           }
+        checkChoiceDisjoint(pos);
+      }
+  }
 
+
+  /**
+   * Check that all choice generics are disjoint.
+   *
+   * @param pos for error messages.
+   */
+  public void checkChoiceDisjoint(SourcePosition pos)
+  {
+    var g = choiceGenerics();
+    if (g != null)
+      {
         int i1 = 0;
         for (var t1 : g)
           {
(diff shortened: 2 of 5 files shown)


fridis (commented):
I think this is not what I intended with #1591. IMHO `option T` should work, even if `T` is `nil`. Otherwise library code with a type parameter `T` could not use type `option T` without requiring `T` never to be `nil`. 
The problem should be fixed in the backends that essentially have to support a choice where the variants are effectively the same type.

michaellilltokiwa:
@fridis hmm, I'm not really convinced.

Considering this example:
```
  nil_option(T type) option T is nil

  a(T type, x option T) =>
    match x
      v T => "val: $v"
      nil => "nil!"

  chck (a (nil_option nil)) "nil!"
```

How would the backends know which tag to choose for `nil_option nil`. It could be the one for `T:nil` or for `nil` which could lead to `a (nil_option nil))` yielding "nil!" or "val: nil".

fridis:
In this example, if you called
```
  p := nil_option unit
```
then `p` would have a tag that is `0` for `unit` and `1` for `nil`, so `p.tag=1` The same for 
```
  q := nil_option nil
```
where `q` would have a tag that is `0` for `T=nil` and `1` for the empty option, so `q.tag=1`. Then, the match in `a` should run into the second case.
What is not possible in this case is to do `match y` since then both cases would be `nil` and this should result in a frontend error. But the `match` in `a` should work for any `T`. 
You can even create more bizarre cases like
```
  one_of_three(A, B, C type, a A, b B, c D, which i32) A | B | C =>
    if      which = 1 then a
    else if which = 2 then b
    else c
  show(A, B, C type, x A | B | C) =>
    match x
      A => say "A"
      B => say "B"
      C => say "C"
```
and then do 
```
  for i in 0..2 do
    v := one_of_three nil nil nil i
    show v
```
which should print `A B C`.

PR #2363, opened 2023-12-13
fix inference when wrapping numliteral in lazy

### LINKED ISSUE #2358: type inference involving null-ary and numliteral
```
ex is

  something (a Lazy u32) is
  _ := something 0
```
yields
```
/home/sam/playground/test.fz:4:18: error 2: Incompatible types in assignment
  _ := something 0
-----------------^
assignment to field : 'ex.#fun0.call.result'
expected formal type: 'u32'
actual type found   : 'i32'
assignable to       : 'i32'
for value assigned  : '0'
To solve this, you could convert the value using + '.as_u32'.
```

michaellilltokiwa:
fixes #2358

### DIFF
--- a/src/dev/flang/ast/NumLiteral.java
+++ b/src/dev/flang/ast/NumLiteral.java
@@ -830,6 +830,29 @@ public class NumLiteral extends Constant
 
 
   /**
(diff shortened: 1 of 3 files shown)


fridis (approved):
Good.

PR #2364, opened 2023-12-13
makefile add missing `FZ_JVM`

michaellilltokiwa:


fridis (changes requested):
Good. Can you add `$(CLASS_FILES_BE_JVM_RUNTIME)` to the list for `FZ_JVM` since these are needed to run code with the JVM backend.

michaellilltokiwa:
@fridis Good point. Added as suggested.

fridis (approved):

PR #2373, opened 2023-12-15
be/c: `-fno-omit-frame-pointer`

michaellilltokiwa:
Enable frame pointers as fedora and ubuntu decided to do. More information on why this might be a good idea: https://fedoraproject.org/wiki/Changes/fno-omit-frame-pointer and:
https://lobste.rs/s/avrfxz/ubuntu_24_04_lts_will_enable_frame


fridis (approved):
Hurts my soul to disable an optimization, but we are at a stage where the performance impact of this change is very likely outweighed by the debugging and profiling advantage.

PR #2375, opened 2023-12-15
lib: add comment for `thrush` operator, just for completeness

fridis:


lib/composition.fz
```diff
@@ -52,6 +52,15 @@ public composition is
   # Haskell id
   public id(T type) T -> T is x->x
 
+  # ?
+  # reverse apply
+  # Bird: thrush
+  # BQN: ?
+  # Haskell: $
+  #
+  # this is already defined in univers:
```
michaellilltokiwa, line 61: universe

fridis: fixed.

michaellilltokiwa (approved):

PR #2392, opened 2023-12-19
Automatic select `t.0`/`t.1`,... instead of `t.values.0`, `t.values.1`, ...

fridis:
This simplifies the code to access the elements of, e.g., a tuple where it is not possible to write `t.2` instead of `t.values.2`.

This syntax sugar comes with the following restrictions:

- it is only applicable to types declared in library features

- the result type of the feature must be an open type parameter

- there must not be any function arguments in the call

- there must be exactly one such feature visible

For this, I also had to fix type inference from actual args in case actuals get replaced.

This fixes a bug that was revealed by automatically calling features with open type parameters (eg., t.0 instead of t.values.0 in tuples): When the actual expression gets replaced by something else, the type inference still used the original expression resulting in potentially wrong types being inferred.

This patch no longer stores the actual argument that was found, but instead stores the call that was found.  To determine the actual type, the actual is taken form the call resulting in the new actual in case it was replaced during resolveTypes or syntax sugar phases.

This also cleans up the somewhat clumsy check to prevent infinite recursion and uses a simple flag instead.


michaellilltokiwa:
Cool. I also tried implementing `t.0` as shorthand for `t.values.0` a few weeks ago but looking at this patch my efforts where destined to fail... :-)

michaellilltokiwa (approved):

fridis:
> Cool. I also tried implementing `t.0` as shorthand for `t.values.0` a few weeks ago but looking at this patch my efforts where destined to fail... :-)

Well, the change in `Call.java` was not the problem, the problem was fixing the existing bug in `Impl.java`...

PR #2398, opened 2023-12-21
be/c: Java intrinsics

michaellilltokiwa:
Initial implementation of java intrinsics for the c backend.

### DIFF
--- a/src/dev/flang/be/c/C.java
+++ b/src/dev/flang/be/c/C.java
@@ -1985,8 +1985,8 @@ public class C extends ANY
   {
     var constString = constString(CExpr.call("fzE_java_string_to_utf8_bytes", new List<>(field)), CExpr.call("strlen", new List<>(CExpr.call("fzE_java_string_to_utf8_bytes", new List<>(field)))));
     return CExpr.compoundLiteral(
-      "fzT_1error",
-      ".fzF_0_msg = " +
+      _names.struct(_fuir.clazz_error()),
+      "." + _names.fieldName(_fuir.clazzArg(_fuir.clazz_error(), 0)).code() + " = " +
         CExpr
           .call(CNames.HEAP_CLONE._name, new List<>(constString.adrOf(), constString.sizeOfExpr()))
           .code()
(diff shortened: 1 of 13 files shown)


src/dev/flang/be/c/C.java
```diff
@@ -557,6 +555,14 @@ public CStmnt contract(int cl, FUIR.ContractKind ck, CExpr cc)
   static private final boolean FUZION_DEBUG_TAIL_CALL = "true".equals(System.getenv("FUZION_DEBUG_TAIL_CALL"));
 
 
+  /*
+   * If you want the c-backend to link the JVM,
+   * set this environment variable to e.g.:
+   * FUZION_JVM_PATH=/usr/lib/jvm/java-21-openjdk-amd64
+   */
+  static final String FUZION_JVM_PATH = System.getenv("FUZION_JVM_PATH");
```
michaellilltokiwa, line 563: @fridis @maxteufel Is there another/better way to achieve this than via environment variable?

michaellilltokiwa: @maxteufel @fridis Yes, I think so too. But if we used `JAVA_HOME` then the jvm would be linked for every fuzion application if `JAVA_HOME` is set. This way linking only happens if this _special_ env variable is set. 

Best solution would probably be if we had a mechanism of determining which intrinsics are used and link only if fuzion.java* is actually used but AFAIK this is currently not possible?

src/dev/flang/be/c/C.java
```diff
@@ -927,6 +951,11 @@ private void createCode(CFile cf, COptions _options) throws IOException
         cf.println("GC_INIT(); /* Optional on Linux/X86 */");
       }
 
+    if (FUZION_JVM_PATH != null)
+      {
+        cf.println("fzE_init_jvm();");
```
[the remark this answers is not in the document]
michaellilltokiwa: @maxteufel I think currently it does not or do you know how to do it?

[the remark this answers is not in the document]
michaellilltokiwa: @maxteufel  I implemented your suggestion.

src/dev/flang/be/c/C.java
michaellilltokiwa: changed this

michaellilltokiwa:
updated in #2604

PR #2415, opened 2024-01-04
tests/reg_issue2237: fix NAME not being overridden

maxteufel:


fridis (changes requested):
Can you do the same change to reg_issue37_inhGenericOuter?

michaellilltokiwa (approved):

PR #2427, opened 2024-01-09
lib: `Java_Ref Any`-> `Java_Ref J_Value`

michaellilltokiwa:
this should make it impossible to call Java_Object etc. accidentally, since the pre-condition of J_Value is always false.


michaellilltokiwa:
> I would personally like to block this until I'm finished with the fzjava integration into the JVM backend. You partial merges from the development branch have already caused unexpected behavior.

That's fine by me.

PR #2428, opened 2024-01-09
be/c: enable garbage collector by default

michaellilltokiwa:
Tests are still all green on my machine.
MacOS/Windows actions will likely be broken after merging this PR even though gc-packages are already included to be installed.
Include-path and library-path probably need to be set correctly and the documentation of the libgc packages is rather limited unfortunately.


michaellilltokiwa:
> have you also tested this with the flang.dev examples, @michaellilltokiwa?

Yes, that is what i meant with "Tests are still all green".

PR #2430, opened 2024-01-09
fuir: permit regular expressions to debug AbstractInterpreter

fridis:
The debug output enabled via
```
  FUZION_JAVA_OPTIONS=-Ddev.flang.fuir.analysis.AbstractInterpreter.DEBUG=true
```
results in large amounts of output that could be hard to handle.

With this patch, one can now use a regex to print only abstract interpretation of those features whose name match a given regular expression.


fridis:
> There is one case where this won't work, when you want the expression to be `false`. But that can possibly be ignored.

It also won't work for `true`. But both cases can be solved,e.g., by `(false)` or `(true)`.

michaellilltokiwa (approved):

PR #2435, opened 2024-01-10
fix: "no targets for access"

### LINKED ISSUE #2231: no targets for access
```
ex  is
  a := mut ()->i32 ()->0
  eff : simple_effect is
    inner := 7
    a <- ()->inner
    something is
      abort

  eff.go ()->
    ex.eff.env.something
  say a.get.call
```
```

error 1: no targets for access of ((Effect_Call (option ex.eff.something)).call).#^Effect_Call.call within (Effect_Call (option ex.eff.something)).call

```

### LINKED ISSUE #2273: 100doors result in `no targets for access of...
My attempt to implement Rosetta code's [100 doors](https://rosettacode.org/wiki/100_doors)
```
open, closed.
door : choice open closed is
  toggle door => match door.this
    open   => closed
    closed => open
  redef as_string =>  match door.this
    open   => "open"
    closed => "closed"

d := (mutate.array door).type.new mutate 100 closed
for i in 0..99 do
  for j in i..99 do
    d[j] := d[j].toggle
say d
```
results in this
```
 > ./build/bin/fz doors.fz 

error 1: no targets for access of (((mutate.#type mutate).array.#type (mutate.array door) door).new mutate).length within ((mutate.#type mutate).array.#type (mutate.array door) door).new mutate

```

### LINKED ISSUE #2280: AoC issue: no target for `((Effect_Call (outcome void)).call).#^Effect_Call.call`
This code
```
dec1 is

  # 0 to 9
  codepoint_type_ascii_digit => u32 0x30 .. 0x39
  # NYI: this should be public in codepoint, had to copy it over!

  _ := io.stdin.with ()->
    for

      t := u32 0, s
      # NYI: would be nice if `u32` would be inferred and not needed explicitly here

      l := io.buffered.read_line
      # NYI: read_line should return a choice of String | end_of_file or similar

      l2 := l.as_codepoints
             .as_list       # NYI: would be good not to need this
             .map (.val)    # NYI: would be good not to need this
             .filter codepoint_type_ascii_digit.contains    # using brand new partial application!

      first := l2.head.get (u32 0)
      #      last  :=  l2.is_empty ? 0 : l2.last        # NYI: does not work since ? : is not lazy on its arguments!

      last  :=  {if l2.is_empty then u32 0 else l2.last}
      # NYI: why this assymmetry between head returning option and last returning the element or panicking?
      # NYI: would be nice if `u32` would be inferred and not needed explicitly here!
      # NYI: can we get rid of the `{`/`}`

      s := t + first + last
    do
    # while l != "" do
      say s
```
produces
```
 > ../../../work/build/bin/fz dec1_strange_error.fz 

error 1: no targets for access of ((Effect_Call (outcome void)).call).#^Effect_Call.call within (Effect_Call (outcome void)).call

```

michaellilltokiwa:
fixes https://github.com/tokiwa-software/fuzion/issues/2231
fixes https://github.com/tokiwa-software/fuzion/issues/2273
fixes https://github.com/tokiwa-software/fuzion/issues/2280

Essentially the problem was/is that the DFA finds that the lambda passed to `go` will never return. This leads to a push of `null`/`void` in the AbstractInterpreter, stopping any further processing of this code block.
But the backends do not know that this calls never returns and continue processing the code block. For the C-backend this works by accident, I think.
It did however break the JVM-Backend.

To reiterate the situation is that the JVM-backend generates code the DFA did not analyze. This is problemetic since
`accessedClazzes` did not contain a call to outer ref `#^Effect_Call.call` which led to the bugs described in the issues.
```
ex  is
  eff : simple_effect is
    something is
      abort

  _ := eff.go ()->
    ex.eff.env.something
```

### DIFF
--- a/src/dev/flang/fuir/analysis/dfa/Call.java
+++ b/src/dev/flang/fuir/analysis/dfa/Call.java
@@ -258,7 +258,7 @@ public class Call extends ANY implements Comparable<Call>, Context
                                                 yield null; }
           };
       }
-    else if (_returns)
+    else if (_returns || !_dfa._fuir.clazzIsVoidType(_dfa._fuir.clazzResultClazz(_cc)))
       {
         var rf = _dfa._fuir.clazzResultField(_cc);
         if (_pre)
@@ -273,7 +273,7 @@ public class Call extends ANY implements Comparable<Call>, Context
           {
             result = Value.UNIT;
           }
-        else
+        else if(_returns)
           {
             // should not be possible to return void (_result should be null):
             if (CHECKS) check
--- a/tests/reg_issue2231/Makefile
+++ b/tests/reg_issue2231/Makefile
@@ -1,0 +1,25 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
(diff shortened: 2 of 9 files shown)


michaellilltokiwa:
meanwhile fixed differently

PR #2454, opened 2024-01-12
Ast remove types intern

michaellilltokiwa:
depends on #2447


src/dev/flang/air/Clazzes.java
```diff
@@ -292,9 +292,14 @@ public static Clazz create(AbstractType actualType, int select, Clazz outer)
             outer = o;  // short-circuit outer relation if suitable outer was found
           }
 
-        // NYI this should be, but tests/unary fails currently:
-        // o._type.compareTo(actualType) == 0 && ...
-        if (o._type == actualType && actualType != Types.t_ERROR &&
+        if (o._type.compareTo(actualType) == 0 &&
+            // example where the following logic is relevant:
```
michaellilltokiwa, line 296: @tokiwa-software/developers Do you agree with this assessment?

fridis (commented):
Happy to see the code getting simpler, but would like to know about performance before pulling this in:
In the early days of Fuzion, `Types.intern` was very important since even small examples resulted in large numbers of instances of `Type`. I hope this problem is solved now, but I am not sure. 
Did you see an performance changes with this?

michaellilltokiwa:
> Happy to see the code getting simpler, but would like to know about performance before pulling this in: In the early days of Fuzion, `Types.intern` was very important since even small examples resulted in large numbers of instances of `Type`. I hope this problem is solved now, but I am not sure. Did you see an performance changes with this?

I did not notice any significant performance changes. Also the test run seems to run in a similar time as before.

fridis (approved):

PR #2481, opened 2024-01-19
JVM backend runs javaHello.fz

fridis:
I have added the following changes to Max' work

- fixed order of `res`/`Expr.UNIT` in `Pair<>` results, which caused the original classfile verifier errors. This is very tricky: The result that produces a value on the stack must be the first in the pair, the second value is for any code that may produce side effects. Currently, the code mixes side-effects and producing the result value on the stack which may mix up the execution order of side effects. 
- fixed accesses to `fuzion.Java_Object.javaRef`, these must use the actual type of the feature that inherits from `fuzion.Java_Object`
- added some DFA handling for intrinsics, but this needs to be checked for completeness, in particular all fields that are read by any intrinsics must be marked so during DFA.
- Added code to wrap result in `outcome` for `call_v0`, this is still missing for `call_s0` and `call_c0`.
- Added `JavaError`. A call into Java code that returns with a checked exception results in the single instance of `JavaError` being returned. This needs to be checked by `call_v0/s0/c0`, e.g. using `instanceof`. If an error occured, it is saved in `((FuzionThread)Thread.currentThread())._thrownException`, this must then be wrapped into an `outcome`
- I think I broke the interpreter, needs to be checked. Maybe was broken before.
- removed `skip_jvm` from `tests/javaBase`, this should work now.
- probably some more things I forgot now.


fridis:
replaced by #2487.

PR #2485, opened 2024-01-23
fix VerifyError, returning result field that does not exist

### LINKED ISSUE #2477: `VerifyError` on feature that does not return
With this code 
```
a Any => do
_ := a
```
I get
```
 > ./build/bin/fz tj.fz

error 1: java.lang.VerifyError: (class: fzC_a, method: fzRoutine signature: ()LfzI_Any;) Unable to pop operand off an empty stack
	at fzC_universe.fzRoutine(Unknown Source)
	at fzC_universe.fz_run(Unknown Source)
	at dev.flang.be.jvm.runtime.FuzionThread.lambda$new$1(FuzionThread.java:90)
	at dev.flang.util.Errors.runAndExit(Errors.java:845)
	at dev.flang.be.jvm.runtime.FuzionThread.lambda$new$2(FuzionThread.java:99)
	at java.base/java.lang.Thread.run(Thread.java:833)


```

michaellilltokiwa:
fixes #2477

I did not add a regression test, since it would not terminate.

### DIFF
--- a/src/dev/flang/be/jvm/JVM.java
+++ b/src/dev/flang/be/jvm/JVM.java
@@ -1003,8 +1003,7 @@ should be avoided as much as possible.
           traceReturn(cl, pre)
           .andThen(getf)
           .andThen(
-            fieldExists(r)
-              ? ft.return0()
+            fieldExists(r) ||
               /**
                * Example where fieldExists is false but we still need a return:
                *
@@ -1015,8 +1014,8 @@ should be avoided as much as possible.
                *     .read
                * _ := test0 unit_like unit
                */
-              : _fuir.clazzIsChoice(t)
-              ? Expr.RETURN
+            ft == PrimitiveType.type_void
+              ? ft.return0()
               /*
                * For special cases like:
                *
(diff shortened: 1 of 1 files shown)


src/dev/flang/be/jvm/JVM.java
```diff
@@ -1002,7 +1002,30 @@ else if (r == -1)   // a constructor
...
+               *
+               * unit_like : choice unit is
+               * test0(T type, a T) =>
+               *   concur
+               *     .atomic a
+               *     .read
+               * _ := test0 unit_like unit
+               */
+              : _fuir.clazzIsChoice(t)
+              ? Expr.RETURN
```
fridis, line 1019: Good to have the comment with the example. 
In this case, I think the condition should better be
```
  ft == PrimitiveType.type_void
  ? ft.return0()
```
such that this also covers non-choice unit types. 

Then the code is the same as in the first case, so we could have
```
          .andThen(
            fieldExists(r) ||
              /**
               * Example where fieldExists is false but we still need a return:
               *
               * unit_like : choice unit is
               * test0(T type, a T) =>
               *   concur
               *     .atomic a
               *     .read
               * _ := test0 unit_like unit
               */
            ft == PrimitiveType.type_void
              ? Expr.RETURN
```

fridis (changes requested):

fridis (approved):

PR #2487, opened 2024-01-25
JVM backend runs `example/javaGraphics/gfx.fz` and `example/webserver/webserver.fz`

maxteufel:
Please do add comments on what can be improved, cleaned up, and so on.


src/dev/flang/be/interpreter/JavaInterface.java
```diff
@@ -243,6 +243,54 @@ static Object instanceToJavaObject(Instance i)
   }
 
 
+  /**
+   * Extract Java object from a fuzion.sys.Pointer stored by a JavaRef
+   *
+   * @param i a JavaRef, must be a fuzion.sys.Pointer
```
michaellilltokiwa, line 249: `@param r`

src/dev/flang/be/jvm/Intrinsix.java
```diff
@@ -245,129 +245,277 @@ private static Expr locked(Expr e)
...
+            .andThen(call_jlStringArg(jvm, cc, args, 1))
+            .andThen(call_jlStringArg(jvm, cc, args, 2))
+            .andThen(args.get(3))
+            .andThen(jvm.getfield(data)) // args
+            .andThen(Expr.checkcast(JAVA_LANG_OBJECT.array()))
+            .andThen(Expr.invokeStatic(Names.RUNTIME_CLASS,
+                                       "fuzion_java_call_s0",
+                                       "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;[Ljava/lang/Object;)Ljava/lang/Object;",
+                                       Names.JAVA_LANG_OBJECT))
+            .andThen(Expr.checkcast(JAVA_LANG_OBJECT));
```
michaellilltokiwa, line 499: what is the purpose of these checkcasts?

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -101,6 +107,13 @@ private SystemErrNo(final int errno)
   public static final String CLASS_PREFIX      = "fzC_";
 
 
+  /**
+   * Result value used when returning from a call into Java code in case an
+   * exception was thrown by that code.
+   */
+  public static final JavaError _JAVA_ERROR_ = new JavaError();;
```
michaellilltokiwa, line 114: ;;

michaellilltokiwa (approved):
Cool

src/dev/flang/be/jvm/Intrinsix.java
```diff
@@ -607,16 +752,7 @@ else if (in.equals("fuzion.sys.internal_array.setel"))
...
-        "fuzion.java.array_length",
-        "fuzion.java.array_to_java_object0",
-        "fuzion.java.call_c0",
-        "fuzion.java.call_s0",
-        "fuzion.java.call_v0",
-        "fuzion.java.get_field0",
-        "fuzion.java.get_static_field0",
-        "fuzion.java.java_string_to_string",
-        "fuzion.java.string_to_java_object0"
+      {
```
fridis, line 755: please add a comment here that the names of not-implemented intrinsics should go here.

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -711,6 +724,190 @@ public static byte[] fuzion_sys_env_vars_get0(Object d)
...
+      }
+    catch (NoSuchMethodException e)
+      {
+        Errors.fatal("NoSuchMethodException when calling fuzion.java.call_constructor calling " +
+                           ("new " + clName) + sig);
+        return null; // not reached
+      }
+  }
+
+
```
fridis, line 910: would be good to have comments for these methods.

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -711,6 +724,190 @@ public static byte[] fuzion_sys_env_vars_get0(Object d)
...
+  public static Object fuzion_java_call_s0(String clName, String name, String sig, Object[] args)
+  {
+    if (PRECONDITIONS) require
+      (clName != null);
+
+    Method m = null;
+    Constructor co = null;
```
fridis, line 846: unused var.

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -711,6 +724,190 @@ public static byte[] fuzion_sys_env_vars_get0(Object d)
...
+    try
+      {
+        cl = Class.forName(clName);
+      }
+    catch (ClassNotFoundException e)
+      {
+        Errors.fatal("ClassNotFoundException when calling fuzion.java.call_static for class " +
+                           clName + " calling " + name + sig);
+        cl = Object.class; // not reached.
+      }
```
fridis, line 862: Except for the error message, this code is a copy of the same code in `fuzion_java_call_v0`, can you extract it into a helper method?

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -711,6 +724,190 @@ public static byte[] fuzion_sys_env_vars_get0(Object d)
...
+    try
+      {
+        cl = Class.forName(clName);
+      }
+    catch (ClassNotFoundException e)
+      {
+        Errors.fatal("ClassNotFoundException when calling fuzion.java.call_constructor for class " +
+                           clName + " calling " + ("new " + clName) + sig);
+        cl = Object.class; // not reached.
+      }
```
fridis, line 896: Except for the error message, this code is a copy of the same code in `fuzion_java_call_v0` and `fuzion_java_call_s0`, can you extract it into a helper method?

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -1849,6 +1864,101 @@ static void setArrayElementsToAnything(Call cl, int argnum, String intrinsicName
...
+              var v = switch (cl._dfa._fuir.getSpecialId(oc))
+                {
+                  case c_i8, c_u16, c_i16, c_i32, c_i64,
+                    c_f32, c_f64 -> new NumericValue(cl._dfa, oc);
+                  case c_bool -> cl._dfa._bool;
+                  case c_unit -> Value.UNIT;
+                  default -> Value.UNIT;
+                };
+              var okay = new TaggedValue(cl._dfa, oc, v, 0);
+              var err = new TaggedValue(cl._dfa, oc, cl._dfa.newInstance(cl._dfa._fuir.clazzChoice(rc, 1), null), 1);
```
fridis, line 1886: `oc` here should probably be `rc` (2x).

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -1849,6 +1864,101 @@ static void setArrayElementsToAnything(Call cl, int argnum, String intrinsicName
...
+              yield okay.join(err);
+            }
+            default -> {
+              var jref = cl._dfa._fuir.lookupJavaRef(oc);
+              var jobj = cl._dfa.newInstance(oc, null);
+              jobj.setField(cl._dfa, jref, Value.UNKNOWN_JAVA_REF);
+              var okay = new TaggedValue(cl._dfa, rc, jobj, 0);
+              var err = new TaggedValue(cl._dfa, rc, cl._dfa.newInstance(cl._dfa._fuir.clazzChoice(rc, 1), null), 1);
+              yield okay.join(err);
```
fridis, line 1895: These three lines are repeated compared to the previous case (except for the likely bug that `oc` is used there, and the different variable name `v` vs. `jobj`). Can you try to move these out of the switch and avoid the duplication?

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -1849,6 +1864,101 @@ static void setArrayElementsToAnything(Call cl, int argnum, String intrinsicName
         jobj.setField(cl._dfa, jref, Value.UNKNOWN_JAVA_REF);
         return jobj;
       });
+  }
+  static Value newFuzionJavaCall(Call cl) {
+    var rc = cl._dfa._fuir.clazzResultClazz(cl._cc);
+    if (cl._dfa._fuir.clazzBaseName(rc).startsWith("outcome"))
```
fridis, line 1870: I think this comparison was introduced by me in the Interpreter a long time ago. This is ugly and error-prone, it must at least be marked using `// NYI: HACK: should properly check if rc is an outcome` or similar, best would be to implement a proper check, i.e., FUIR should know `outcome` and have methods to check if a class is based on that feature.

fridis (changes requested):
Nice work that brings Fuzion a giant step ahead!

I have a few nit-picking requests form comments and code improvements and I think I found one possible bug, please have a look , fix the bug in case I am right and improve the code in case my comments are reasonable and easily feasible.

Makefile
```diff
@@ -532,7 +532,7 @@ $(CLASS_FILES_BE_JVM_CLASSFILE): $(JAVA_FILES_BE_JVM_CLASSFILE) $(CLASS_FILES_UT
 	$(JAVAC) -cp $(CLASSES_DIR) -d $(CLASSES_DIR) $(JAVA_FILES_BE_JVM_CLASSFILE)
 	touch $@
 
-$(CLASS_FILES_BE_JVM_RUNTIME): $(JAVA_FILES_BE_JVM_RUNTIME) $(CLASS_FILES_UTIL)
+$(CLASS_FILES_BE_JVM_RUNTIME): $(JAVA_FILES_BE_JVM_RUNTIME) $(CLASS_FILES_UTIL) $(CLASS_FILES_BE_INTERPRETER)
```
fridis, line 535: This might be done in a separate clean-up PR: The `dev.flang.be.jvm.runtime` package should contain only the minimum code required to be packaged with a jar created by the JVM backend. So the dependency on the interpreter basically requires all the interpreter to be part of this, which is a bit too much. It would be much better to move the essential code to `dev.flang.be.jvm.runtime` and introduce a dependence from the interpreter backend on the JVM backend's runtime. 
Related to this: I am already a little unhappy about the dependency on `dev.flang.util`, but would not clean this up yet since we will likely need things like `SourceFile` and `SourcePosition` for better error reporting and debugging of code created by the JVM backend.

fridis (approved):

PR #2499, opened 2024-02-01
add support for stackmap table attribute.

### LINKED ISSUE #2179: JVM backend should use Java 17 or Java 21 JVM spec, not Java 5
The main thing blocking a more recent version is the lack of strack frame entries that should be created.

michaellilltokiwa:
fixes #2179

### DIFF
--- a/src/dev/flang/be/jvm/JVM.java
+++ b/src/dev/flang/be/jvm/JVM.java
@@ -905,7 +905,7 @@ should be avoided as much as possible.
     if (!_types.isScalar(cl))  // not calls like `u8 0x20` or `f32 3.14`.
       {
         var cf = _types.classFile(cl);
-        var vti = _types.resultType(cl).vti(cf);
+        var vti = _types.resultType(cl).vti();
         result = result.andThen(new0(cl))
           .andThen(cl == _fuir.clazzUniverse()
                    ? Expr.DUP.andThen(Expr.putstatic(_names.javaClass(cl),
@@ -984,7 +984,7 @@ should be avoided as much as possible.
         var jt = _types.javaType(t);
         return
           traceReturn(cl, pre)
-          .andThen(jt.load(current_index(cl), cf))
+          .andThen(jt.load(current_index(cl)))
           .andThen(jt.return0());
       }
     else
@@ -999,7 +999,7 @@ should be avoided as much as possible.
 
         var ft = _types.resultType(t);
         var getf =
-          fieldExists(r) ? (Expr.aload(current_index(cl), ft, _types.javaType(cl).vti(cf) )
+          fieldExists(r) ? (Expr.aload(current_index(cl), ft, _types.javaType(cl).vti())
                             .andThen(getfield(r)))
                          : Expr.UNIT;
         return
@@ -1195,13 +1195,13 @@ should be avoided as much as possible.
             do
               {
                 bc_combined = bc_combined
-                  .andThen(javaTypeOfTarget(cl).load(0, cf));
+                  .andThen(javaTypeOfTarget(cl).load(0));
                 for(var i = 0; i<_fuir.clazzArgCount(cl); i++)
                   {
                     var at = _fuir.clazzArgClazz(cl, i);
                     var jti = _types.resultType(at);
                     bc_combined = bc_combined
-                      .andThen(jti.load(argSlot(cl, i), cf));
+                      .andThen(jti.load(argSlot(cl, i)));
                   }
                 bc_combined = bc_combined
                   .andThen(Expr.invokeStatic(_names.javaClass(cl),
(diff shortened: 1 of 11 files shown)


src/dev/flang/be/jvm/classfile/Expr.java
```diff
@@ -285,7 +586,7 @@ public static Expr commentAlways(String msg)
    * create invokestatic bytecode to call given class, name and descr producing
    * given result type on the stack.
    */
-  public static Expr invokeStatic(String cls, String name, String descr, JavaType rt)
+  public static Expr invokeStatic(String cls, String name, String descr, JavaType rt, int argCount)
```
fridis, line 589: Instead of adding `argCount` as an argument here, try `ClassFileConstants.slotCountForArgs(descr)`.

src/dev/flang/be/jvm/classfile/Expr.java
```diff
@@ -309,8 +619,11 @@ public void code(ClassFile.ByteCodeWriter ba, ClassFile cf)
    * create invokespecial bytecode to call given class, name and descr producing
    * void result type.
    */
-  public static Expr invokeSpecial(String cls, String name, String descr)
+  public static Expr invokeSpecial(String cls, String name, String descr, int argCount)
```
fridis, line 622: Instead of adding `argCount` as an argument here, try `ClassFileConstants.slotCountForArgs(descr)`.

src/dev/flang/be/jvm/classfile/Expr.java
```diff
@@ -355,7 +685,7 @@ public void code(ClassFile.ByteCodeWriter ba, ClassFile cf)
    * create invokevirtual bytecode to call given class, name and descr producing
    * given result type on the stack.
    */
-  public static Expr invokeVirtual(String cls, String name, String descr, JavaType rt)
+  public static Expr invokeVirtual(String cls, String name, String descr, JavaType rt, int argCount)
```
fridis, line 688: Instead of adding `argCount` as an argument here, try `ClassFileConstants.slotCountForArgs(descr)`.

src/dev/flang/be/jvm/classfile/Expr.java
```diff
@@ -388,7 +728,7 @@ public void code(ClassFile.ByteCodeWriter ba, ClassFile cf)
    *
    * @return Code to produce bytecode for the interface call.
    */
-  public static Expr invokeInterface(String cls, String name, String descr, JavaType rt)
+  public static Expr invokeInterface(String cls, String name, String descr, JavaType rt, int argCount)
```
fridis, line 731: Instead of adding `argCount` as an argument here, try `ClassFileConstants.slotCountForArgs(descr)`.

src/dev/flang/be/jvm/classfile/ClassFileConstants.java
```diff
@@ -475,7 +513,7 @@ public String refDescriptor()
       throw new Error("JavaType.refDescriptor only defined for ref type, not for " + this);
     }
 
-    public Expr load(int index)
+    public Expr load(int index, ClassFile cf)
```
fridis, line 516: The `cf` argument is added here to create cpool entries used by the `VerificationType`.  This early creation of cpool entries does not match my intended design for Expr where cpool entries are created late during calls to `code`. This has the advantage that code that turns out to be unused or duplicated or reused in a different class file still works and does not produce any unused cpool entries. 
It would be great if the VerificationType could delay the creation of the cpool entries such that they are not created by a call to `vti(cf)` but during `VerifictionType.write`. This could be done by passing a lambda to `new VerificationType` or by adding arguments with extra data for all cases needed.

src/dev/flang/be/jvm/JVM.java
```diff
@@ -994,10 +997,9 @@ else if (r == -1)   // a constructor
 
         */
 
-        var jt = _types.javaType(t);
         var ft = _types.resultType(t);
         var getf =
-          fieldExists(r) ? (Expr.aload(current_index(cl), jt)
+          fieldExists(r) ? (Expr.aload(current_index(cl), ft, new VerificationType(VerificationType.type.Object, 2 /* NYI magic constant... */) )
```
fridis, line 1002: no magic please. Might be solved by creating cpool entries later, see other comment.

src/dev/flang/be/jvm/classfile/Expr.java
```diff
@@ -92,6 +174,200 @@ public void code(ClassFile.ByteCodeWriter ba, ClassFile cf)
     {
       ba.write(_bc);
     }
+
+    @Override
+    public void buildStackMapTable(StackMapTable smt, Stack<VerificationType> stack, List<VerificationType> locals)
+    {
+      if (_bc == BC_RETURN ||
```
fridis, line 181: I think it would be better to do a `switch (_bc[0]) { case O_return, O_ireturn, ... -> ` here.

fridis (commented):
Very nice step ahead, I was very annoyed by the class file version being stuck at Java 1.5...

I have a few suggestions to improve the code, but I am not sure if it is better to implement these in a separate patch or in this patch, please decide what is best for you.

fridis (approved):
Very nice!

PR #2516, opened 2024-02-07
lib: add `infix ::`

fridis:
This is convenient to create lists where the next element can be expressed as a function of the previous element, e.g., to produce all integers, we can use

  0 :: +1

with `0` as head and `+1` as a partial function to create the next element, so the resulting list will be 0,1,2,3,4,5,...


michaellilltokiwa (approved):
Very useful :-)

PR #2523, opened 2024-02-08
fix typo: ones -> once

michaellilltokiwa:


fridis (changes requested):
I think the original name is correct, this is an infinite list of only `1`, it may well be evaluated more than once. 

Might maybe be better to use something like
```
 stars => say "eval stars!"; lst "*" stars
```
to avoid this confusion.

PR #2525, opened 2024-02-08
fix `array_to_java_object` for java primitives

michaellilltokiwa:
I added tests for `array_to_java_object` for fuzion all features that map to java primitives: i8, i16, i32, i64, u8, u16, u32, u64, f32, f64, bool

Turns out this was broken in interpreter and also the DFA.


fridis (changes requested):
Maybe extend the test with cases for `u8`, `u32`, `u64`.

fridis (approved):

PR #2526, opened 2024-02-08
be/c: introduce constants for typeClazzID (uint32_t), typeTagName (in…

michaellilltokiwa:
…t32_t)

### DIFF
--- a/src/dev/flang/be/c/CTypes.java
+++ b/src/dev/flang/be/c/CTypes.java
@@ -49,13 +49,13 @@ public class CTypes extends ANY
   /*
    * type of clazzId field in ref instances
    */
-  private static final String typeClazzID = scalar(SpecialClazzes.c_u32);
+  private static final String typeClazzID = "uint32_t";
 
 
   /*
    * type of the tag field in choices
    */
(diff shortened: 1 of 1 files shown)


src/dev/flang/be/c/CTypes.java
```diff
@@ -43,6 +43,21 @@ public class CTypes extends ANY
...
+  /*
+   * type of clazzId field in ref instances
+   */
+  private static final String typeClazzID = scalar(SpecialClazzes.c_u32);
+
+
+  /*
+   * type of the tag field in choices
+   */
+  private static final String typeTagName = scalar(SpecialClazzes.c_i32);
```
fridis, line 58: These should IMHO be just `"uint32_t"` or `"int32_t"` since the type is a free choice of the backend, there is no relation to features `i32` or `u32`.

fridis (changes requested):
It is good to have these constants. However, I would not create any reference to Fuzion clazzes here since the types are pure implementation choices by the backend,  e.g., a 8-bit integer for a tag would be fine as well (but leave it at 32 bits, please:-).

fridis (approved):

PR #2565, opened 2024-02-15
Jvm fix tailrec optimization

fridis:
For tests like `fib32.fz`, tail recursion optimization was not performed by the JVM backend resulting in jenkins test failures due to different stack trace outputs. 

The main problems were 
- that primitive types stored in local fields prevented tail call optimization since an operation like `a + 3` on a local field `a` of type `u32` does not create a reference to `a`, and
- the tail recursive call itself does not keep the current instance alife


fridis:
This was fixed with PR #2697.

PR #2574, opened 2024-02-19
be/c: emit compile script with *.c+binary-file

michaellilltokiwa:

### DIFF
--- a/src/dev/flang/be/c/C.java
+++ b/src/dev/flang/be/c/C.java
@@ -30,7 +30,10 @@ import java.io.IOException;
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


michaellilltokiwa:
@maxteufel Hmm, I see you point. Maybe we should emit the binary plus a directory containing all artifacts?

fridis:
I would first like to understand what the use case is for this script. 
One use case I can think of is for us during debugging of the C backend to be able to recompile the C code easily. But for this, it is IMHO sufficient to output the C compiler command line when `-verbose=n` is set high enough.

michaellilltokiwa:
My idea was for in the future to add scripts to be able to compile the c-source with e.g. emscripten to wasm, cross-compile for windows or offer a script to do compilation with gcc. But I agree with @maxteufel that these scripts should not be emitted by default. Furthermore the *.c-file should probably also be deleted unless user sets an option like `-XkeepGeneratedCode=on` or similar. @fridis @maxteufel Do you share this opinion?

PR #2582, opened 2024-02-20
fe: improve type inference for mixed numeric types

### LINKED ISSUE #1069: type inference for numeric types as type parameters could be improved
Type inferencing for array elements works quite well, e.g., this example results in `array (option f64)`:
```
 > cat test_infer_array.fz
a := [0.3, option 3.142, nil, 42.0]
type_of(T type, a T) ref Type is T
for v in a do say "$v {type_of v}"
 > ./build/bin/fz test_infer_array.fz
0.3 Type of 'option f64'
3.142 Type of 'option f64'
--nil-- Type of 'option f64'
42.0 Type of 'option f64'
```
but this slightly changed example no longer works since numeric literals default types are used:
```
 > cat test_infer_array1.fz
a := [0.3, option 3.142, nil, 42]
type_of(T type, a T) ref Type is T
for v in a do say "$v {type_of v}"
 > ./build/bin/fz test_infer_array1.fz

/home/fridi/fuzion/work/test_infer_array1.fz:1:6: error 1: Incompatible types in array elements
a := [0.3, option 3.142, nil, 42]
-----^
block returns value of type 'f64' at /home/fridi/fuzion/work/test_infer_array1.fz:1:7:
a := [0.3, option 3.142, nil, 42]
------^
block returns value of type 'option<f64>' at /home/fridi/fuzion/work/test_infer_array1.fz:1:12:
a := [0.3, option 3.142, nil, 42]
-----------^
block returns value of type 'nil' at /home/fridi/fuzion/work/test_infer_array1.fz:1:26:
a := [0.3, option 3.142, nil, 42]
-------------------------^
block returns value of type 'i32' at /home/fridi/fuzion/work/test_infer_array1.fz:1:31:
a := [0.3, option 3.142, nil, 42]
------------------------------^

one error.

```
same issue with conditionals: 
```
 > cat test_infer_array2.fz
f(b bool) =>
  if b
    42
  else
    option 3.14

say (f true)
say (f false)
 > ./build/bin/fz test_infer_array2.fz

/home/fridi/fuzion/work/test_infer_array2.fz:2:3: error 1: Incompatible types in branches of if statement
  if b
--^
block returns value of type 'i32' at /home/fridi/fuzion/work/test_infer_array2.fz:3:5:
    42
----^
block returns value of type 'option<f64>' at /home/fridi/fuzion/work/test_infer_array2.fz:5:5:
    option 3.14
----^

one error.
```

### LINKED ISSUE #2282: Type inference from numeric literal joined with explicit numeric type should not default to `i32`/`f64`
This example currently does not compile since `4` defaults to type `i32`:
```
 > echo "a := u32 3; b := if true then 4 else a" | $FUZION/bin/fz -

-:1:18: error 1: Incompatible types in branches of if expression
a := u32 3; b := if true then 4 else a
-----------------^
block returns value of type 'i32' at -:1:31:
a := u32 3; b := if true then 4 else a
------------------------------^
block returns value of type 'u32' at -:1:38:
a := u32 3; b := if true then 4 else a
-------------------------------------^

one error.
```
The same holds for loop index variables as in 
```
  for a := 0, b
      b := read_u32
  do
```
that currently requires an explicit type for `0`:
```
  for a := u32 0, b
      b := read_u32
  do
```

michaellilltokiwa:
fixes #1069
fixes #2282 

This patch enhances type inference for numbers if they are unified with results from other branches like in arrays, if-else or match-cases.

Examples:
- `a := [0.3, option 3.142, nil, 42]`   # infers that all the numbers to be of type `option f64`
- `if b then 42  else option 3.14`     # infers that all the numbers to be of type `option f64`

### DIFF
--- a/src/dev/flang/ast/AbstractMatch.java
+++ b/src/dev/flang/ast/AbstractMatch.java
@@ -128,7 +128,7 @@ public abstract class AbstractMatch extends Expr
    */
   private AbstractType typeFromCases()
   {
-    var result = Expr.union(cases().map2(x -> x.code()), false);
+    var result = Expr.union(cases().map2(x -> x.code()));
     if (result == Types.t_ERROR)
       {
         new IncompatibleResultsOnBranches(pos(),
--- a/src/dev/flang/ast/AbstractMatch.java
+++ b/src/dev/flang/ast/AbstractMatch.java
@@ -128,12 +128,7 @@ public abstract class AbstractMatch extends Expr
    */
   private AbstractType typeFromCases()
   {
-    AbstractType result = Types.resolved.t_void;
-    for (var c: cases())
-      {
-        var t = c.code().typeForInferencing();
-        result = result == null || t == null ? null : result.union(t);
-      }
+    var result = Expr.union(cases().map2(x -> x.code()), false);
     if (result == Types.t_ERROR)
       {
         new IncompatibleResultsOnBranches(pos(),
--- a/src/dev/flang/ast/Expr.java
+++ b/src/dev/flang/ast/Expr.java
@@ -203,6 +203,80 @@ public abstract class Expr extends HasGlobalIndex implements HasSourcePosition
 
 
   /**
+   * typeForUnion returns the type of this expression or null if the type is
+   * still unknown, i.e., before or during type resolution.  This is redefined
+   * by sub-classes of Expr to provide type information.
+   *
+   * @return this Expr's type or null if not known.
+   */
+  AbstractType typeForUnion()
+  {
+    return typeForInferencing();
+  }
+
+
+  /**
+   * Get the result type of a union of all results of exprs.
+   *
+   * @param exprs the expression to unionize
+   *
+   * @param ignoreUnknow ignore any expressions whose type is not yet known.
+   *
+   * @return the union of exprs result type or null if not yet known.
+   */
+  public static AbstractType union(List<Expr> exprs, boolean ignoreUnknow)
+  {
+    AbstractType t = Types.resolved.t_void;
+
+    // First pass:
+    // Union of the types of the expressions
+    // that are sure about their types.
+    for (var e : exprs)
+      {
+        var et = e.typeForUnion();
+        if (et != null)
+          {
+            t = t.union(et);
+          }
+      }
+
+    // Propagate the found type to all expression
+    for (var e : exprs)
+      {
+        e.propagateExpectedType(t);
+      }
+
+    // Second pass:
+    // Union of the types of the expressions
+    AbstractType result = Types.resolved.t_void;
+    for (var e : exprs)
+      {
+        var et = e.typeForInferencing();
+        if (et == null && ignoreUnknow)
+          {
+            continue;
+          }
+        result = result == null || et == null
+          ? null : result.union(et);
+      }
+
+    return result;
+  }
+
+
+  /**
+   * During type inference: Inform this expression that it is
+   * expected to result in the given type.
+   *
+   * @param t the expected type.
+   */
+  protected void propagateExpectedType(AbstractType t)
+  {
+
+  }
+
+
+  /**
    * The source code position of this expression that produces the result value
    * of this Expression. This is usually equal to this Expression's position,
    * unless we have a block of the form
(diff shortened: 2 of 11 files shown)


src/dev/flang/ast/Expr.java
```diff
@@ -202,6 +202,80 @@ AbstractType typeForInferencing()
...
+    AbstractType result = Types.resolved.t_void;
+    for (var e : exprs)
+      {
+        var et = e.typeForInferencing();
+        if (et == null && ignoreUnknow)
+          {
+            continue;
+          }
+        result = result == null || et == null
+          ? null : result.union(et);
```
fridis, line 260: Avoid `break`, `continue` and `return` (unless at the end of a method), in most cases code is clearer without these. Here, you can use
```suggestion
        if (et != null || !ignoreUnknow)
          {
            result = result == null || et == null
              ? null : result.union(et);
          }
```

src/dev/flang/ast/InlineArray.java
```diff
@@ -124,14 +124,7 @@ AbstractType typeForInferencing()
...
       {
-        AbstractType t = Types.resolved.t_void;
-        for (var e : _elements)
-          {
-            var et = e.typeForInferencing();
-            t =
-              t  == null ? null :
-              et == null ? null : t.union(et);
-          }
+        var t = Expr.union(_elements, false);
```
fridis, line 127: Can you add a comment explaining when `ignoreUnknown` does not work here? I could imagine an array as follows
```
  f i32->i32 := x->x+1
  a := [f,f,x->x+2]
```
Where type inference for `a` could work fine if `itnoreUnknown`  was `true`.

src/dev/flang/ast/NumLiteral.java
```diff
@@ -423,6 +426,63 @@ AbstractType typeForInferencing()
...
+    if (PRECONDITIONS) require
+      (_type == null || extractNumericType(t) == null || _type.compareTo(extractNumericType(t)) == 0);
+
+    _type = extractNumericType(t);
+  }
+
+
+  private AbstractType extractNumericType(AbstractType t)
+  {
+    var numericTypes = new TreeSet<AbstractType>(
```
fridis, line 461: This should better be a field in `Types.resolved` to avoid re-creation of this `TreeSet` for every call to `propagateExpectedType`.

src/dev/flang/ast/NumLiteral.java
```diff
@@ -732,13 +792,13 @@ String toString(BigInteger v)
 
   /**
    * Check if type t is one of the known integer types i8, i16, i32, i64, u8,
```
fridis, line 794: `integer` -> `numeric`

tests/reg_issue1069/reg_issue1069.fz
```diff
@@ -0,0 +1,36 @@
...
+  for v in a do say "$v {type_of v}"
+
+  f(b bool) =>
+    if b
+      42
+    else
+      option 3.14
+
+  say (f true)
+  say (f false)
```
fridis, line 36: Would be good to also do `say (type_of (f true))` and  `say (type_of (f false))` (even though this is the same)

fridis (changes requested):
Very nice!

A general comment: in the PR detail message, would be good to have a bit more text than a reference to an issue. Not much, just  a small example or similar. 

Just some requests for minor improvements.

src/dev/flang/ast/InlineArray.java
michaellilltokiwa: I changed ignoreUnkown to `true` but example does not work yet due to other limitations.

src/dev/flang/ast/AbstractMatch.java
```diff
@@ -128,12 +128,7 @@ public void visitExpressions(ExpressionVisitor v)
    */
   private AbstractType typeFromCases()
   {
-    AbstractType result = Types.resolved.t_void;
-    for (var c: cases())
-      {
-        var t = c.code().typeForInferencing();
-        result = result == null || t == null ? null : result.union(t);
-      }
+    var result = Expr.union(cases().map2(x -> x.code()), false);
```
fridis, line 131: Sorry, missed this in the original review: Why is the `ignoreUnkown` here set to `false`? For `if` this is `true` and I do not see the fundamental difference between `match` and `if`. 
If it could be `true` as well we could get rid of this argument altogether!

fridis (changes requested):
Thanks for the update!

Sorry, one more point I overlooked in the first review.

fridis (approved):
very nice!

PR #2598, opened 2024-02-22
fe: inference, lambda return type containing tp to infer

### LINKED ISSUE #1355: type inference: the "flat_map issue"
flatMapSequence has args `B type, f T -> Sequence B`. It should be possible to infer B to be `codepoint` for the following example:
```
ex =>
  a := [1,2,3]
    .flatMapSequence (x -> [codepoint 35])
```
or if we provide a type for the target - `a` in this case - B should be inferred to be of type `String`:
```
ex =>
  a Sequence String := [1,2,3]
    .flatMapSequence (x -> [unit])
```

michaellilltokiwa:
fixes #1355

This patch adds type inference for e.g. 
`flat_map(B type, f T -> Sequence B) Sequence B =>...`
where the type parameter to infer is part of the lambdas result type.

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -28,6 +28,7 @@ package dev.flang.ast;
 
 import java.util.Arrays;
 import java.util.ListIterator;
+import java.util.Optional;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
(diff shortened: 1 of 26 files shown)


src/dev/flang/ast/Call.java
```diff
@@ -2476,23 +2477,37 @@ private boolean inferGenericLambdaResult(Resolution res,
...
+                    addPair(foundAt, ri, pos, rt);
+                    result = true;
                   }
-                addPair(foundAt, ri, pos, rt);
+              }
+          }
+        else if (at.containsUndefined(false) && frmLmdRt.dependsOnGenerics())
+          {
+            var rt = al.propagateTypeAndInferResult(res, outer, at, false);
+            // NYI What if there is more than one type parameter to infer?
```
fridis, line 2505: please add NYI-category.

I think you need to restructure the code of `extractType` and `indexOfGenericArgument` such that you only have on method that combines these two, i.e., remove `indexOfGenericArgument` and instead change `extractType` to return a list of tuples consisting for indices and corresponding actual type parameters.

tests/reg_issue1352/issue_1352.fz.expected_err
```diff
@@ -19,4 +19,13 @@ with argument count of the lambda expression equal to the number of type paramet
 assigned field must be given explicitly.
 To solve this, declare an explicit type for the target field, e.g., 'f (i32, i32) -> bool := x, y -> x > y'.
 
-3 errors.
+
+--CURDIR--/issue_1352.fz:27:10: error 4: No type information can be inferred from a lambda expression
+  d => b ->
+---------^
```
fridis, line 25: Would be great if this would appear instead of `error 2` reported above. But this could be fixed separatedly in a new PR.

src/dev/flang/ast/Call.java
```diff
@@ -2476,23 +2477,37 @@ private boolean inferGenericLambdaResult(Resolution res,
...
+            var et = rt.extractType(frmLmdRt);
+            if (et != null)
+              {
+                _generics = _generics.setOrClone(frmLmdRt.indexOfGenericArgument(), et);
+                addPair(foundAt, 0, pos, rt);
```
fridis, line 2510: It think, in this call to `addPair(foundAt, 0, pos, rt)` the argument `0` should be the index of the generic argument and `rt` should be `et`.

PR #2602, opened 2024-02-23
java intrinsics, change string_to_java_object0 to pass 0-terminated str

michaellilltokiwa:
This is necessary for c backend, otherwise we can not know the length of the string.

### DIFF
--- a/lib/fuzion/java.fz
+++ b/lib/fuzion/java.fz
@@ -207,7 +207,7 @@ public fuzion.java is
   # convert a string to a Java String object
   #
   public string_to_java_object(s String) Java_Object =>
-    string_to_java_object0 s.utf8.as_array.internal_array.data
+    string_to_java_object0 (sys.c_string s)
 
 
   # intrinsic to convert a string to a Java String object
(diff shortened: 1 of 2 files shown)


lib/fuzion/java.fz
```diff
@@ -207,7 +207,7 @@ public fuzion.java is
   # convert a string to a Java String object
   #
   public string_to_java_object(s String) Java_Object =>
-    string_to_java_object0 s.utf8.as_array.internal_array.data
+    string_to_java_object0 (sys.c_string s)
```
fridis, line 210: can you instead pass the byte-length to `string_to_java_object0`?

src/dev/flang/be/jvm/runtime/Runtime.java
```diff
@@ -745,15 +742,15 @@ public static byte[] fuzion_sys_env_vars_get0(Object d)
...
+   * @param b byte array consisting of a string encoded as UTF-8 bytes (0-terminated)
    *
    * @return the string from the array, as an instance of Java's String
    */
   public static String fuzion_java_string_to_java_object0(byte[] b)
   {
     unsafeIntrinsic();
 
-    return new String(b, StandardCharsets.UTF_8);
+    return new String(b, 0, b.length -1, StandardCharsets.UTF_8);
```
fridis, line 753: I would prefer not to add the overhead of the extra byte array creation in the JVM backend. If it is too hard to avoid this here, at least we shoudl add a `NYI: OPTIMZATION` comment.

fridis (changes requested):

PR #2604, opened 2024-02-23
be/c: java intrinsics

michaellilltokiwa:
initial implementation of java intrinsics for the c backend.


michaellilltokiwa:
#2602 has to merged first

fridis (approved):
Very nice. 

Would be good to have issues —best with failing examples—  for these open points: 

- `JNIEnv*` must be per-thread

- A remaining problem with this is interaction with the JVM's GC. If a reference to a Java object is obtained via JNI (e.g., via `GetObjectField`), it is stored as a local reference for the current thread (current `JNIEnv*`), so it will be reclaimed when this `JNIEnv*` is detached.  
The reference should better be registered globally using `JNIEnv->NewGlobalRef` and freed as soon as the corresponding Fuzion instance becomes inaccessible (e.g., by making these instances local to some Java-scope effect such that they all become unusable once the effect is left or by adding some kind for finalizer or phantom reference feature to Fuzion).

PR #2609, opened 2024-02-23
lib/tests: Add features to `pad`, `pad_left` and `pad_right` to String

fridis:
These are counterparts to `trim` that make it easier to create tabular output as in
```
  for i in -10..10 do
    say "{($i).pad_left 4}² = {($(i**2)).pad_left 4}."
```
Also added a simple test for these features.

### DIFF
--- a/lib/String.fz
+++ b/lib/String.fz
@@ -574,6 +574,87 @@ public String ref : property.equatable, property.hashable, property.orderable is
     String.type.from_bytes (utf8.as_list.reverse.drop_while c->c.is_ascii_white_space).reverse
 
 
+  # pad this string at the end with spaces such that its `codepoint_length` is at least `n`.
+  #
+  public pad(n i32)
+  pre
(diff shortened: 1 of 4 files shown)


lib/String.fz
```diff
@@ -574,6 +574,87 @@ public String ref : property.equatable, property.hashable, property.orderable is
...
+  #
+  public pad(p String, n i32)
+  pre
+    safety: p.codepoint_length > 0
+    debug: n >= 0
+  post
+    debug: result.codepoint_length >= n
+    debug: ((result.codepoint_length >= n+p.codepoint_length): result = String.this)
+  =>
+    pl := p.codepoint_length
```
michaellilltokiwa, line 623: Cool.
I think readability might improve a bit if expression `(max 0 (n-codepoint_length+pl-1)/pl)` was pulled out and given a proper name.

PR #2615, opened 2024-02-26
fe: raise error if effect is stored in field

michaellilltokiwa:
Effects instances must always be obtained from env. With this patch an error is raised if it is attempted to store an effect in field.

### DIFF
--- a/lib/panic.fz
+++ b/lib/panic.fz
@@ -87,5 +87,5 @@ public Panic_Handler ref is
   # removed and the result of do_panic should be returned instead).
   #
   public use(R type, code, def ()->R) =>
-    p := panic Panic_Handler.this unit
-    p.run code def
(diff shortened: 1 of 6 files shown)


fridis (commented):
This is not my intention, it should IMHO be possible to do anything with an effect instance once it was obtained from `env`, or even if it was created by calling its constructor. 
Instead, any operation defined in that effect would check that it is actually installed  when calling `replace`. This is currently not implemented efficiently but done manually, e.g., in `mutate.check_and_replace`.
Limiting the assignment to fields is generally not sufficient since references to an effect could also be created by passing arguments to features or even via calling an operation on the effect if the outer reference somehow escapes that call.  
I have created an issue #2626 for this and suggest to dump this PR for now.

PR #2625, opened 2024-02-28
lib: add file universe, describing universe and how it may be used

michaellilltokiwa:


michaellilltokiwa:
@maxteufel Yes, good idea.

fridis:
Since we define a number of features in the `universe`, e..g., `type_of` or `infix =`, I think it is good to have this file, such that we have a default place to put such features if there is no other good place. Not suggesting to move any of these, though.

Issue #2559, opened 2024-02-14
feature declarations in lazy expressions should be possible


fridis:
A lazy expression such as `b` in `a : b` may currently not contain any feature declarations, not even the declaration of a match-field as in this example: 
```
 > echo "say (0 : (match true { a TRUE => { nil }; * => { nil } }))" | ./build/bin/fz -

-:1:11: error 1: IMPLEMENTATION RESTRICTION: An expression used as a lazy value cannot contain feature declarations
say (0 : (match true { a TRUE => { nil }; * => { nil } }))
----------^
Declared features:
declared 'a' at -:1:24:
say (0 : (match true { a TRUE => { nil }; * => { nil } }))
-----------------------^
This is an implementation restriction that should be removed in a future version of Fuzion.

To solve this, create a helper feature 'lazy_value' that calculates the value as follows:

  lazy_value => {
  {
    match true
    a TRUE => {
      set #exprResult0 := nil
    }
    FALSE => {
      set #exprResult0 := nil
    }
    
    universe.this.#exprResult0.call
  }
}

and then use 'lazy_value' as instead of the original expression.

one error.
```
we need to check if this restriction should be relaxed or fully removed.

michaellilltokiwa:
The example now works but the implementation restriction is not yet fully fixed.

PR #2628, opened 2024-02-28
fe: surpress error `declarationsInLazy` in case of other errors

michaellilltokiwa:
this shows up frequently in the list of errors for me as the two last errors.


fridis (changes requested):
Good, these were always annoying!

Can you add a reference to #2559, best using a `NYI: UNDER DEVELOPMENT:` marker?

PR #2629, opened 2024-02-28
parser: if parsing fails result in plain expr instead of call

michaellilltokiwa:


michaellilltokiwa:
> can you explain what this _does_? the title seems very vague without context

Sorry. :-) I wanted to get rid of the special case in `loadCalledFeatureUnlessTargetVoid` which is only needed because the Parser creates a `new Call(..., error-string)` on syntax errors.

michaellilltokiwa:
> i'm a bit hesitant, the call had a `tokenSourcePos` included; is there possibly any information about the error that we lose by doing this?

I don't think so since in this case parsing has horribly failed anyway.

PR #2631, opened 2024-02-28
fe: add possibility to omit `.type.`

michaellilltokiwa:


fridis (commented):
Looks great! 

Will have a detailed look into it tomorrow.

tests/dot_type_call_ambiguity/dot_type_call_ambiguity.fz
```diff
@@ -0,0 +1,32 @@
...
+
+dot_type_call_ambiguity is
+
+  a => 3
+
+  type.a => 4
+
+  # NYI should result in an error: Duplicate feature declaration or similar.
+
+  say ex.a
```
fridis: `ex` does not seem to exist here.

tests/dot_type_call_ambiguity/dot_type_call_ambiguity.fz
```diff
@@ -0,0 +1,32 @@
...
+#
+# -----------------------------------------------------------------------
+
+dot_type_call_ambiguity is
+
+  a => 3
+
+  type.a => 4
+
+  # NYI should result in an error: Duplicate feature declaration or similar.
```
fridis: We can solve this later, but here are my thoughts: 

A `duplicate feature` is a bit difficult to define when you take visibility into account. 

I prefer to cause an error at an ambiguous call and to provide a means to disambiguate the calls:
```
ex is 
  a => 3
  type.a => 4

say ex.a    # should cause an error, ambiguous!
```
To make a call on the type feature, one could write
```
say ex.type.a
```
It is a bit harder to make a call on the instance, one ugly solution might be a temp field:
```
  temp := ex
  temp.a
```
Nicer might be to add an `instance` (or `self`, `me`, `this`) to `Any.fz`
```
public Any ref is
  ...
  public instance => Any.this
```
then we might be able to do
```
say ex.instance.a
```

fridis (commented):
I am very happy with this PR! Espacially, I think I could use some of this to later permit feature declarations to serve as type aliases, which also requires converting calls into types.

fridis (approved):
please create 2 issues for my unresolved comments.

PR #2632, opened 2024-02-29
tests: parallel use only half of cores returned by nproc.

michaellilltokiwa:
For my system `nproc --all` returns 12, hyperthreading cores are counted as well. But this makes my system a bit laggy.

[ci skip]


fridis:
Does this affect the github builds? Would like to see them run as fast as possible.

michaellilltokiwa:
@fridis Yes. I thought github runners have only one core available but this does not seem to be true (anymore). They have 2-4 cores available. I updated the logic to run up to six tests in parallel.

PR #2635, opened 2024-02-29
lib/io: introduce dir, an interface to traverse directories

maxteufel:


lib/io/dir/read.fz
```diff
@@ -0,0 +1,65 @@
...
+  #
+  module read(fd i64) outcome String =>
+    r := ro.read fd
+    replace
+    r
+
+
+  # the default directory reading operation reading the next entry via fuzion.sys.fileio.read_dir
+  #
+  type.default_read_handler : io.dir.Read_Handler is
```
michaellilltokiwa, line 42: This should probably be public

PR #2644, opened 2024-03-01
fe: automatic unwrapping of values

### LINKED ISSUE #1964: Automatic unwrap for `atomic`, `mutate.new`, etc.
Types that wrap a value such as `mutate.new` or `atomic` are inconvenient to use, e.g., you must do 
```
  s1 := a1.read + a2.read
  s2 := m1.get + m2.get
```
to add two atomic or mutable `i32` values. It would be nice if this happened automatically and we could write
 ```
  s1 := a1 + a2
  s2 := m1 + m2
```
i.e., we would implicitly call `read`/`get` if an operation is performed that is only defined for the unwrapped type. To do this, we could define a feature
```
  auto_unwrap(T type, E type... : effect) is
    unwrap T ! E is abstract
```
and have `atomic` and `mutate.new` inherit from this. Then the compiler should now about this and automatically add calls to `unwrap` where needed. Of course, the implementations of `unwrap` could then have some effects.

michaellilltokiwa:
fixes #1964


fridis (approved):
Very nice. Just would like to keep #1964 open until this also works if used as a call target, which should be done in a separate PR. 
To auto-unwrap `x`, e.g., in `x+1`, the auto-unwrapping should be done as part of call resolution whenever the called feature cannot be found.

PR #2650, opened 2024-03-04
lib: short-hand in String to get a codepoint's numeric representation

maxteufel:


lib/String.fz
```diff
@@ -751,6 +751,13 @@ public String ref : property.equatable, property.hashable, property.orderable is
     as_codepoints ∀ cp -> cp.is_ascii
 
 
+  public val u32
```
michaellilltokiwa, line 754: I think this needs a comment ;-)

fridis (commented):
I am not particularly happy about this, seems convenient but somewhat arbitrary. 
I think I prefer either changing the type of a String literal that consists of a single code point to `codepoint`, or adding a codepoint literal `'x'`, with a slight preference for the former.

PR #2669, opened 2024-03-07
parser: fix require condition when creating source range

michaellilltokiwa:
This led to lots of test failures in the language server.

### DIFF
--- a/include/posix.c
+++ b/include/posix.c
@@ -504,7 +504,7 @@ void fzE_init()
 int64_t fzE_thread_create(void* code, void* args)
 {
 #ifdef FUZION_ENABLE_THREADS
-  pthread_t * pt = fzE_malloc_safe(sizeof(pthread_t));;
+  pthread_t * pt = fzE_malloc_safe(sizeof(pthread_t));
 #ifdef GC_THREADS
   int res = GC_pthread_create(pt,NULL,code,args);
 #else
--- a/include/win.c
+++ b/include/win.c
@@ -543,7 +543,7 @@ void fzE_init()
 int64_t fzE_thread_create(void* code, void* args)
 {
 #ifdef FUZION_ENABLE_THREADS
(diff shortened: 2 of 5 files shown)


src/dev/flang/parser/Lexer.java
```diff
@@ -1198,7 +1198,9 @@ public SourceRange sourceRange(int prevPos, int pos, int endPos)
        endPos  <= byteLength());
 
     return pos < endPos ? sourceRange(pos, endPos)
-                        : sourceRange(prevPos, prevPos+1);
+                        : prevPos+1 <= byteLength()
+                        ? sourceRange(prevPos, prevPos+1)
+                        : null;
```
fridis, line 1203: Can you instead do
```suggestion
                        : sourceRange(prevPos, Math.min(byteLength(), prevPos+1));
```
Such that the position is not fully lost in this case?

fridis (changes requested):
Sorry for breaking this :-)
A change request to not lose the position.

include/posix.c
```diff
@@ -504,7 +504,7 @@ void fzE_init()
 int64_t fzE_thread_create(void* code, void* args)
 {
 #ifdef FUZION_ENABLE_THREADS
-  pthread_t * pt = fzE_malloc_safe(sizeof(pthread_t));;
+  pthread_t * pt = fzE_malloc_safe(sizeof(pthread_t));
```
fridis, line 507: unrelated, but ok.

include/win.c
```diff
@@ -543,7 +543,7 @@ void fzE_init()
 int64_t fzE_thread_create(void* code, void* args)
 {
 #ifdef FUZION_ENABLE_THREADS
-  pthread_t * pt = fzE_malloc_safe(sizeof(pthread_t));;
+  pthread_t * pt = fzE_malloc_safe(sizeof(pthread_t));
```
fridis, line 546: unrelated, but ok.

fridis (approved):

PR #2676, opened 2024-03-07
fe: change type of strconst to `codepoint` if exactly one codepoint

### LINKED ISSUE #2320: AoC issue: codepoint literals
This is inspired by the 1 Dec AdventOfCode puzzle: 
The code to get the unicode value from a single constant character is somewhat clumsy: 
```
  v := "0".as_codepoints.first.val
```
It might be nice to have codepoint literals similar to String literals, one possibility would be to use `' ` like a `char` constants in Java:
```
  v := '0'.val
```
An alternative would be to treat string literals that can be encoded as a single codepoint as a codepoint constant, but this would have the strange effect that `"I"` and `"you"` would have different types.

michaellilltokiwa:
This PR as a PoC to show what happens if StrConst is of type codepoint if exactly one codepoint.

fixes #2320


michaellilltokiwa:
@tokiwa-software/developers opinions please :-)

fridis (commented):
I think I like this. This will basically bring the behavior we have for numeric literals, where `1.1e0` is `f64` and  `1.1e1` is `i32` to `String` vs. `codepoint`, but maybe that is ok. 
We might (in a separate patch) make sure that this works well with `redef` as in 
```
  a is
    redef as_string => "a"
```

lib/array.fz
```diff
@@ -285,7 +285,8 @@ module:public array(
       safety: length ≥ 0
   =>
 
-    indices => 0..length-1
+    // NYI: BUG: when removing `interval i32`: Illegal forward or cyclic type inference
+    indices interval i32 => 0..length-1
```
fridis, line 289: Can you create an issue for this?

fridis (approved):

PR #2679, opened 2024-03-07
tests/fz_cmd: update expected error

maxteufel:
The new `)` got added in #2671 but it was forgotten to update the test outputs.


michaellilltokiwa (approved):
Thanks :-D

PR #2697, opened 2024-03-08
Dfa tailcall fix

fridis:


michaellilltokiwa:
Is it possible to put all logic that is needed in the DFA that specifically is about TCO in a separate class? This would help me understand this a little more.

PR #2698, opened 2024-03-08
backends: do not continue processing if DFA for Site always results i…

michaellilltokiwa:
…n void


michaellilltokiwa:
@fridis I believe there might be a better approach than this?

src/dev/flang/fuir/analysis/dfa/Site.java
```diff
@@ -69,6 +70,12 @@ public class Site extends ANY implements Comparable<Site>
   TreeSet<Integer> _accesses = new TreeSet<>();
 
 
+  /**
+   * Results of analyzing this Site
+   */
+  private List<Boolean> results = new List<Boolean>();
```
fridis, line 76: Instead of this list, I would just add a boolean flag `_mayReturn` and initialize it with `false`. Then, in `DFA.Analyse.call` and `DFA.Analyse.match`, you should set this flag to `true` if `res != null`. 
A further complication that should be tested and that might need special handling: What if the called feature has a precondition and the precondition does not return? Or a precondition that does return followed by a call that does not. Do these cases work well?

fridis (changes requested):
I think I would have done it similarly, A little worried that too many instances of `Site` will be created, but performance issues should be solved later.

fridis (approved):

PR #2699, opened 2024-03-08
util: change Pair from class to record

michaellilltokiwa:


fridis (approved):
Too bad that Java records require accessor methods, so they induce more typing for accesses...

PR #2720, opened 2024-03-12
Use DFA instead of CFG for missing effects errors

fridis:
This avoids the extra phase for CFG while the information required to produce errors for missing effects is already available during DFA.
    
Since DFA so far was not run with the interpreter, it is now done explicitly before the `Interpreter` is created.

The DFA analysis accuracy for Instances had to be increased to include effects in environment, otherwise function arguments were merged in a way that DFA could no longer distinguish those called with an effect installed and those without.
    
This gives a performance increase of about 4% for a HelloWorld:
```    
     > time for run in {1..10}; do ./build/bin/fz -e 'say "Hello World!"'; done
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    
    real    0m18,977s
    user    1m13,099s
    sys     0m3,082s
     > cd -
    /home/fridi/fuzion/clean/fuzion
     > time for run in {1..10}; do ./build/bin/fz -e 'say "Hello World!"'; done
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    Hello World!
    
    real    0m19,774s
    user    1m17,926s
    sys     0m3,563s
     > calc 19.774/18.977
    1.041998208357
```    
However, it adds the DFA phase to the interpreter which adds about 25% runtime to an interpreted HelloWorld...

Renamed `FUIR.clazzIntrinsicName` as `FUIR.clazzOriginalName` since this is no longer used only for intrinsics, but also for `effect.type.unsafe_get`.

Small changes were required in the base lib and tests to avoid false positive errors for not installed effects, in particular added `effect.type.get_if_installed`. Added a negative test `local_mutate_neg` that produces a compile-time missing effect error. 

This PR does not improve the produced error (#2722), but tries to keep the output unchanged for now.


michaellilltokiwa (approved):

src/dev/flang/be/effects/Effects.java
```diff
@@ -56,15 +54,15 @@ public class Effects extends CFG  // NYI: remove CFG, still used by Effects.chec
 
 
   /**
-   * Map from clazz cl to set of effects ecl that are required for a call to cl.
+   * The options provided to the fz comment.
```
fridis, line 57:
```suggestion
   * The options provided to the fz command.
```

PR #2721, opened 2024-03-13
Java interface: make it possible to pass objects of value null to Java

maxteufel:


src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -1910,6 +1910,14 @@ static void setArrayElementsToAnything(Call cl, int argnum, String intrinsicName
         jobj.setField(cl._dfa, jref, Value.UNKNOWN_JAVA_REF);
         return jobj;
       });
+    put("fuzion.java.null_to_java_object", cl ->
+      {
+        var rc = cl._dfa._fuir.clazzResultClazz(cl._cc);
```
michaellilltokiwa, line 1915: @maxteufel what if you replace all of this up to return by newFuzionJavaCall(cl) as in `fuzion.java.call_c0`

lib/String.fz
```diff
@@ -39,6 +39,10 @@ public String ref : property.equatable, property.hashable, property.orderable is
   # returns true if string is empty or contains whitespace only
   public is_blank => utf8 ∀ u -> u.is_ascii_white_space
 
+  # true if this string represents an object String that is null
+  # when passed on to the Java interface
+  module is_java_null_string => false
```
michaellilltokiwa, line 44: I would prefer if this where not in String but in fuzion.java:
```
fuzion.java is
...
    module String.is_java_null_string => false
    public String.type.java_null_string String => ...
```
Not sure if this is possible?...

lib/fuzion/java.fz
```diff
@@ -207,7 +207,10 @@ public fuzion.java is
   # convert a string to a Java String object
   #
   public string_to_java_object(s String) Java_Object =>
-    string_to_java_object0 s.utf8.as_array.internal_array
+    if s.is_java_null_string
+      null_to_java_object Java_Object
```
michaellilltokiwa, line 211: could this be `Java_String` instead of `JavaObject`?

PR #2726, opened 2024-03-14
be/interpreter: fix ClassCastException

michaellilltokiwa:
Problem was that the test socket did not use interpreter for starting server. Now it does but had to fix two intrinsics.

Also some unrelated cleanups/improvements that I did not want to add yet another PR for.


fridis (changes requested):
I think there are more changes in this PR than what you intended to put into it...
Please split up into several PRs.

fridis (approved):

PR #2727, opened 2024-03-14
macOS: try fix finding libjvm.dylib

michaellilltokiwa:
I did not test this.

[ci skip]


fridis (changes requested):
Good if this fixed the macOS builds.

Can you add some context into the commit message or as a comment to the code, in particular: 

- what error are you trying to fix (I guess the failing MacOS builds, but some detail would be good)
- why do you think that this solves the problem and where did you get this information from 

Thanks.

michaellilltokiwa:
@fridis yeah, I should have added this information...

fridis (approved):

PR #2732, opened 2024-03-14
dfa/tests: Improve error produced in case of not installed effect

### LINKED ISSUE #2722: Poor error message for not installed effect
This error:
```
 > ./build/bin/fz -e "lm : mutate is {}; _ := lm.env"

error 1: Used effect `lm` never instantiated.

one error.
```
is wrong (the problem is that the effect is not installed in this context, it may well be instantiated and installed in other contexts) and missing any context like a source code position or, better, a call trace leading to the problem.

fridis:
Add positive and negative tests for not installed effects, fix #2722.

A typical error for a missing effect now looks like this:
```
part2a.fz:4:15: error 1: Failed to verify that effect 'a' is installed in current environment.
use_a => _ := a.env
--------------^^^^^

effect environment 'c' for call to use_a at part2a.fz:27:10:
c.go ()->use_a
---------^^^^^
effect environment 'c' for call to #fun8.call at $MODULE/effect.fz:207:23:
    cf := Effect_Call f
----------------------^
effect environment 'c' for call to (c.go unit).#fun3.call at $MODULE/effect.fz:153:18:
      set res := f()
-----------------^
effect environment 'c' for call to (c.Effect_Call unit).call
install effect 'c', old envionment was '--empty--' for call to c.abortable (c.Effect_Call unit) at $MODULE/effect.fz:208:5:
    abortable cf
----^
effect environment '--empty--' for call to c.go unit at part2a.fz:15:3:
c.go unit fc
--^^
effect environment '--empty--' for call to universe#0
effect environment '--empty--' at program entry
```


src/dev/flang/fuir/analysis/dfa/Call.java
```diff
@@ -322,12 +324,36 @@ public String toString()
       .append(r == null ? "*** VOID ***" : r);
     if (_env != null)
       {
-        sb.append(_env.toString());
+        sb.append(" ENV: ");
+        sb.append(dev.flang.util.Terminal.INTENSE_PURPLE + _env.toString() + dev.flang.util.Terminal.RESET);
```
michaellilltokiwa, line 328: I think `Errors.effe` could be used here as well since it is already used in this file below.

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -1385,6 +1385,12 @@ static void setArrayElementsToAnything(Call cl, int argnum, String intrinsicName
   static void setArrayI64ElementsToAnything(Call cl, int argnum, String intrinsicName) { setArrayElementsToAnything(cl, argnum, intrinsicName, FUIR.SpecialClazzes.c_i64); }
 
 
+  /**
+   * Name of intrinsic `effect.abortable`.
+   */
+  static final String _effect_abortable_name_ = "effect.abortable";
```
michaellilltokiwa, line 1391: This should probably be moved to `FuzionConstants`

src/dev/flang/fuir/analysis/dfa/Context.java
```diff
@@ -39,9 +42,32 @@ public interface Context
...
+      return "effect environment " + Errors.effe(Env.envAsString(null)) + " at program entry";
+    }
+  };
+
+
   /*----------------------------  constants  ----------------------------*/
 
 
+  /**
+   * Singleton instnace of MainEntryPoint.
```
michaellilltokiwa, line 66: instance

src/dev/flang/fuir/analysis/AbstractInterpreter.java
```diff
@@ -63,8 +65,51 @@ public class AbstractInterpreter<VALUE, RESULT> extends ANY
...
    * that processes this code.
    */
-  public static abstract class ProcessStatement<VALUE, RESULT> extends ANY
+  public static abstract class ProcessStatement<VALUE, RESULT> extends ANY implements HasSourcePosition
   {
+
+    /**
+     * FUIR we are currently interpreting
+     */
+    FUIR _fuir;
```
michaellilltokiwa, line 74: What is the benefit of having a _fuir field here? Is it not possible to use the final _fuir field of AbstractInterpreter?

src/dev/flang/fuir/analysis/AbstractInterpreter.java
```diff
@@ -610,6 +655,12 @@ public RESULT process(int cl, boolean pre, Stack<VALUE> stack, int c, int i)
         say("process "+_fuir.clazzAsString(cl)+"."+c+"."+i+":\t"+_fuir.codeAtAsString(cl, c, i)+" stack is "+stack);
       }
     var s = _fuir.codeAt(c, i);
+
+    _processor._fuir = _fuir;
```
michaellilltokiwa, line 659: Since this only needed for obtaining the SourcePosition I think there should be an option/setting for the Processor to choose if we really need this in the current processor. E.g. for the Interpreter2 I may want to turn this off.

michaellilltokiwa (approved):

src/dev/flang/fuir/analysis/AbstractInterpreter.java
fridis: The `Processor` currently is a static class, so it cannot access `_fuir` from the surrounding class. It might, however, make sense to change this.

src/dev/flang/fuir/analysis/AbstractInterpreter.java
fridis: I do not think this is very expensive, but I would like to further reduce the overhead of this as follows

- `c`/`i` should be joined into a `site` id, see #2473 
- `cl`/`pre` should become a single `int`, not sure if that should be a clazz id or something new 
- maybe, the `site` id could also contain combined `cl`/`pre`. Currently, I think that clazzes that origininate in the same feature share their code blocks, but it might make sense to replicate the code blocks to get unique `site` id. 
- Finally, if  `Processor` is not static class `_fuir` could beremoved. 

So, eventually, we might end up with a single `int` to identify the site.

PR #2743, opened 2024-03-15
dfa: add call site to context of Instance and Call

fridis:
DFA: Take call site into account when tracing instances and calls

This increase the DFA accuracy such that tests like flang.dev's `content/design/examples/monadiclifting_helloabort.fz` no longer produce errors.

The general pattern of code that is improved by are calls that differ by their location, i.e., in the code

```
(io.out p1).go (() -> do_something)
(io.out p2).go (() -> do_something_else)
```

we can now distinguish the two instances of `io.out` such that their print handlers `p1` and `p2` will not be mixed up.

For this, support for site indices was added to IR.java. A site index is a single integer that combines a code block index and an expr index in that code block into a single value.  Eventually, code block indices and expr indices should be removed completely from IR and FUIR and replaced by site indices where site..site+n corresponds to (c,i=0)..(c,i=n).

This patch adds helper a methods `IR.addCode`/`IR.getExpr` that are used in `IR.java` and `MIR.java`.

Since the more accurate analysis results in `error` instances produced by the interface to Java code to be treated separately, this patch adds code to properly initialize the `msg` field in those `error` instances.

### DIFF
--- a/bin/run_tests_parallel.sh
+++ b/bin/run_tests_parallel.sh
@@ -69,7 +69,7 @@ rm -rf "$BUILD_DIR"/run_tests.failures
 # print collected results up until interruption
 trap "echo """"; cat ""$BUILD_DIR""/run_tests.results ""$BUILD_DIR""/run_tests.failures; exit 130;" INT
 
-N=$(($(nproc --all || echo 1)>6 ? 6 : $(nproc --all || echo 1)))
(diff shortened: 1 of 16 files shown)


fridis:
This has been replaced by #2750, #2751, #2752, #2753, #2754 and #2755, so closing this PR.

PR #2746, opened 2024-03-18
types/clazzes: use upper-/lower-case to match fuzion source code

michaellilltokiwa:


fridis (approved):
Yes, should be like that. 
Only the commit message should be changed. `Use upper and lowercase in feature names` or `respect case in feature names` or similar, `caseing` does not exist.

PR #2756, opened 2024-03-19
lib: add effect to reuse internal_arrays used for intrinsics results

michaellilltokiwa:
also removed some duplicate `replace`s
and made `read_lines` implementation a bit more readable.

### DIFF
--- a/lib/fuzion/sys/fileio.fz
+++ b/lib/fuzion/sys/fileio.fz
@@ -68,10 +68,11 @@ module fileio is
   private get_file_size(
                         # the (relative or absolute) file name, using platform specific path separators
                         path String) outcome i64 =>
-    md := fuzion.sys.internal_array_init i64 4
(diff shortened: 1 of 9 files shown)


fridis (commented):
I needed some time to understand the purpose of this. Apparently, you want to avoid the overhead of allocating this temporary array. I would not be too worried about this allocation, I expect DFA with better lifespan analysis to be able to tellt the backends to allocate these one the stack or thread-locally. 
This effect would result in `fz -effects` reporting the need for `fuzion.sys.heap` effect for any code using feature that use this effect. This might be confusing.

michaellilltokiwa:
@fridis The idea is to avoid heap allocations in exchange for always using thread local arrays when using intrinsics. I'm not at all sure if this is worth it. The reporting of the effect is something I thought about as well. But IMHO effects that are not visible should not be reported by `-effects`... So this may need some change?

PR #2758, opened 2024-03-19
ast/Loop: replace streams with a list-based approach

maxteufel:
Replace the usage of `Stream` in `ast/Loop.java` by a list based approach. The documented logic has been updated to represent the new version. This depends on #2757. This does not remove `Stream` from the standard library yet, this will happen in a follow-up pull-request.

Marking as a draft for now because there is a single test failure related to the stack map table creation in `tests/reg_issue2376`.

### DIFF
--- a/src/dev/flang/ast/Feature.java
+++ b/src/dev/flang/ast/Feature.java
@@ -276,7 +276,10 @@ public class Feature extends AbstractFeature
   public boolean isIndexVarUpdatedByLoop() { return _isIndexVarUpdatedByLoop; }
 
 
-  boolean _newValueWhatever = false;
+  /**
+   * Is this a loop's variable that is being iterated over using the `in` keyword?
+   */
+  boolean _isLoopIterator = false;
 
 
   /**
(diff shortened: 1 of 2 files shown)


michaellilltokiwa (approved):
Very cool!

michaellilltokiwa:
This patch fixes the issue of test reg_issue2376 but creates other problems.
The root cause of the problem is that the loop code creates a match in universe and the case field is also created in universe.
This does not seem to work yet with the jvm-backend...

```patch
diff --git a/src/dev/flang/be/jvm/Types.java b/src/dev/flang/be/jvm/Types.java
index 787340b36..7d2d660c3 100644
--- a/src/dev/flang/be/jvm/Types.java
+++ b/src/dev/flang/be/jvm/Types.java
@@ -444,8 +444,12 @@ public class Types extends ANY implements ClassFileConstants
       case c_sys_ptr -> JAVA_LANG_OBJECT;
       default        ->
         {
-          if (cl == _fuir.clazzUniverse()                        ||
-              !_fuir.clazzIsRef(cl) && _fuir.clazzIsUnitType(cl) ||
+          if (cl == _fuir.clazzUniverse())
+            {
+              yield UNIVERSE_TYPE;
+            }
+          else if (
+             !_fuir.clazzIsRef(cl) && _fuir.clazzIsUnitType(cl) ||
               _fuir.clazzIsVoidType(cl)  // Java's void is not really the same, but Java does not have a real 'void' type.
               )
             {
```

PR #2772, opened 2024-03-21
lib: remove uses of `private` keyword

michaellilltokiwa:
I think this makes the code a bit more readable. I hope you agree?


fridis (approved):
I do agree.

PR #2776, opened 2024-03-21
java intrinsics: change `array_get` from intrinsic_constructor to ins…

michaellilltokiwa:
…trinsic

this fixes "Used abstract feature.." in javabase test:
```
  # NYI does not work yet: error 1: Used abstract feature 'String.utf8' is not implemented by 'String'
  # say (fuzion.java.array_to_java_object ["1","2","3"])
```
but now a new error:
java.lang.ClassCastException: class fzC_Const_String cannot be cast to class fzC_String (fzC_Const_String and fzC_String are in unnamed module of loader dev.flang.be.jvm.Runner @658c5a19)


fridis:
> java.lang.ClassCastException: class fzC_Const_String cannot be cast to class fzC_String

This should probably cast to `fzI_String`, `fzC_String` should only be used if the JVM backend is sure that there are no instantiated child features of `fzC_String`. So the code that decides if the need an interface or only a class depending on the children does not work as expected here.

fridis (approved):

PR #2778, opened 2024-03-21
tests: int, reduce amount of numbers tested

michaellilltokiwa:
it is unecessary to test all these numbers and this change reduces runtime in interpreter

[ci skip]

### DIFF
--- a/tests/int/int_test.fz
+++ b/tests/int/int_test.fz
@@ -36,8 +36,8 @@ int_test is
 
 
   # test basic int operations with small numbers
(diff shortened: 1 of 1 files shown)


fridis (commented):
Is the performance gain relevant? I would rather increase the number of values tested unless the cost is really high!

michaellilltokiwa:
@maxteufel @fridis 

rough numbers:
old:   real    3m43,813s
new: real    0m55,271s

PR #2783, opened 2024-03-22
readme: add macos section

michaellilltokiwa:
[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ Scorecard](https://api.securityscorecards.dev/projects/github.com/tokiwa-softwar
    * [Documentation](#documentation)
    * [Clone](#clone)
    * [Requirements](#requirements)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -157,6 +159,22 @@ Check [fuzion-lang.dev](https://fuzion-lang.dev) for language and implementation
 - GNU make
 - libgc
 
+### MacOS
+
+> This command should install all requirements:
```
fridis, line 164: Independent of this patch, but also affecting the other sections: `requirements` is not correct here, should be something like `required packages` or similar. Can you change this in a separate PR?

fridis (approved):

README.md
michaellilltokiwa: @fridis maybe: dependencies?

PR #2787, opened 2024-03-25
ast: replace NYI by explanatory comment

michaellilltokiwa:
[ci skip]


michaellilltokiwa:
Btw. I'm not sure the explained reason is the only reason why outerrefs are needed for abstract features.

PR #2789, opened 2024-03-25
fix `AbstractFeature.toString()` for states below FINDING_DECLARATIONS

michaellilltokiwa:


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -1730,7 +1730,7 @@ public Expr visit(FeatureVisitor v, AbstractFeature outer)
    */
   public String toString()
   {
-    return visibility() + " " +
+    return (state().atLeast(State.FINDING_DECLARATIONS) ? visibility() + " " : "") +
```
fridis, line 1733: Would be much better to move this special handling to `visibility()` which already has a `NYI:` comment when `outer()` is called.

michaellilltokiwa: @fridis I agree :-)

PR #2796, opened 2024-03-25
be/interpreter: re-enable call stacks

michaellilltokiwa:


src/dev/flang/be/interpreter/Excecutor.java
```diff
@@ -565,10 +569,100 @@ public Object contract(int cl, ContractKind ck, Value cc)
    */
   Value callOnInstance(int cc, Instance cur, Value outer, List<Value> args, boolean pre)
   {
+    FuzionThread.current()._callStackFrames.push(_fuir.clazzForInterpreter(cc));
+    var pos = this.pos();
```
michaellilltokiwa, line 573: @fridis How bad is this considering performance?

fridis: `Thread.current()` is very cheap, `FUIR.clazzForInterpreter()` is probably slightly more expensive than just pushing `cc`, which will be turned into `Integer.valueOf(cc)`. 
I fear that `this.pos()` might be the most expensive part, maybe add `AbstractInterpreter.site()` that returns `_fuir.siteFromCI(_c,_i)`, that is maybe most future-proof.
We might even make a site unique for each outer clazz in the future, then you should be able to drop the `_callStackFrames` as well and use something like `_fuir.siteOuterClazz(site)` or similar.

michaellilltokiwa:
@maxteufel Yes. Currently all expected_err containing contract failures are broken for the interpreter. With this PR the expected_err should be almost the same as the ones recorded with the old interpeter. The difference is that previously e.g. call.toString was used, now clazzAsStringNew is used which gives subtlety different output.

PR #2817, opened 2024-03-28
fe/me: always use .fum file as input for middle end

michaellilltokiwa:
Since errors can not be written to fum file, I needed to add a `Errors.showAndExit()` after `new FrontEnd()`

### DIFF
--- a/src/dev/flang/tools/Fuzion.java
+++ b/src/dev/flang/tools/Fuzion.java
@@ -471,7 +471,7 @@ public class Fuzion extends Tool
     {
       var mir = fe.createMIR();                                                           f.timer("createMIR");
       var air = new MiddleEnd(fe._options, mir, fe.mainModule() /* NYI: remove */).air(); f.timer("me");
-      var fuir = new Optimizer(fe._options, air).fuir();                              f.timer("ir");
+      var fuir = new Optimizer(fe._options, air).fuir();                                  f.timer("ir");
       process(fe._options, fuir);
     }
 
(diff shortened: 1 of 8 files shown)


fridis (commented):
This looks very promising. The hard part is to get rid of the awful dependency from `Clazz` to `SrcModule`. 
In the current version of your solution, you instead introduce a dependency from `Clazz` to `Module`. I first thought this would introduce a cyclic dependency, but it does not. Nevertheless, it might be better to have a dedicated class `fe.FeatureLookup` or similar that contains only the essential methods needed in `air` to make it easier to change this if needed. Maybe `FeatureLookup` could be just an interface that is implemented by `Module`.

src/dev/flang/tools/Fuzion.java
```diff
@@ -472,7 +470,7 @@ boolean takesApplicationArgs()
     void processFrontEnd(Fuzion f, FrontEnd fe)
     {
       var mir = fe.createMIR();                                                       f.timer("createMIR");
```
fridis, line 472:
```suggestion
      var mir = fe.createMIR();                                                           f.timer("createMIR");
```

src/dev/flang/tools/Fuzion.java
```diff
@@ -472,7 +470,7 @@ boolean takesApplicationArgs()
...
-      var air = new MiddleEnd(fe._options, mir, fe.module() /* NYI: remove */).air(); f.timer("me");
+      var air = new MiddleEnd(fe._options, mir, fe.mainModule() /* NYI: remove */).air(); f.timer("me");
       var fuir = new Optimizer(fe._options, air).fuir();                              f.timer("ir");
```
fridis, line 474:
```suggestion
      var fuir = new Optimizer(fe._options, air).fuir();                                  f.timer("ir");
```

fridis (approved):
Very nice. Just two minor suggestions to satisfy my obsession with vertical alignment.

PR #2846, opened 2024-04-06
fe: write universe to fum-file

michaellilltokiwa:


src/dev/flang/fe/LibraryModule.java
```diff
@@ -544,6 +544,10 @@ else if (k == -1)
    *   +        +--------+---------------+-----------------------------------------------+
    *   |        | m      | DeclFeatures  | features declared in this module              |
    *   +        +--------+---------------+-----------------------------------------------+
+   *   |        | 1      | hasUniverse   | if this module has a universe written out     |
+   *   +        +--------+---------------+-----------------------------------------------+
+   *   |        | 0-1    | universe      | universe                                      |
+   *   +        +--------+---------------+-----------------------------------------------+
```
fridis, line 550: A few issues here (but read all before fixing anything): 

- the type `hasUniverse`should be `bool`
- the condition of `universe` should be `hasUniverse`, as follows
- the asciidoc above would need to be changed as well (bad that this is so redundant, need to check how to make the asciidoc more readable and remove the additional tables).

So this should better be

```suggestion
   *   |        | 1      | bool   | hasUniverse: 1 if this module has a universe written out, 0 otherwise    |
   *   +-------+--------+---------------+-----------------------------------------------+
   *   | hasUniverse | 1    | universe      | universe feature                              |
   *   +        +--------+---------------+-----------------------------------------------+
```
But then, I think it would be much better if the universe would just be written as part of the `DeclFeatures`, then we do not need a more complex file format, but we need a way to identify the universe, e.g., by a special name or by `outer` having a special value or similar.

src/dev/flang/fe/LibraryOut.java
```diff
@@ -152,6 +156,12 @@ class LibraryOut extends ANY
             moduleRef(m);
           }
         allDeclFeatures(sm);
+        var universeContainsCode = !sm._universe.impl().containsOnlyDeclarations();
+        _data.writeBool(universeContainsCode);
+        if (universeContainsCode)
+          {
+            feature(sm._universe);
+          }
```
fridis, line 164: Should IMHO be part of `allDeclFeatures`.

fridis (changes requested):
I hope we can achieve this with fewer changes to the .fum file format, see details in my comments.

fridis (approved):

PR #2857, opened 2024-04-09
ast: AbstractType improve structure and documentation

michaellilltokiwa:


fridis (changes requested):
Looks so much better! I am not proud of my original code... 
One point where I would be a bit less specific in the documentation is the  `@throws Error` entries: Noone should rely on these to be thrown, so I would either make them less strict
```
@throws Error if corresponding checks are enabled and this is not resolved or isGenerics()
```
or just make this part of the main documentation
```
 Requires that this is resolved and !isGenerics().
```
and do not mention what happens otherwise.

fridis (approved):

PR #2858, opened 2024-04-09
ast: change loop scoping semantics

michaellilltokiwa:
previously this worked:
```
  for i := 10 do
  say i
```
but this does not work:
```
  for x in 1..10
        i := 10
  do
  say i
```

This patch changes the semantics of the loop in a way that the first example also results in a feature not found error.

I think this makes loop scoping more intuitive.

We loose the ability to access any non-iterating index variables until the first iterating index variable.

@tokiwa-software/developers Not sure if this change of semantics is desirable but created this PR to get your opinions.


fridis (approved):
This is an important fix, the original scoping is just confusing. If any index variable should be accessed after the loop, it should be returned by an `else` clause as in 

```
  i := for i0 := 10 while cond else i0
  say i
```

PR #2869, opened 2024-04-10
java: suppress unchecked and rawtypes warnings

michaellilltokiwa:


fridis (approved):
I would very much prefer not to use `@SuppressWarnings`, but I see that Java's type erasure for generics makes this very hard in some of these cases. Generics in Java are broken...
Anyway, if someone finds time, would be good to get rid of some of these warnings.

PR #2874, opened 2024-04-10
interpreter: remove dependency on AST

michaellilltokiwa:


fridis (approved):
Nice to see that code is getting simpler when dependencies are removed!

PR #2881, opened 2024-04-11
lib: add io.buffered.read_line_while

michaellilltokiwa:


lib/io/buffered/reader.fz
```diff
@@ -248,3 +248,18 @@ public read_lines array String ! reader =>
     .split "\n"
     .map (s -> if s.ends_with "\r" then s.substring 0 s.byte_length-1 else s)
     .as_array
+
+
+# Read input line by line while `f` is true.
```
fridis, line 253:
```suggestion
# Read input line by line calling `f` for each line until `f` returns false
# or end_of_file is reached.
```

fridis (commented):
just minor comments...

PR #2884, opened 2024-04-11
ast: show warning if -> is used in operator

### LINKED ISSUE #2878: lambda arrow is parsed as part of operator
```
ex =>

  a (f String -> bool) is

  a s->!s.is_empty # does not work: ->! is parsed as operator
  a (s-> !s.is_empty) # works
```

michaellilltokiwa:
fixes #2878

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -215,6 +215,16 @@ public class Call extends AbstractCall
   Call _movedTo = null;
 
 
+  /**
+   * This field is set to the target of the call
+   * before calling typeForCallTarget().
+   *
+   * Can then be used to produce errors that are target
+   * dependent.
+   */
+  public Call _targetOf_forErrorSolutions = null;
+
+
   /*-------------------------- constructors ---------------------------*/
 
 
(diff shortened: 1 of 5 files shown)


fridis (changes requested):
Very helpful to have this information in case of this error. 

However, it would be much nicer to not have a warning, but to have a suggested solution for the error that `s` is not found. 

To get this, you might do something like this: 
```
  if (_target instanceof Call tc) 
    {
      tc._targetOf_forErrorSolutions = this;
    }
```
before the call to `typeForCallTarget` and then use this field when producing the feature not found error for the call to `s` to produce a suggested solution.

src/dev/flang/ast/Call.java
```diff
@@ -821,14 +837,23 @@ private boolean loadCalledFeatureUnlessTargetVoid(Resolution res, AbstractFeatur
...
+                        && this._targetOf_forErrorSolutions._name.length() > "infix ->".length())
+                      {
+                        AstErrors.usingLambdaOpertatorAsPrefixInOperator(this._targetOf_forErrorSolutions.pos());
+                      }
+                    else
+                      {
+                        AstErrors.calledFeatureNotFound(this,
+                                                        calledName,
+                                                        tf,
+                                                        _target,
```
fridis, line 851: I would still report the `calledFeatureNotFound` error since that might still be the problem in case of a valid use of an infix operator like `->>`, but add the suggested solution only in this case.

fridis (changes requested):
Getting better, but I was not clear enough in my original comment, see explanation below.

fridis (approved):
Very nice now, and the logic is mostly within `AstErrors`, which is where it should be.

PR #2896, opened 2024-04-15
lib: move transducers to module

michaellilltokiwa:


fridis (changes requested):
I prefer having this kind of functionality in base.fum and not have too many micro-modules.

PR #2897, opened 2024-04-15
lib: move stopwatch/bench to module perf

michaellilltokiwa:


fridis (changes requested):
I prefer having this as part of base.fum.

PR #2907, opened 2024-04-15
fuir: Replace code/index pairs by site in FUIR APIs

fridis:


src/dev/flang/ir/IR.java
```diff
@@ -446,54 +344,67 @@ else if (e instanceof Check c)
...
     if (PRECONDITIONS) require
-      (ix >= 0, withinCode(c, ix));
+      (s >= SITE_BASE, withinCode(s));
 
+    return exprKind(getExpr(s));
+  }
+
+
+  /**
+   * Helper for `codeAt` to determine the ExprKind forn an Object that is either
```
michaellilltokiwa, line 398: forn -> for

src/dev/flang/fuir/FUIR.java
```diff
@@ -2337,45 +2295,37 @@ public int codeIndex(int c, int ix, int delta)
...
    * @param delta the negative number of instructions to go back.
    *
-   * @return the index of the expression 'n' expressions before 'ix', or a
+   * @return the site of the expression 'delta' expressions before 's', or a
    * negative value '-m' if that instruction can be found 'm' recursive calls up.
    */
-  private int codeIndex2(int c, int i, int ix, int delta)
+  private int codeIndex2(int si, int s, int delta)
   {
-    if (i == ix)
```
michaellilltokiwa, line 2353: What is the meaning of this?

michaellilltokiwa (changes requested):

src/dev/flang/fuir/FUIR.java
fridis: The meaning is that `c`/`i` is at the site we are looking for. This has been replaced by `si == s` a few lines below, where I added a corresponding comment to the `return` since I found the original code too confusion.

michaellilltokiwa: I read this the wrong way. I thought the line with if is still present, but it has been removed...

michaellilltokiwa (approved):

PR #2911, opened 2024-04-16
lib: add `abortable_effect`

michaellilltokiwa:
The idea here is that all abortable effects must inherit from `abortable_effect`. In the future we can then do cleanup of any opened resources during the computation of abortable effects.

I also changed the default for effect.abortable to false.


fridis (changes requested):
I like to see this code being cleaned up, in particular to have `go`  instead of `run`. However, this patch takes some power from the effect mechanism that I would like to keep:

- a `try` effect should IMHO be possible to implement in user code, which is currently not the case since `run` is not `public` (which was already the case before your patch) and it becomes harder if an `abort` does not provide a way to return a different values, which is made harder or impossible by this patch.
- I think `return` is the standard algebraic effect term for `abort`, so when this is removed, `abort` should at least be documented that this performs a `return` in the algebraic-effect sense.

BTW: `try` currently uses `set` in case an exception is thrown, but this can be avoided even without using `mutate` be creating a new instance of `try` in performing a `replace` with this new value.

Can you check the effect examples from fuzion-lang.dev if they work after this patch?

michaellilltokiwa:
> I like to see this code being cleaned up, in particular to have `go` instead of `run`. However, this patch takes some power from the effect mechanism that I would like to keep:
> 
>     * a `try` effect should IMHO be possible to implement in user code, which is currently not the case since `run` is not `public` (which was already the case before your patch) and it becomes harder if an `abort` does not provide a way to return a different values, which is made harder or impossible by this patch.
> 
>     * I think `return` is the standard algebraic effect term for `abort`, so when this is removed, `abort` should at least be documented that this performs a `return` in the algebraic-effect sense.
> 
> 
> BTW: `try` currently uses `set` in case an exception is thrown, but this can be avoided even without using `mutate` be creating a new instance of `try` in performing a `replace` with this new value.
> 
> Can you check the effect examples from fuzion-lang.dev if they work after this patch?

Sorry for trying to take away `return` again. This was due to sloppiness.

I fail to understand your comment about not being able implement a custom try effect. This should IMHO still be possible with this patch?

Regarding the replace suggestion, is there a reason `effect.replace` is not public? Also this only works if the effect is not inherited, right?

lib/effect.fz
```diff
@@ -36,19 +36,24 @@ module:public effect (
...
 
+  # evaluate the current mode of the effect
+  #
   match r
-    _ effect_mode.plain   =>
-    i effect_mode.inst    => run i.f ()->unit
     _ effect_mode.repl    => replace
     _ effect_mode.default => default
-    _ effect_mode.new     =>
+    * =>
```
fridis, line 44: Better keep all the single cases and not use `*`, this helps when adding a new modes to find all uses. Also, `effect_mode.inst` has a side effect, but it is unclear if this is still used anywhere.

fridis (commented):
Just a minor comment on `match *`. 
Can you fix the merge conflicts?
Otherwise, I think this basically looks ok.

PR #2925, opened 2024-04-18
lib: rename `option.get default` -> `option.or default`

michaellilltokiwa:


lib/option.fz
```diff
@@ -96,9 +96,17 @@ is
...
 
   # synonym for infix >>=
+  #
   public and_then (f T -> option T) => option.this >>= f
 
 
+  # if this option is nil return the result of f
+  # otherwise just return this option.
+  #
+  public or_then (f () -> option T) option T =>
```
fridis, line 106: Should IMHO be called `or_else`, or just `or`, this is at least the name chosen in Java's `Optional`.  The type of `f` should be `Lazy (option T)` for symmetry with `get`/`or`/`or_else` (whatever the final name will be) below.

lib/option.fz
```diff
@@ -131,7 +139,7 @@ is
 
   # unwraps an option if it exists, returns default value otherwise.
   #
-  public get(default Lazy T)
+  public or(default Lazy T)
```
fridis, line 142: Should IMHO be called `or_else`, this is at least the name chosen in Java`s `Optional`.

fridis (changes requested):
Nice. 
I have checked the names of the corresponding methods in Java's `Optinal` and I prefer the ones used there, but it might make sense to check other languages as well.

fridis (approved):

PR #2930, opened 2024-04-19
be/interpreter: Layout, performance

michaellilltokiwa:
In my tests yields about 10% performance gain since size() if called very frequently.


fridis (approved):

michaellilltokiwa:
@fridis okay, your suggestion sound even better

fridis (approved):

PR #2931, opened 2024-04-19
util: fix a comment

michaellilltokiwa:
[ci skip]

### DIFF
--- a/src/dev/flang/util/Map2Int.java
+++ b/src/dev/flang/util/Map2Int.java
@@ -136,7 +136,9 @@ public class Map2Int<T> extends ANY
   /**
    * get the index of an element
    *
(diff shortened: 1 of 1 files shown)


src/dev/flang/util/Map2Int.java
```diff
@@ -136,7 +136,7 @@ public int get(T t)
   /**
    * get the index of an element
    *
-   * @return t's index or _base-1 if t was not added.
+   * @return the T that belongs to index i.
```
fridis, line 139: suggesting to be more explicit here:
```suggestion
   * @param i an index returned by a call to add
   *
   * @return the T passed to add when it returned i.
```

fridis (approved):

PR #2940, opened 2024-04-19
util: Precalculate results of SourceFile.codePointInLine

fridis:
This speeds up building java.base.fum from 1m44s to 27s on my machine.

PR #2941, opened 2024-04-20
Speedup line num

fridis:
This is based on PR #2940 that should be merged first, only the added commits need to be reviewed. 

This avoids binary search for line numbers by storing line numbers with every position of a code point.

### DIFF
--- a/src/dev/flang/util/SourceFile.java
+++ b/src/dev/flang/util/SourceFile.java
@@ -124,6 +124,13 @@ public class SourceFile extends ANY
   /*-----------------------------  statics  -----------------------------*/
 
 
+  /**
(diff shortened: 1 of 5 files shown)


src/dev/flang/util/SourceFile.java
```diff
@@ -620,37 +663,98 @@ public void setPos(int newPos)
    * line. result[0] is unused and set to -1,
    *
    * The result is cached in _lines.
+   *
+   * As a side-effect, this determines the indenetation data for
```
michaellilltokiwa: indentation

michaellilltokiwa (approved):

PR #2953, opened 2024-04-22
parser/ast: Always parse actual as expr, convert to type as needed

fridis:
The parser no longer parses actual arguments twice, as expression and as type, but only as an expression.  If the actual argument turns out to e used as a type argument, it is converted to a type later one using `Expr.asParsedType`.

For this to work, the syntax for a lambda had to be relaxed to allow an arbitrary tuple as the left-hand-side of the `->`, and this tuple is later converted into a list of variable names (if used as a lambda) or a list of types (if used as a function type).

The argument `allowTypeThatIsNotExpression` of `Parser.skipType` was alway `true` and could hence be removed.

Renamed `Expr.asUnresolvedType` as `Expr.asParsedType` and (re-) implemented it for `Block`, `Function`,

Removed class `Actual`, actual arguments are now always parsed as `opExpr` and converted to a type using `Expr.asParsedType` if needed.

Removed `Call._actualsNew`, using `Call._actuals` instead.

Moved following methods and fields from `Call` to `ParsedCall`, mostly without any change: `NO_PARENTHESES`, `findChainedBoolean`, `isChainedBoolRHS`, `isInfixOperator`, `chainedBoolTarget`, `propagateExpectedTypeForPartial`, `checkPartialAmbiguity`, `applyPartially`, `splitOffTypeArgs`, `pushCall`,  `resolvedImplicitSelect`, `resolveImmediateFunctionCall`, `isImmediateFunctionCall`.

Removed now unused `Call.hasParentheses`, `Call.asExprList`.

Added `tests/asParsedType` and  `tests/asParsedType_negative` to check some tricky cases of types that are parsed as expressions and later converted to types.


src/dev/flang/parser/Parser.java
```diff
@@ -1757,45 +1726,9 @@ Actual actualSpace()
          ;
```
michaellilltokiwa, line 1726: grammar rule above seems outdated?

michaellilltokiwa (approved):

src/dev/flang/parser/Parser.java
fridis: Maybe I will change this, left it intentionally since it the OperatorExpr will be turned into a type. Have to think about this.

PR #2957, opened 2024-04-23
dfa/Instance: fix toString for NO_SITE

michaellilltokiwa:


fridis (approved):
Would be good to change the documentation, implementation and all uses that this returns notAvailable also in the case when this currently returns null.

PR #2959, opened 2024-04-24
lib/net: update comment

michaellilltokiwa:
[ci skip]

### DIFF
--- a/lib/net/client.fz
+++ b/lib/net/client.fz
@@ -29,11 +29,11 @@
 #
 # Minimal example:
 # ```
-# rh : net.Request_Handler is
-#   handle_request(chan net.channel) T =>
-#     say chan.get_peer_address
+# rh : net.Request_Handler unit is
+#   handle_request unit =>
+#     say net.channel.env.get_peer_address
 #
-# net.client rh net.family.ipv4 net.protocol.tcp "tokiwa.software" 80
+# net.client rh net.family.ipv4 net.protocol.tcp "tokiwa.software" (u16 80) # NYI: type inference
 # ```
 #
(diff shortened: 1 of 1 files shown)


lib/net/client.fz
```diff
@@ -29,11 +29,11 @@
...
 # ```
-# rh : net.Request_Handler is
-#   handle_request(chan net.channel) T =>
-#     say chan.get_peer_address
+# rh : net.Request_Handler unit is
+#   handle_request unit =>
+#     say net.channel.env.get_peer_address
 #
-# net.client rh net.family.ipv4 net.protocol.tcp "tokiwa.software" 80
+# net.client rh net.family.ipv4 net.protocol.tcp "tokiwa.software" (u16 80) # NYI: type inference
```
fridis, line 36: surprised if this type inference does not work. Please check and add an issue if `u16` is really needed!

fridis (changes requested):

lib/net/client.fz
michaellilltokiwa: Was also surprised but it really does not work:
```

/home/sam/playground/test.fz:6:68: error 1: Incompatible types when passing argument in a call
  net.client rh net.family.ipv4 net.protocol.tcp "tokiwa.software" 80
-------------------------------------------------------------------^^
Actual type for argument #5 'port' does not match expected type.
In call to          : 'net.client'
expected formal type: 'u16'
actual type found   : 'i32'
assignable to       : 'i32'
for value assigned  : '80'
To solve this, you could convert the value using + '.as_u16'.

one error.
```

PR #2960, opened 2024-04-24
lib: add container.Persistent_Map

michaellilltokiwa:
We currently have Map (which is immutable) and Mutable_Map. Persistent_Map is kind of in between those two. It should allow efficient adding of elements while keeping the immutable property of Map.


fridis (approved):
Good, we should add implementations as well!

ps_map was originally intended to be a persistent map, but with the disadvantage that is permits only one modification efficiently, i.e., `s.add e` can be called once on `s` efficiently, but not twice.  I will have to check how this can be made a child of `Persitent_Map`.

PR #2965, opened 2024-04-25
lib/exit: change type to u8

michaellilltokiwa:
depends on previous PR to be merged first

[ci skip]


michaellilltokiwa:
> i think we still need the `0 ≤ code ≤ 127` preconditions because u8 ranges from 0 to 255

Oh :-) you are right, will simplify to `code ≤ 127`

PR #2979, opened 2024-04-29
parser: move parsing of `.env` and `.type` to `callTail`

fridis:
This avoids speculative parsings to distinguish a call `a.b.c` from an env `a.b.env` or a type `a.b.type` and simplifies the parsers.

Since with this change, a call is no longer always an instance of `Call`, added a new rule `pureCall` that returns only a call and is used, e.g., in inherits clauses.

This reduces the time to build java.base.fum by about 10% (from 21s down to 19s on my machine).

### DIFF
--- a/src/dev/flang/ast/Partial.java
+++ b/src/dev/flang/ast/Partial.java
@@ -125,7 +125,7 @@ public class Partial extends AbstractLambda
    *
    * @return the corresponding lambda expression.
    */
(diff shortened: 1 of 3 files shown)


michaellilltokiwa (approved):

michaellilltokiwa:
Cool to see the parser getting simpler

PR #3008, opened 2024-05-03
ast: type inference for arguments for feature with generics

### LINKED ISSUE #3007: type inference for features with generics is broken
```
ex =>
  a(T type, u i64, str T) is
  a 0 "hello"
```
```
/home/sam/playground/test.fz:3:5: error 1: Incompatible types when passing argument in a call
  a 0 "hello"
----^
Actual type for argument #1 'u' does not match expected type.
In call to          : 'ex.a'
expected formal type: 'i64'
actual type found   : 'i32'
assignable to       : 'i32'
for value assigned  : '0'
To solve this, you could convert the value using + '.as_i64'.

one error.
```

michaellilltokiwa:
fixes #3007

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -1806,15 +1806,25 @@ public class Call extends AbstractCall
                   {
                     count++;
                     var actual = resolveTypeForNextActual(pass == 0 ? null : t, aargs, res, outer);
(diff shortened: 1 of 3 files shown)


fridis (approved):
It might be better to check if `t` depends on the generics of the called feature that we currently try to determine, we might need to improve this when type inference becomes more powerful.

PR #3017, opened 2024-05-06
run_tests: add workaround for macOS

michaellilltokiwa:
currently failing since date does not support nanosecond resolution on macOS

### DIFF
--- a/bin/run_tests.sh
+++ b/bin/run_tests.sh
@@ -43,7 +43,18 @@ trap "echo """"; cat ""$BUILD_DIR""/run_tests.results ""$BUILD_DIR""/run_tests.f
 
 echo "$(echo "$TESTS" | wc -l) tests."
 
-START_TIME_TOTAL=$(date +%s%N | cut -b1-13)
+
+# get nanoseconds, with workaround for macOS
+nanosec () {
+  if date --help | grep nanoseconds > /dev/null; then
+    date +%s%N | cut -b1-13
+  else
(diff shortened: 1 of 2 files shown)


bin/run_tests.sh
```diff
@@ -43,7 +43,18 @@ trap "echo """"; cat ""$BUILD_DIR""/run_tests.results ""$BUILD_DIR""/run_tests.f
...
 echo "$(echo "$TESTS" | wc -l) tests."
 
-START_TIME_TOTAL=$(date +%s%N | cut -b1-13)
+
+# get nanoseconds, with workaround for macOS
+nanosec () {
+  if date --help | grep nanoseconds > /dev/null; then
+    date +%s%N | cut -b1-13
+  else
+    date +%s000000 | cut -b1-13
```
fridis, line 52: I think this should be `date +%s000000000` since nanoseconds are 10⁻⁹ s.
Can we reduce the overhead on non-macOS by doing something like
```
  (date +s%N || date +s%000000000) | cut -b1-13
```
instead?

fridis (changes requested):

bin/run_tests.sh
michaellilltokiwa: Ugh, embarrassing conversion mistake... :-) 
We can not do the suggested since date+s%N does not error on macOS. It returns a time with a capital N at the end.

fridis (approved):

PR #3037, opened 2024-05-10
Postcondition inheritance

fridis:
First step to support inheritance of post-conditions. 

For now, post conditions from redefined features are cloned and added to the redefining feature. 

This works as long as the post-conditions are simple enough, in particular they cannot access inner features of the surrounding feature and they cannot use lambda expressions or anything else that causes syntax sugar that results in declaration of inner features.


src/dev/flang/ast/ParsedCall.java
```diff
@@ -125,9 +125,30 @@ public ParsedCall(Expr target, ParsedName name, int select)
...
 
+  /**
+   * When inheriting a post-condition during redefintion, this creates a clone
+   * of the inherited condition.
+   *
+   * @param to the redefining feature that inherits a contract
+   *
+   * @param from the redefined feature this contract should inherit from.
+   */
+  public Expr clonePostCondition(AbstractFeature to, AbstractFeature from)
```
michaellilltokiwa, line 146: Not sure why this is needed? I would expect that ParsedCalls are not present anymore when cloning of post conditions happens?

src/dev/flang/ast/Feature.java
```diff
@@ -911,12 +911,21 @@ public void addResultField(Resolution res)
...
           {
-            protected boolean isResultField() { return true; }
+            public boolean isResultField() { return true; }
           };
       }
   }
 
 
+  /**
+   * Is this a case-field decalred in a match-clause?
```
michaellilltokiwa, line 921: declared

src/dev/flang/util/List.java
```diff
@@ -277,6 +277,39 @@ public String toString(String pre, String sep, String term, Function<T, String>
...
+  /**
+   * add adds given element to the list after possibly cloning the list in case
+   * it was frozen.
+   *
+   * @param e element to add
+   *
+   * @return this if !isFrozen(), a clone of this otherwise, in any case with e
+   * added.
+   */
+  public List<T> addAfterUnfreeze(T e)
```
michaellilltokiwa, line 305: I stumbled here while trying to understand what the purpose of this is. I think this needs an explanation why this is done like this.

src/dev/flang/fe/SourceModule.java
```diff
@@ -1727,6 +1752,34 @@ private void checkDuplicateFeatures(AbstractFeature outer, FeatureName fn, List<
   }
 
 
+  /**
+   *
```
michaellilltokiwa, line 1756: Missing comment?

michaellilltokiwa (changes requested):

src/dev/flang/fe/SourceModule.java
```diff
@@ -1753,10 +1753,15 @@ private void checkDuplicateFeatures(AbstractFeature outer, FeatureName fn, List<
 
 
   /**
+   * Check that code in contract does not access inner features apart from
+   * arguments, result field, outer refs or case fields (in case condition uses
+   * a `match`).  Procude AstErrors if needed.
```
michaellilltokiwa, line 1758: Produce

michaellilltokiwa (approved):

PR #3060, opened 2024-05-15
fe: change semantics of choice inheritance

### LINKED ISSUE #2975: inherit from choice
We may want to loosen the restriction of not being able to inherit from choice:

This currently gives an error.
```
ex is

  some_choice (T type) : choice T nil is
  a : some_choice i32 is
  b : some_choice i32 is
```

Maybe an option and outcome can then have a common ancestor?

michaellilltokiwa:
i.e. this loosens the restriction
- may not inherit from a choice to
- may inherit from one choice only.

fixes #2975

With this patch it is not yet possible to inherit from e.g. `option`. See comment below.


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -490,14 +490,16 @@ else if (this.compareTo(Types.resolved.f_choice) == 0)
...
 
-            if (p.calledFeature().isBaseChoice())
+            if (p.calledFeature().isChoice())
               {
                 if (lastP != null)
                   {
                     AstErrors.repeatedInheritanceOfChoice(p.pos(), lastP.pos());
                   }
                 lastP = p;
+                // NYI: BUG: we need sth. like hand down to figure out the choice generics.
```
michaellilltokiwa, line 500: It is not sufficient anymore to just use actualTypeParameters of `p`. Need to change this

lib/option.fz
```diff
@@ -163,18 +163,18 @@ is
...
-#
-# Using this enables to write
-#
-#   o := option x
-#
-# insted of
-#
-#   o option TypeOfX := x
-#
-public option(T type, o option T) => o
```
fridis, line 180: please do not remove this, `type.from` does not permit type inference  and is more typing even if type inference is not used.

lib/option.fz
```diff
@@ -163,18 +163,18 @@ is
...
+  # option with 1 argument provides an short-hand to wrap a value into an option
+  #
+  # Using this enables to write
+  #
+  #   o := option x
+  #
+  # insted of
+  #
+  #   o option TypeOfX := x
+  #
```
fridis, line 179: comment is incorrect.

lib/outcome.fz
```diff
@@ -190,18 +190,18 @@ is
...
+  #
+  # Using this enables to write
+  #
+  #   o := outcome x
+  #
+  # instead of
+  #
+  #   o outcome TypeOfX := x
+  #
+  public type.from(o outcome T) => o
```
fridis, line 207: dto, please keep the old feature for type inference to work.

fridis (changes requested):
Can you do the API changes in a separate PR?  I like this patch, but not the API changes...

michaellilltokiwa:
> Can you do the API changes in a separate PR? I like this patch, but not the API changes...

ok, I had to add some logic in `lookup0` (last line) such that choices are not returned for normal calls.
```
                    (!v.isField() || !foundFieldInScope) &&
                    !(use instanceof Call c && !c._isInheritanceCall && v.isChoice()))
```

tests/choice_inheritance/skip
```diff
@@ -0,0 +1 @@
+NYI
```
fridis, line 1: why skip this, what does not work yet?

src/dev/flang/ast/AbstractFeature.java
fridis: would be good to have an example that does not work, at least in the comment. Even better would be to fix it :-)...

fridis (changes requested):
Ask for some more comments and documentation why example is skipped.

src/dev/flang/ast/AbstractFeature.java
michaellilltokiwa: @fridis sorry for not being clearer. previously the choice generics where just taken from the inherited choice feature.
e.g.
`option(T type): choice T nil is`

but now we can have e.g.:
`a : option String is`
To figure out the choice generics of a we need to hand down `String` and find the "one" feature that inherits the choice and take those generics.
This will need to be implemented in another PR.

tests/choice_inheritance/skip
michaellilltokiwa: see comment above.

fridis (approved):
nice.

PR #3061, opened 2024-05-15
air: in abstractCalled-check use `isCalled` instead of `isInstantiated`

michaellilltokiwa:
intrinsics may return `String` which will actually be `ConstString` so it would be okay if `String` is instantiad and and abstract is called on it.

### DIFF
--- a/src/dev/flang/air/Clazz.java
+++ b/src/dev/flang/air/Clazz.java
@@ -1985,7 +1985,7 @@ public class Clazz extends ANY implements Comparable<Clazz>
    */
   public void check()
   {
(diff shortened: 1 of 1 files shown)


fridis (commented):
It is unclear to me what this improves. Can you give an example that either causes an unjustified error before which is gone now or an example that did not create an error even though it should but does so now.

michaellilltokiwa:
> It is unclear to me what this improves. Can you give an example that either causes an unjustified error before which is gone now or an example that did not create an error even though it should but does so now.

This is _needed_ for #3062

fridis (approved):

PR #3065, opened 2024-05-16
tests: add a test for wrap-around semantics

maxteufel:


michaellilltokiwa:
Since min and max are abstract features on wrap_around you could define these tests generically:
test (T type: wrap_around) is
  chck (T.max +° 7 = 6) "T.max + 7 = 6"
  ...
test u8
test u16
...
```

But there is no real benefit other than the test being more concise.

michaellilltokiwa (approved):

PR #3068, opened 2024-05-16
parser/ast: use `universe` in type qualifier

### LINKED ISSUE #2682: universe can not be used in full type qualifier
```
ex is

  # some other unit feature
  unit is

  some(T type) is

  some universe.unit
```
```
/home/sam/playground/test.fz:5:3: error 1: Wrong number of type parameters
  some universe.unit
--^^^^
Wrong number of actual type parameters in call:
Called feature: ex.some
expected one generic argument for 'T'
found none.

one error.

```

michaellilltokiwa:
fixes #2682

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -405,7 +405,9 @@ public class Call extends AbstractCall
                AbstractType type)
   {
     if (PRECONDITIONS) require
-      (Errors.any() || generics.stream().allMatch(g -> !g.containsError()));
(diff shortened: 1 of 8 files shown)


fridis (changes requested):
Good, this is needed to have fully qualified types starting in the universe as a counterpart to fully qualified calls. 

It would be nicer if both, calls and types starting with `universe` would be parsed in a similar way, i.e., instead of allowing `universe` as a `ParsedName`, better add a rule `universeType` similar to `universeCall`. 

Then, it should also not be required to have these additional errors for type parameters or `universe` not being used as the first name in a qualified type.

PR #3080, opened 2024-05-17
ast: Do not create error for auto-generated fields in choice

fridis:
This problem occures in fuzion-lang.dev's design/exmaples/red_choice.fz example, were subsequent errors are produced for the `result` field of a choice that contains code.

This patch first changes the order in which errors are produces such that an error for a choice that contains code is produced first. Then, any subsequent errors for artificial fields in a choice are suppressed.


src/dev/flang/ast/Feature.java
```diff
@@ -1591,6 +1582,22 @@ private void checkChoiceAndAddInternalFields(Resolution res)
         }
       }
 
+    res._module.forEachDeclaredOrInheritedFeature(this,
+                                                  p ->
+      {
+        // choice type must not have any fields
+        if (p.isField() && !p.isOuterRef() &&
+            (!Errors.any() || !(p instanceof Feature pf && (pf.isArtificialField() || /* do not report auto-generated fields like `result` in choice if there are other problems */
```
michaellilltokiwa, line 1590: This is hard to read for me. I think we can apply De Morgan here?:
`!(Errors.any() && (p instanceof Feature pf && (pf.isArtificialField() || pf.isResultField()))`

michaellilltokiwa (changes requested):

michaellilltokiwa (approved):

PR #3082, opened 2024-05-19
lib: add prefix and postfix operators for `|x|` instead of `x.abs`

fridis:
Combining prefix and postfix operators permits creation of unary operators that surround their operand. This is an experiment to check if this is useful for the `abs` operation.

If not, we might extend the parser to support something like `prepost | |` operators.


michaellilltokiwa:
I'm quite surprised that this does not seem to break anything! :-)

PR #3090, opened 2024-05-21
lib/mutate: add array2

maxteufel:


michaellilltokiwa:
Good, i wonder though if it would be better to have this be more similar to 1-dim array. I.e. we would than need a two-dimensional buffer...

michaellilltokiwa (approved):

PR #3091, opened 2024-05-21
be/c: enable more warnings -Wextra -Wpedantic etc.

michaellilltokiwa:
had to change some of the code. one of the reasons: apparently casting to a union is a gnu extension and not standard c.


fridis (approved):
I have some bad memories of `-pedantic`, but I forgot what exact problem we used to have. Just in case this starts being too painful, consider removing that option again.

Issue #1875, opened 2023-08-21
FUIR stack should be empty at end of each basic block


fridis:
The FUIR code uses a stack with the constraint that all values within a stack are used at the end of a basic block, values passed between basic blocks are passed via fields.  This constraint, however, is not always met at the moment, which needs to be fixed.

PR #3094, opened 2024-05-21
lib: explicitly ignore results of expressions

michaellilltokiwa:


michaellilltokiwa:
Yes, see this issue #1875. In the future we may want to ignore unused results of e.g. loops implicitly though. Then this would become simpler.

fridis:
> This looks terribly complicated to me, especially in the case of `for` loops and similar. Is there any motivation behind this change?

The motivation is that ignoring a result accidentally is usually a bug that indicates that error handling code is missing or dead code was written. Of course, the APIs and expressions like loops should be designed in a way that this does not create code bloat, e.g. instead of having a mutable map with `put` that returns the old value
```
  _ := m.put k v1
  old_v := m.put k v2
  _ := m.put k v3
```
we should better have two APIs calls, e.g., 
```
  m.put k v1
  old_v := m.replace k v2
  m.put k v3
```

fridis (approved):

PR #3100, opened 2024-05-22
Postcondition features

fridis:
Sorry, this has become a giant PR. The main thing this does is turning post conditions into pure syntax sugar handled by the front end only. A post condition of the form
```
  f
   post
     cc
  =>
    code
```
will be converted into
```
  f =>
    #result := code
    POST_f #result
  POST_f(result) =>
    if !cc then fuzion.runtime.postcondition_fault "cc"
```
So there is a new postcondtion feature that is called by f to check the postcondition.

For redefined features, the postcondition features of the redefined features are called by the postcondition feature of the redefining feature.

Postconditions are now handled by an effect `fuzion.runtime.post_fault` that can be replaced to handle postcondition failure in Fuzion code directly.

A number of additional changes made it into this patch:

- handling of the `result` fields has been simplified and changed, basically this is now always an internal field with name #result, except for within post condition features, where it is an argument called `result`.

- fum files no longer contain post conditions explicitly, but post condition features.

- post conditions may now contain lambdas, inline arrays, etc.

- tests for postconditions have been extended and a test catch_postcondition using different ways to handle a post condition in Fuzion code has been added.


src/dev/flang/fe/SourceModule.java
```diff
@@ -1761,27 +1761,6 @@ private void checkDuplicateFeatures(AbstractFeature outer, FeatureName fn, List<
    */
   private void checkContractAccesses(AbstractFeature f)
```
michaellilltokiwa, line 1762: This method seems to not do anything anymore? Remove or add NYI?

michaellilltokiwa (approved):

src/dev/flang/fe/SourceModule.java
```diff
@@ -1761,27 +1761,6 @@ private void checkDuplicateFeatures(AbstractFeature outer, FeatureName fn, List<
...
   {
```
fridis, line 1763:
```suggestion
  private void checkContractAccesses(AbstractFeature f)
  { //NYI: CLEANUP:  remove, has no effect.
```

PR #3112, opened 2024-05-24
fe: refine lookup0 to consider the scope of definition/usage of a fea…

### LINKED ISSUE #90: feature visibility in nested scopes does not work
A feature declared in a nested scope should not be visible outside that scope, e.g., 
```
 > cat testVisi.fz
testVisi is
  if false
    x => say "Hello"
  x
 > ./build/bin/fz testVisi
Hello
```
This problem causes many test cases not to produce the expected error in tests/visibility_negative.

### LINKED ISSUE #764: possibly uninitialized loop variables can be accessed without causing an error
```
ex_loop_no_error is
  add_ints(elems Sequence i32) i32 is
    for
      elem in elems
      res := elem, res + elem
    else
      res

  add_strings(elems Sequence string) string is
    for
      elem in elems
      res := elem, res + elem
    else
      res

  say (add_ints [1, 2, 3, 4, 5])
  say (add_ints (array 0 i->i))
  say (add_strings ["hello", "world"])
  say (add_strings (array 0 i->" "))
```

in this case adding no i32s returns `-555555555` (`32797392` with C backend) while adding no strings causes a runtime error/segfault.

### LINKED ISSUE #1396: Scoping issue when reusing name in match cases leads to `ambiguous targets found`
```
test_process =>

  target is
    my_call is
  create choice target error is
    error ""

  match create
    p target =>
      p.my_call
    error =>

  match create
    p target =>
      concur.thread.spawn (() ->
        p.my_call)
    error =>
```
```
/home/sam/playground/test.fz:16:9: error 1: Ambiguous targets found for call to 'p'
        p.my_call)
--------^
Found several possible call targets within the current feature at the same outer level in 'test_process' found 'test_process.p' defined at /home/sam/playground/test.fz:9:5:
    p target =>
----^
and in 'test_process' found 'test_process.p' defined at /home/sam/playground/test.fz:14:5:
    p target =>
----^
To solve this, you may rename one of these features.

one error.
```

### LINKED ISSUE #2365: AoC issue: ambiguous targets for invisible feature in else clause
In this nested loop example 
```
for _ := for j in 0.. else _ := j
do
```
The access to 'j' in the else clause should produce an error that `j` is not visible  or accessible here. Instead, we get this:
```
 > ./build/bin/fz test_ambig.fz 

/home/fridi/fuzion/advent_of_code/fuzion_aoc/2023/13/test_ambig.fz:1:33: error 1: Ambiguous targets found for call to 'j'
for _ := for j in 0.. else _ := j
--------------------------------^
Found several possible call targets within the current feature at different levels of outer features:
in '#loop2' found '#loop2.j' defined at /home/fridi/fuzion/advent_of_code/fuzion_aoc/2023/13/test_ambig.fz:1:14:
for _ := for j in 0.. else _ := j
-------------^
and in 'universe' found 'j' defined at /home/fridi/fuzion/advent_of_code/fuzion_aoc/2023/13/test_ambig.fz:1:14:
for _ := for j in 0.. else _ := j
-------------^
To solve this, you may qualify the feature using '#loop2.this.j' or 'universe.this.j'.

one error.
```

michaellilltokiwa:
…ture

fixes #90
fixes #764
fixes #1396
fixes #2365


tests/visibility_negative/visibility_negative.fz
```diff
@@ -126,118 +126,118 @@ visibility_negative is
...
-        chck (f1 == 11 || f1 == 77 || f1 == 143) "can access f1"
+        chck (f1 = 11 || f1 = 77 || f1 = 143) "can access f1"
         use f2   // 41. should flag an error: feature not found / not visible
         use f3   // 42. should flag an error: feature not found / not visible
         use f4   // 43. should flag an error: feature not found / not visible
         use f5   // 44. should flag an error: feature not found / not visible
         use f6   // 45. should flag an error: feature not found / not visible
         use f7   // 46. should flag an error: feature not found / not visible
-        use f8   // 47. should flag an error: feature not found / not visible
+        use f8   // 47. NYI: UNDER DEVELOPMENT: flag an error: feature not found / not visible
```
michaellilltokiwa, line 136: I think i do not agree with these test. `f8` is defined in `a1` and used here in `b1`. Should IMHO be fine even though it defined "later" in the source text than `b1`. I think the DFA needs to find if this can lead to uninitialized access?

src/dev/flang/ast/Block.java
```diff
@@ -358,10 +358,19 @@ public Expr propagateExpectedType(Resolution res, AbstractFeature outer, Abstrac
...
         _expressions.add(new Block(new List<>()));
       }
-    Expr resExpr = removeResultExpression();
+
+    // we must not remove result expression just yet.
+    // we rely on it being present in SourceModule.inScope()
+    var idx = resultExpressionIndex();
+    Expr resExpr = resultExpressionIndex() >= 0
+      ? _expressions.get(idx)
+      : null;
```
fridis, line 367: Can you use `var resExpr = resultExpression()` instead, which is inherted from `AbstractBlock`?

src/dev/flang/ast/Feature.java
```diff
@@ -310,6 +310,15 @@ public Set<AbstractFeature> redefines()
   public boolean _addedLate = false;
 
 
+  /*
+   * true if this feature is found to be
+   * declared in a block with
+   * _newscope=true (e.g. if/else, loop)
+   * or in a case-block
+   */
```
fridis, line 318: I do not fully understand this, imagine this code:
```
f0 =>
  if cc1 then
     f1 => 
       f2 =>
       if cc2 then
         f3 =>
     {
       f4 => 
     }
```
Then I assume  `f1` and `f3` are `_scoped` and `f0` is not. What about `f2` and `f4`?
Would be good to clarify this.

src/dev/flang/ast/Function.java
```diff
@@ -207,14 +207,16 @@ public Expr propagateExpectedType(Resolution res, AbstractFeature outer, Abstrac
...
+    var e = _expr.visit(new FeatureVisitor()
       {
         public Expr action(Call c, AbstractFeature outer)
         {
           return c.updateTarget(res, outer);
         }
       },
       _feature);
+    if (CHECKS) check
+      (e == _expr);
```
fridis, line 219: can you add a comment why `_expr` cannot be modified here?

src/dev/flang/fe/SourceModule.java
```diff
@@ -1195,7 +1226,8 @@ private List<FeatureAndOuter> lookup0(AbstractFeature outer, String name, Expr u
             for (var v : e.getValue())
               {
                 if ((use == null || (hidden != featureVisible(use.pos()._sourceFile, v))) &&
-                    (!v.isField() || !foundFieldInScope))
+                    (!v.isField() || !foundFieldInScope) &&
+                    (use == null || /* NYI: can we restrict this evaluation for outers? */ inScope(use, v)))
```
fridis, line 1230: explain the NYI a bit better.

tests/reg_issue1396/reg_issue1396.fz
```diff
@@ -0,0 +1,40 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test
+#
+# -----------------------------------------------------------------------
+
+reg_issue1396 is
```
fridis, line 24: Can you add some explanation what this tests and what used to be wrong with it?

tests/reg_issue2365/reg_issue2365.fz
```diff
@@ -0,0 +1,27 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test
+#
+# -----------------------------------------------------------------------
+
+reg_issue2365 is
```
fridis, line 24: Can you add some explanation what this tests and what used to be wrong with it?

tests/reg_issue3128/reg_issue3128.fz
```diff
@@ -0,0 +1,29 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test
+#
+# -----------------------------------------------------------------------
+
+reg_issue3128 is
```
fridis, line 24: Can you add some explanation what this tests and what used to be wrong with it?

tests/reg_issue764/reg_issue764.fz
```diff
@@ -40,4 +40,4 @@ reg_issue764 is
   say (add_ints [1, 2, 3, 4, 5])
   say (add_ints (array 0 i->i))
   say (add_strings ["hello", "world"])
-  say (add_strings (array 0 i->" "))
+  say (add_strings (array 0 (i -> $" ")))
```
fridis, line 43: I was confused here, would be good to add a description comment at the beginning of this test (before `reg_issue764 is`) explaining what this does. 

Does this work after your patch: 
```
  count(elemens Sequence T) =>
    for 
       res := 0, res+1
       _ in elems
     else
       res
```
?
This was at least my original intention, even though I am not sure that this is necessarily a good idea.

tests/reg_issue90/reg_issue90.fz
```diff
@@ -0,0 +1,28 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test
+#
+# -----------------------------------------------------------------------
+
+reg_issue90 is
```
fridis, line 24: Can you add some explanation what this tests and what used to be wrong with it?

fridis (changes requested):
Very nice fix! 

Sorry for being very picky and commenting on many minor things, I would like to see this merged soon!

Also, please add issues for the remaining problems once this has been merged.

tests/reg_issue764/reg_issue764.fz
michaellilltokiwa: Yes, above code works after patch.

fridis (approved):

PR #3125, opened 2024-05-24
be/c: add option `-Wno-c23-extensions`

michaellilltokiwa:
fixes some failures when using clang 18 or higher.


fridis:
Not knowing what the exact problem is: I generally prefer fixing the generated code such that it compiles with the latest version of C.

PR #3155, opened 2024-05-29
fe: fix cryptic feature names in err output

### LINKED ISSUE #1771: feature name `_` shown for anonymous features results in cryptic `#_123`
When creating user output like error messages, internal names for anonymous feature and types should be converted back to the original text `_` or some useful human readable text.

Here is one example using `#_0` in the error message:
```
 > echo "f(_ type : String) => { }; f i32" | ./build/bin/fz -

-:1:28: error 1: Incompatible type parameter
f(_ type : String) => { }; f i32
---------------------------^
formal type parameter '#_0' with constraint 'String'
actual type parameter 'i32'

one error.
```

michaellilltokiwa:
fixes #1771

### DIFF
--- a/src/dev/flang/air/Clazz.java
+++ b/src/dev/flang/air/Clazz.java
@@ -1361,7 +1361,7 @@ public class Clazz extends ANY implements Comparable<Clazz>
             ? "ref "
             : ""
             )
-         + feature().featureName().baseName()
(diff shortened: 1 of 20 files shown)


fridis (approved):
Not quite sure if using `result` for expression-result is the right thing to do, but is probably better than `#exprResult123`.

PR #3169, opened 2024-06-03
parser: setMinIndent for `if` that are not part of an `else if`

### LINKED ISSUE #2691: parser: outer `else` falsely matched with inner `if`
This is related to #2371, another surprise: This example
```
for i in 1..10
do
  if i = 2 then say "A: found 2"
else
  say "A: done"
```
works as expected:
```
 > ./build//bin/fz test_loops.fz
A: found 2
A: done
```
while the same code with the `if` moved behind the `do`
```
for i in 1..10
do if i = 2 then say "B: found 2"
else
  say "B: done"
```
executes the `else` inside the loop as part of the `if` statment:
```
 > ./build//bin/fz test_loops2.fz
B: done
B: found 2
B: done
B: done
B: done
B: done
B: done
B: done
B: done
B: done
```
The same problem may occur with nested `if` statements:
```
if true
then if false then
else
  panic "*** unreachable ***"
```
which should not call `panic`, but it does:
```
 > ./build//bin/fz test_if.fz
*** panic: *** unreachable ***
```

michaellilltokiwa:
fixes #2691

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -2794,10 +2794,13 @@ ifexpr      : "if" exprInLine thenPart elseBlock
   {
     return relaxLineAndSpaceLimit(() -> {
         SourcePosition pos = tokenSourcePos();
+        // set new min indent unless this `if` is part of an `else if`
+        var old = lastToken() == Token.t_else ? null : setMinIndent(tokenPos());
         match(Token.t_if, "ifexpr");
         Expr e = exprInLine();
         Block b = thenPart(false);
         var els = elseBlock();
+        if (old != null) { setMinIndent(old); }
         return new If(pos, e, b,
           // do no use empty blocks as else blocks since the source position
           // of those block might be somewhere unexpected.
--- a/lib/container/ps_map.fz
+++ b/lib/container/ps_map.fz
@@ -267,7 +267,8 @@ O(n log n).
         at0 := (at
                 + (if (size & sz0) != 0 nt else 0)
                 + (if (size & sz ) != 0 sz else 0))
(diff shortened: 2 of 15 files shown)


src/dev/flang/parser/Parser.java
```diff
@@ -2794,10 +2794,13 @@ If ifexpr()
   {
     return relaxLineAndSpaceLimit(() -> {
         SourcePosition pos = tokenSourcePos();
+        // set new min indent unless this `if` is part of an `else if`
+        var old = lastToken() == Token.t_else ? null : setMinIndent(tokenPos());
         match(Token.t_if, "ifexpr");
         Expr e = exprInLine();
         Block b = thenPart(false);
         var els = elseBlock();
+        if (old != null) { setMinIndent(old); }
```
fridis, line 2803: This will omit the `setMinIndent` if used in a precondition `pre else` like
```
  f
  pre else if cc then true 
  else false
  is ...
```
Same problem if used in a loop's `else`
```
  while cc1
  else if cc2 then say "a" 
  else say "b"
```
thse should better be required to be
```
  f
  pre else if cc then true 
           else false
  is ...
```
and
```
  while cc1
  else if cc2 then say "a" 
       else say "b"
```
Maybe we need to change the grammar for `elseBlock` to check if fallowed by `if` and if this is the case and the `elseBlock` is used as part of an `if`, only then `setMIntIndent` should be skipped.

fridis (changes requested):
Nice. 
I think I found two situations where this does not work well yet, and a suggestion to fix this in the parser that would not require `Lexer.lastToken()`.

michaellilltokiwa:
superseded

PR #3174, opened 2024-06-04
ast: remove unique anonymous feature id

michaellilltokiwa:
for some unknown reason this behaves differently on windows:
```
RUN reg_issue2402.fz 1c1
< Type of 'String.type.concat#2.#anonymous2'
---
> Type of 'String.type.concat#2.#anonymous10'
```


fridis (commented):
The patch is good and should be merged.
But before that, I would like to understand why there is a difference on Windows. We should have reproducible builds, to there should not be a difference here. 
My guess is that this code from `SourceModule.java` might be the culprit: 
```
                    Files.list(d._dir)
                      .filter(p -> isValidSourceFile(p))
                      .sorted()
                      .forEach(p ->
```
If `sorted()` results in a different order of the files on Windows, we might get different ids for anonymous features. 
Can you check if this is the problem and change it, e.g., by sorting using a platform independent comparator?

Issue #2586, opened 2024-02-21
windows: output of utf-8 not working 


michaellilltokiwa:
On my machine `chcp.com 65001` does the trick. Not sure why this does not apply to github...
```
sam@DESKTOP-CA0927S UCRT64 ~/fuzion
# ./build/bin/fz -modules=java.base tests/javaBase/javaHello.fz
UTF-8
Hello Java ?!
Hello Java ?!
Hello Java ?!
Hello Java ?!
Hello Java ?!
string has 16 bytes: [72, 101, 108, 108, 111, 32, 74, 97, 118, 97, 32, -16, -97, -116, -115, 33]
Hello Java ­ƒîì!
[1, 2, 3]
[instance[Java.java.io.PrintStream]]
[false, true, false, true]
[0, 1, -128, 127]
[0, 1, -32768, 32767]
[0, 1, 0, 65535]
[0, 1, -2147483648, 2147483647]
[0, 1, -9223372036854775808, 9223372036854775807]
[0.0, 1.0, 3.1415927, 3.4028235E38]
[0.0, 1.0, 3.141592653589793, 1.7976931348623157E308]

sam@DESKTOP-CA0927S UCRT64 ~/fuzion
# chcp.com 65001
Aktive Codepage: 65001.

sam@DESKTOP-CA0927S UCRT64 ~/fuzion
# ./build/bin/fz -modules=java.base tests/javaBase/javaHello.fz 
UTF-8
Hello Java 🌍!
Hello Java 🌍!
Hello Java 🌍!
Hello Java 🌍!
Hello Java 🌍!
string has 16 bytes: [72, 101, 108, 108, 111, 32, 74, 97, 118, 97, 32, -16, -97, -116, -115, 33]
Hello Java 🌍!
[1, 2, 3]
[instance[Java.java.io.PrintStream]]
[false, true, false, true]
[0, 1, -128, 127]
[0, 1, -32768, 32767]
[0, 1, 0, 65535]
[0, 1, -2147483648, 2147483647]
[0, 1, -9223372036854775808, 9223372036854775807]
[0.0, 1.0, 3.1415927, 3.4028235E38]
[0.0, 1.0, 3.141592653589793, 1.7976931348623157E308]

sam@DESKTOP-CA0927S UCRT64 ~/fuzion
#
```

michaellilltokiwa:
Problem is not just on github but also my machine. Somehow the world-emoji gets garbled when piping to file.

[...]

PR #3175, opened 2024-06-04
tests: add workaround for #2586

michaellilltokiwa:


michaellilltokiwa:
> This is terrible, but I do not see a better solution at this point.

I agree that this is terrible :-)

PR #3188, opened 2024-06-06
fe: fix reproducibility of order of parsing of source files

michaellilltokiwa:
this is likely the reason the jvm tests fail on windows currently

from the java documentation of `Path.compareTo`:
"The ordering defined by this method is provider specific, and in the case of the default provider, platform specific."


src/dev/flang/fe/SourceModule.java
```diff
@@ -405,7 +405,7 @@ part of the (((inner features))) declarations of the corresponding
 
                     Files.list(d._dir)
                       .filter(p -> isValidSourceFile(p))
-                      .sorted()
+                      .sorted(Comparator.comparing(p -> p.toString(), String.CASE_INSENSITIVE_ORDER))
```
fridis, line 408: Sorry, I am getting a bit picky: This will produce an ambiguous order on systems that are case-sensitive in case there a file names that differ only in case. So maybe better just do
```suggestion
                      .sorted(Comparator.comparing(p -> p.toString()))
```
which I think should work for Windows as well since there will be no file names that differ by case.

fridis (dismissed):
good. 
I have one suggestion for a hypothetical problem and a simpler solution.

src/dev/flang/fe/SourceModule.java
michaellilltokiwa: Yes, that's better..

PR #3202, opened 2024-06-07
lib: remove explicit types for integer constants

simonvonhackewitz:


michaellilltokiwa (approved):

fridis:
> _ No description provided. _

Woulds be good to provide some description, as in `these types are redundant since the values are used as arguments to calls and type propagation will ensure the numeric literals are of these corresponding types`.

Issue #3191, opened 2024-06-07
Some numerical conversion don't check if the value fits into the target type


simonvonhackewitz:
This produces an error as expected
```
a i32 := 1234578
b i8 := a.as_i8
say "$a.as_i8 is $b"
```

but this doesn't
```
a i32 := 1234578
b u8 := a.as_u8
say "$a.as_u8 is $b" # should produce an error but doesn't
```

simonvonhackewitz:
for `as_u8` this is caused by #3051

simonvonhackewitz:
but there are still some preconditions missing, e.g.
```
a i64 := 70000
b u16 := a.as_u16
say "$a.as_u16 is $b"
```
produces the output `70000.as_u16 is 4464`

michaellilltokiwa:
@simonvonhackewitz #3051 is fixed now. Please take a look at this issue again.

simonvonhackewitz:
solved in #4732

PR #3205, opened 2024-06-10
add regression test for #3191

simonvonhackewitz:
test preconditions of numerical conversion


tests/reg_issue3191/skip
```diff
@@ -0,0 +1 @@
+see #3051 (precondition and postcondition inheritence does not work yet)
```
michaellilltokiwa, line 1: This needs a new empty line at before end of file

tests/reg_issue3191/test_preconditions_of_numerical_conversion.fz
```diff
@@ -0,0 +1,36 @@
...
+# -----------------------------------------------------------------------
+
+reg_issue3191 is
+
+  a i32 := 256
+  b u8 := a.as_u8       # should produce an error due to a failed precondition
+  say "$a.as_u8 is $b"
+
+  c i64 := -1
+  d u32 := c.as_u32     # same here
```
michaellilltokiwa, line 31: I think it is better to copy the comment above.
First it is easier to understand, second if the example is extended then the comment might become wrong.

michaellilltokiwa (changes requested):

PR #3244, opened 2024-06-14
tests: remove some NYI that now work

michaellilltokiwa:


tests/valWithDynOuter/valWithDynOuter.fz
```diff
@@ -99,16 +99,6 @@ valWithDynOuter is
     redef a => "t"
     redef x => "$f-$a:t.x"
 
-/* NYI: this does not compile yet
```
michaellilltokiwa, line 102: `redef e` is redefining a constructor. So it is correct that this does not compile

Issue #2386, opened 2023-12-17
AoC issue: Strange wrong-number-of-arguments-in-call errors


fridis:
See this:
```
 > echo 'say 4abc' | ./build/bin/fz -

-:1:1: error 1: Wrong number of actual arguments in call
say 4abc
^^^
Number of actual arguments is 2, while call expects no arguments.
Called feature: 'say'
Formal arguments: 


-:1:6: error 2: Could not find called feature
say 4abc
-----^^^
Feature not found: 'abc' (no arguments)
Target feature: 'universe'
In call: 'abc'

2 errors.
```
There are two issues with this: 
- first, `4abc` apparently gets parsed as two separate arguments, it would be better if the lexer would create just a single token for a numeric literal or, even better, an error token. 
- Then, even if these are parsed as two arguments, the error message says that `say` expects no arguments, while there are two versions of `say`, one with no argument and one with one argument.

PR #3258, opened 2024-06-24
lexer: raise error if num literal is followed by letters.

michaellilltokiwa:
#2386 is actually two issues. This fixes only the first one.

### DIFF
--- a/src/dev/flang/parser/Lexer.java
+++ b/src/dev/flang/parser/Lexer.java
@@ -1790,8 +1790,16 @@ PLUSMINUS   : "+"
             yield new Literal();
           }
       }
-      default -> new Literal(m);
-      };
+      default ->
+      {
+        if (kind(p) == K_LETTER)
+        {
+          Errors.error(sourcePos(),
+                       "Broken numeric literal, expected anything but a letter following a numeric literal.",
+                       null);
+        }
+        yield new Literal(m);
+      }};
   }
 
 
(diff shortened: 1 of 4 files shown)


michaellilltokiwa:
> ```
> $ fz -e 'say 1ABC'
> 
> command line:1:6: error 1: Broken numeric literal, expected anything but a letter following a numeric literal.
> say 1ABC
> -----^
> 
> 
> command line:1:1: error 2: Different count of arguments needed when calling feature
> say 1ABC
> ^^^
> Feature not found: 'say' (2 arguments)
> Target feature: 'universe'
> In call: 'say 1ABC'
> To solve this, you might change the actual number of arguments to match the feature 'say' (one argument) at $MODULE/say.fz:31:8:
> public say(s Any) => io.out.println s
> -------^^^
> 
> 2 errors.
> 
> $ fz -e 'say 1E5ABC'
> 
> command line:1:1: error 1: Different count of arguments needed when calling feature
> say 1E5ABC
> ^^^
> Feature not found: 'say' (2 arguments)
> Target feature: 'universe'
> In call: 'say 1E5ABC'
> To solve this, you might change the actual number of arguments to match the feature 'say' (one argument) at $MODULE/say.fz:31:8:
> public say(s Any) => io.out.println s
> -------^^^
> 
> one error.
> ```
> 
> shouldn't the second one also raise this new error?

I think `1E5ABC` is a numliteral from the lexers point of view.
`1 E 5ABC` with E being exponent and 5ABC being a number in hex code.

PR #3260, opened 2024-06-24
Add precondition features

fridis:
**Front end:** 

Preconditions are now checked by precondition features, e.g., the code for 
```
  feat(a t)
    pre cc
  => code
  r := feat v
```
will be converted into 
```
  feat(a t) 
  => code
  pre_feat(a t) => 
    if !cc then
      fuzion.runtime.pre_fault.cause "cc"
  pre_and_call_feat(a t) =>
    pre_feat a
    feat a
  r := pre_and_call_feat  v
```
Two features are added: `pre_feat` that checks the precondition and `pre_and_call_feat` that combines a call to `pre_feat` followed by a call to `feat`. These are basically used to avoid too complex code at a call site since calling `pre_feat` there would require adding temp fields to hold the arguments. 

Additionally, there are boolean variants of `pre_feat` that do not cause a `pre_fault` but return `true` iff preconditions hold. These are called by `pre_feat` to implement precondition inheritance in case of a redefintion. 

A feature `f` that redefines `g` and `h` will end up with precondition `pre_g || pre_h || pre_f` executed in this order. 

For constructors, the precondition feature will be called from the constructor and not via `pre_and_call...` since there is no redefinition for constructors, but there is inheritance where preconditions must be checked as well.

**Middle End**

Contracts are no longer saved to `.fum`-files, but precondition features, the boolean variant and pre_and_call features are stored instead.

**FUIR**

Contracts are no longer part of the FUIR. We might add some support to extract the preconditions and postconditions from the pre and post features at some point, but for now they are gone. 

**DFA**

The DFA effort exploded with this change. To counter this, the site information that is part of a call context is replaced by the clazz that performs the call, which results, e.g., in `codepoints` created in `character_encodings.unicode.data.fz` to be joined together when created in the same feature. 

**Backends**

Code to handle contracts is gone. 

**Addition to base lib:** 

- added `fuzion.runtime.pre_fault` effect to for precondition failure

**Fixed preconditions in the base lib:** 

- added `u64.fits_in_u8` and `u8.fits_in_u8`, which were missing
- added `f32.fits_in_i64`, which was missing
- added comment and used chained boolean expression in `f64.fits_in_i64`
- changed `wrap_around.infix /!` `wrap_around.infix %!`, `int.infix %!` and `uint.infix %!` to return `other != <..>.zero` instead of `true`.  This is used as precondition for `infix /` and `infix %`, which IMHO cannot work for `other` being zero.
- removed explicit precondition to forbid other from being `zero` in `numeric.infix /` and  `numeric.infix %`. This is allowed, e.g., for floats. 

**This required some bug fixes that are part of this PR:** 

- a `fixed` feature `b.f` that redefines `a.f` where `b` inherits from `a` and `b.f` accesses the outer instance `b.this`, the type of that outer instance is now `b` and not `b.this`. 

**Tests**

- added `tests/catch_precondition` that attempts to exhaustively test all combinations of inheriting two preconditions and redefining them. 
- updated expected error output where needed.

**Documentation**

Added a rule for precondition redefinition to the reference manual.


tests/reg_issue1698/reg_issue1698.fz.expected_err_c
```diff
@@ -1 +1 @@
-*** failed pre-condition on call to '(option error).get'
```
michaellilltokiwa, line 1: The context will be available again, when we have stacktrace support?

src/dev/flang/ast/Impl.java
```diff
@@ -425,6 +425,12 @@ boolean containsOnlyDeclarations()
    */
   public void resolveSyntacticSugar2(Resolution res, AbstractFeature outer)
   {
+    if (outer.isConstructor() && outer.preFeature() != null)
+      { // For constuctors, the constructor itself checks the prcondition (while
```
michaellilltokiwa, line 429: typo: precondition

tests/reg_issue1698/reg_issue1698.fz.expected_err_c
fridis: Yes, two things that I plan here: 
- First, the source code position and text of the failed condition should be shown, i.e., instead of just `:` I would like to see `xyz.fz:line:col: debug: condition xyz`
- Then, there should be a Fuzion-level API to obtain a stack trace

src/dev/flang/ast/Visi.java
```diff
@@ -150,6 +150,25 @@ public boolean definesTypeVisibility()
...
 
+
+  /**
+   * Create a new Visi instance without type visibility. This is used for the
+   * visibility of pre / pre bool / pre and call / post features to avoid
+   * errors due to type visibility.
+   *
+   + @return UNSPECIFIED, PRIV, MOD or PUB.
+   */
+  Visi eraseTypeVisibility()
```
michaellilltokiwa, line 161: This seems basically the same as `featureVisibility`.

michaellilltokiwa: But the name is better than `featureVisibility`

michaellilltokiwa (approved):

PR #3265, opened 2024-06-25
lib: change visibility of Random_Handler

michaellilltokiwa:


fridis (approved):
Good. 

Be more verbose in the description, say something like _This is required for user code to define random effects_.

PR #3269, opened 2024-06-25
lib: add buffered writer

maxteufel:


lib/container/Circular_Buffer.fz
```diff
@@ -86,6 +86,12 @@ is
   => abstract
 
 
+  # discard all unread elements
```
michaellilltokiwa, line 89: Maybe drop this and instead user has to create a new instance of a circular buffer?

lib/io/Write_Provider.fz
```diff
@@ -0,0 +1,30 @@
...
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature Write_Provider
+#
+# -----------------------------------------------------------------------
+
+public Write_Provider ref is
+  public write_byte(b u8) outcome unit
```
michaellilltokiwa, line 25: I think `write_byte` is probably unnecessary.

michaellilltokiwa (changes requested):

lib/io/buffered/writer.fz
```diff
@@ -0,0 +1,109 @@
...
+      unit
+    else
+      n := buffer.buffered # we are expecting to write this number of bytes
+      wp.write (buffer.flush n)
+
+
+  # buffered writing of the given byte
+  #
+  public write_byte (b u8) outcome unit =>
+    if buffer.available < 1
```
michaellilltokiwa, line 67: can this be just `write [b]`?

michaellilltokiwa (approved):

lib/io/buffered/writer.fz
```diff
@@ -0,0 +1,104 @@
...
+
+  # buffered writing of the given byte
+  #
+  public write_byte (b u8) outcome unit =>
+    (write [b]).values.1
+
+
+  # buffered writing of the given array
+  #
+  public write (p array u8) tuple i32 (outcome unit) =>
```
michaellilltokiwa, line 72: I missed this before, I'm confused by the return type. Why not `outcome i32`?

PR #3270, opened 2024-06-25
lib: io.dir, align behaviour of backends and add test

michaellilltokiwa:
Previous to this change the result of io.dir.open.read` depended on the backend.

Merging this PR will break windows tests since `win.c` is missing implementations.


include/posix.c
```diff
@@ -88,17 +88,23 @@ void fzE_opendir(const char *pathname, int64_t * result) {
 char * fzE_readdir(intptr_t * dir) {
   struct dirent * d = readdir((DIR *)dir);
 
-  return (d == NULL) ? NULL : &(*d->d_name);
+  return (d == NULL) ? NULL : strdup(d->d_name);
```
fridis, line 91: The must be a corresponding call to `free` that I do not see in this PR:

michaellilltokiwa: @fridis:
> On success, readdir() returns a pointer to a dirent structure.  (This structure may be statically allocated; do not attempt to free(3) it.)

from `man readdir`

michaellilltokiwa: @fridis should I add this as a comment?

PR #3290, opened 2024-07-01
tests: portable nproc

michaellilltokiwa:
also changed apple to workflows to use run_tests_parallel


michaellilltokiwa:
this seems to reduce test time by more than 10min

PR #3291, opened 2024-07-01
ast: On failure of a pre- / post-condition, show the full failing expression

fridis:
The expression shown was often reduced to something like `:` or `<`, now the whole expression is shown, unless that expression comes from a fum file that currently does not have the full information.

### DIFF
--- a/tests/contracts_negative/SquareRoot.fz.expected_err
+++ b/tests/contracts_negative/SquareRoot.fz.expected_err
@@ -1,5 +1,5 @@
 
-error 1: FATAL FAULT `precondition`: ≥
+error 1: FATAL FAULT `precondition`: a ≥ 0
 Call stack:
 fuzion.type.runtime.type.fault.type.install_default#0.#fun264.call#1: $MODULE/fuzion/runtime/fault.fz:38:18:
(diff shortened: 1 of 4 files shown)


tests/contracts_negative/SquareRoot.fz.expected_err
```diff
@@ -16,9 +16,9 @@ fuzion.runtime.pre_fault.cause#1: $MODULE/eff/fallible.fz:35:6:
 fuzion.runtime.precondition_fault#1: $MODULE/fuzion/runtime/pre_fault.fz:58:52:
 public precondition_fault(msg String) => pre_fault.cause msg
 ---------------------------------------------------^^^^^
-SquareRoot.#pre1_sqrt#1: --CURDIR--/SquareRoot.fz:30:9:
+SquareRoot.#pre1_sqrt#1: --CURDIR--/SquareRoot.fz:30:7:
       a ≥ 0
---------^^^
+------^^^^^^^
```
michaellilltokiwa, line 21: sourcerange seems to be wrong? (too long)

fridis: Created an issue for this: #3292

PR #3304, opened 2024-07-02
lib/process: make it possible to use `io.buffered.writer` for writing

maxteufel:


lib/process/spawn.fz
```diff
@@ -100,6 +118,11 @@ private:public Process(id i64, std_in option i64, std_out, std_err i64) ref is
       .with f
 
 
+  public with_in(T type, LM type : mutate, f ()->T) outcome T =>
```
michaellilltokiwa, line 121: What is the value of passing a custom mutate here?

michaellilltokiwa: You could create a custom local mutate like it is done in other places in the stdlib / tests.
```
lm : mutate is
(io.buffered.writer lm (std_in_write_provider) pipe_buffer_size)
      .with f
```

michaellilltokiwa: Ah okay. Then it would probably be good to have one version using a predefined type like with `open_unique_type` but this does not need to be part of this PR.

tests/process/skip_c
```diff
@@ -0,0 +1 @@
+#559
```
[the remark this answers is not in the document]
michaellilltokiwa: Can you instead put the added test in a new test and put the skip file there?

michaellilltokiwa (approved):

PR #3319, opened 2024-07-05
lib: add base64 and base64url encodings

simonvonhackewitz:
base64 and base64url encodings as defined in [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648)


lib/encodings.fz
```diff
@@ -0,0 +1,28 @@
...
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature character_encodings
+#
+#  Author: Michael Lill (michael.lill@tokiwa.software)
```
michaellilltokiwa, line 22: please remove this line :-)

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
+# Base64 encoding and decoding as defined in RFC 4648
```
michaellilltokiwa, line 1: This needs the GPL header

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+public base64 is
+
+  # allows redefinition for base64url
+  module get_alphabet => "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".as_codepoints.as_array
+  module get_char_62 u8 => 43 # +
```
michaellilltokiwa, line 6: get_char_62 and get_char_63 seems same as just calling get_alphabet[62], get_alphabet[63]

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+  module get_char_63 u8 => 47 # /
+  module get_encoding_name => "base64"
+
+  # NYI: compiler does currently not optimize features without arguments to treat them like fields
+  char_62_ascii := get_char_62
+  char_63_ascii := get_char_63
+  alphabet := get_alphabet
+  encoding_name := get_encoding_name
```
michaellilltokiwa, line 14: no need to add a field here. Just `get_encoding_name` below.

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+  public encode (data Sequence u8) String =>
+
+    # 0 -> 0; 1 -> 2; 2 -> 1
+    pad_size := (3 - data.count%3)%3     # length of the padding: number of padded zero bytes in input as well as number of '=' in output
+    in_len := data.count + pad_size      # length of zero padded input data
+
+    # this guarantees that input length is always multiple of 3
+    input := array u8 in_len i->if i < data.count then data[i] else 0
+
+    for b0 := 0, b0 + 3
```
michaellilltokiwa, line 26: IMHO it is more readable to have just one index variable `b`. Then when using those just add the appropriate amount or in the loop define 
```
b0 => b
b1 => b+1
b2 => b+2
```

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+    else
+      decode data.utf8.as_array
+
+
+  # decodes a sequence of ASCII characters, decoding is strict as required by RFC 4648
+  # non alphabet characters, line breaks, missing padding cause errors
+  public decode(data array u8) =>
+
+    # determine padding length '=' is 61 in ASCII
+    pad_size :=
```
michaellilltokiwa, line 69: Either always use pad size or pad length, not both.

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+      # extract four sextets from 24 bit number
+      n0 => ((n >> 18) & 63).as_i32
+      n1 => ((n >> 12) & 63).as_i32
+      n2 => ((n >> 6) & 63).as_i32
+      n3 => (n & 63).as_i32
+
+      # encode sextets
+      block := alphabet[n0] + alphabet[n1] + alphabet[n2] + alphabet[n3] # NYI: OPTIMIZATION: lots of string concatenation
+    else
+      res.substring 0 (res.byte_length - pad_size) + "=" * pad_size
```
michaellilltokiwa, line 43: I prefer either ("=" * pad_size) or "="*pad_size. This only works because of operator precedence where * is _stronger_ than +, but operator precedence does not really make sense for strings...

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+
+  # decodes a base64 string, decoding is strict as required by RFC 4648
+  # non alphabet characters, line breaks, missing padding cause errors
+  public decode_str(data String) =>
+    if data.byte_length != data.codepoint_length
```
michaellilltokiwa, line 48: This if should be removed and all the logic of detecting bad input should be in `decode`.

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+    if !alphabet_check.ok       # check for non base64 characters
+      alphabet_check.err
+
+    else if !(data.length %% 4) # check input length
+      error "length of input data is not multiple of four, as required by RFC4648"
+
+    else                        # do decoding
+
+      # replace padding with zeros (character 'A' / 65 is decoded to zero)
+      input := ((data.take(data.count - pad_size)) ++ ([(u8 65)].cycle.take pad_size)).as_array
```
michaellilltokiwa, line 114: Can this be done with map `data.map x -> if x=61 then (u8 65) else x).as_array`?

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+
+      _dec_sextet(input, pos) =>
```
michaellilltokiwa, line 116: The underscore at start is not necessary, visibility is private be default

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+        else #if n = get_char_63    # case /
+          u32 63
+
+      for
+        res list u8 := nil, res.concat_sequences [b0, b1, b2] # result: contains the decoded data at the end
+        nxt0 := 0, nxt0 + 4
+        nxt1 := 1, nxt1 + 4
+        nxt2 := 2, nxt2 + 4
+        nxt3 := 3, nxt3 + 4
+      while nxt3 < input.length
```
michaellilltokiwa, line 140: I think you should be able to remove the alphabet_check above and check validity of input in this loop
`error := check_input input[pos+nxt0].bind ()->check_input input[pos+nxt1] ...)`
with check_input returning an `outcome unit`
then in else clause of loop
```
if error.is_error
  error.err
else
     # remove zero bytes caused by padding
     outcome res.take(res.count - pad_size)
```

lib/encodings/base64url.fz
```diff
@@ -0,0 +1,8 @@
+# Base64url encoding and decoding as defined in RFC 4648
```
michaellilltokiwa, line 1: GPL header missing

michaellilltokiwa (changes requested):

lib/encodings/base64.fz
```diff
@@ -0,0 +1,155 @@
...
+    # this guarantees that input length is always multiple of 3
+    input := array u8 in_len i->if i < data.count then data[i] else 0
+
+    for b0 := 0, b0 + 3
+        b1 := 1, b1 + 3
+        b2 := 2, b2 + 3
+        res := "", res + block # the encoded input data
+    while b2 < in_len
+      # get 3 bytes (24 bits) and convert them into a number with 24 bits
+      n u32 := (input[b0].as_u32 << 16) + (input[b1].as_u32 << 8) + input[b2].as_u32
```
fridis, line 32: This must be optimized! `input[b0]` requires time in `O(b0)`, so the loop is in `O(in_len²)`!

My suggestion: Do a loop as follows
```
for 
  res := "", res + block # the encoded input data
  i in 0..
  b in data
  last_n := 0, i %% 3 ?  0 : n
  n := last_n << 8 + b 
do 
  block := if i %% 3 then
              n0 => ..
              ...
              block := alphabet[n0]...
            else
              ""
else
   ...
    
```

fridis (changes requested):
Nice PR!

Just a few comments and suggestions, did not read all of it yet.

lib/encodings.fz
simonvonhackewitz: 🙈

lib/encodings/base64.fz
simonvonhackewitz: you're right, its also less error-prone, having to change things in one place only

lib/encodings/base64.fz
simonvonhackewitz: I would argue that it makes sense in some way. When using the same operators as for numbers it seems intuitive to me that they have the same precedence. Also I've tried to avoid parentheses where possible :)
But I do get your point.

lib/encodings/base64.fz
simonvonhackewitz: The logic is in `decode` already, the reason I've added this, is to generate better error messages.  `decode` works on a byte array, so when a string contains a multi byte code point, the error message will contain this code point and not some cryptic letter resulting from the first byte being converted into a code point.

lib/encodings/base64.fz
simonvonhackewitz: That should actually be possible now, since all the error handling is done before decoding the input

lib/encodings/base64.fz
michaellilltokiwa: Then you should probably add a `decode_internal` that returns either a result or a position of the error in the input.
Then in `decode` and in `decode_str` build an appropriate error message.

lib/encodings/base64.fz
michaellilltokiwa: caching `alphabet` is much more reasonable since it is actually somewhat expensive.

michaellilltokiwa: I guess you should probably define it differently though:
[(codepoint 65), (codepoint 66),...] then the compiler can easily turn this into a compile time constant.

lib/encodings/base64.fz
michaellilltokiwa: Another option (probably slightly better performance) is to already return padding as a string. 
"=", "==", "===" and use padding.byte_length and just `+ padding`

lib/encodings/base64.fz
simonvonhackewitz: I tried this, it gave me a bit of a headache, but it works.

The code is not shorter though, but maybe I made it too complicated.
I don't know if there is a difference in performance (in the old version the input is converted into an array so that should not matter). However, I think it's less understandable, at least for new users.

See c04016f132947328c73c4250da020bf62a05380e

What are your thoughts?

lib/encodings/base64.fz
simonvonhackewitz: I'd prefer not to do this, because then I would once again have to figure out what is the error at position `i` in the input and even do so in two places (`decode` and `decode_str`).

My suggestion would be to always obtain the next `codepoint` when the error happens (performance in the error case should not matter anyways) and include it in the error message like this:
`error: invalid base64 input at byte position 3, decoding to Unicode character '🌏'`

I think this should work for ASCII/byte input too.

This is the code to get, the next codepoint (from the byte array that `encode` receives) when the error occurs 
```
inv_char := String.type.from_bytes (data.drop i)
                            .substring_codepoint 0 1
```

lib/encodings/base64.fz
```diff
@@ -0,0 +1,157 @@
...
+    enc24(n u32) =>
+                  # extract four sextets from 24 bit number
+                  n0 => ((n >> 18) & 63).as_i32
+                  n1 => ((n >> 12) & 63).as_i32
+                  n2 => ((n >>  6) & 63).as_i32
+                  n3 => ( n        & 63).as_i32
+                  [alphabet[n0], alphabet[n1], alphabet[n2], alphabet[n3]]
+    for
+      res Sequence u8 := [], next # the encoded input data
+      i := 0, i+1
```
fridis, line 49:
```suggestion
      i in 0..
```
or
```suggestion
      i in data.indices
```

lib/encodings/base64.fz
```diff
@@ -0,0 +1,157 @@
...
+      last_n u32 := 0, i %% 3 ?  0 : n
+      b in data
+      n := (last_n << 8) + b.as_u32
+      next := if i%3=2 then res ++ enc24 n
+              else           res
+    else
+        case := data.length%3 # last input block of 0, 1 or 2 characters
+        if case = 0      then res
+        else if case = 1 then res ++ (enc24 last_n<<4).slice 2 4 ++ [(u8 61), 61] # "=="
+        else                  res ++ (enc24 last_n<<2).slice 1 4 ++ [(u8 61)]     # "="
```
fridis, line 59:
```suggestion
      case := data.length%3 # last input block of 0, 1 or 2 characters
      if case = 0      then res
      else if case = 1 then res ++ (enc24 last_n<<4).slice 2 4 ++ [(u8 61), 61] # "=="
      else                  res ++ (enc24 last_n<<2).slice 1 4 ++ [(u8 61)]     # "="
```

lib/encodings/base64.fz
```diff
@@ -0,0 +1,157 @@
...
+                  n3 => ( n        & 63).as_i32
+                  [alphabet[n0], alphabet[n1], alphabet[n2], alphabet[n3]]
+    for
+      res Sequence u8 := [], next # the encoded input data
+      i := 0, i+1
+      last_n u32 := 0, i %% 3 ?  0 : n
+      b in data
+      n := (last_n << 8) + b.as_u32
+      next := if i%3=2 then res ++ enc24 n
+              else           res
```
fridis, line 54:
```suggestion
              else          res
```

fridis (changes requested):
just fixes in code I suggested.

lib/encodings/base64.fz
fridis: does not work as suggested.

michaellilltokiwa (approved):

PR #3323, opened 2024-07-05
cleanup: remove unused code

michaellilltokiwa:


src/dev/flang/ast/Contract.java
```diff
@@ -739,7 +739,7 @@ else if (cc != null)
                                      new Block(new_code)));
       }
     code._expressions = new_code;
-    var e = res.resolveType(code, pF);
+    res.resolveType(code, pF);
```
fridis, line 742: The result here should not be ignored, I think I forgot to add a check here, should be
```
    var a = res.resolveType(code, pF); 
    if (CHECKS) check
      (code == e);
```
Currently, `resolveType` for a `Block` will always return the same `Block`, but this might change and then we are in trouble here...

fridis (changes requested):
Nice cleanup. 
Found one bug in the original code where we should at least have a `check`.

fridis (approved):

PR #3325, opened 2024-07-07
lib: move `infix ∈` and `infix ∉` to `property.equatable`

fridis:
Also move `element_of` and `not_element_of` from `String` and `numeric` to `property.equatable`.  This results in `a ∈ s` and `a ∉ s` to work for all `equatable` types, which is required for a `Set` anyway.

### DIFF
--- a/lib/property/equatable.fz
+++ b/lib/property/equatable.fz
@@ -50,3 +50,25 @@ public equatable is
   # is this not part of given set
   #
   not_element_of(s container.Set equatable.this) => !(equatable.this.element_of s)
+
+
+  # is `this` contained in `Set` `s`?
+  #
+  # This should usually be called using type inference as in
+  #
+  #   my_set := set_of ["A","B","C"]
+  #   say ("B" ∈ my_set)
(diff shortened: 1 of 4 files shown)


lib/property/equatable.fz
```diff
@@ -40,3 +40,35 @@ public equatable is
...
+  #   say ("B" ∈ my_set)
+  #   say ("D" ∈ my_set)
+  #
+  public infix ∈ (s container.Set equatable.this) => element_of s
+
+
+  # is `this` not contained in `Set` `s`?
+  #
+  # This should usually be called using type inference as in
+  #
```
michaellilltokiwa, line 69: Would be good to use triple backticks for code sections. When can then use this when generating docs.

simonvonhackewitz (approved):

lib/property/equatable.fz
fridis: Let's use four space indentation in the future for code in comment, that should work:

    this is code

.

PR #3332, opened 2024-07-08
lib/net/server: switch to `io.buffered.writer`

maxteufel:


tests/sockets/sockets_test.fz
```diff
@@ -57,7 +57,11 @@ sockets_test is
 
               res := "received: {rr.as_string}\n\n"
 
-              match chan.write res.utf8
+              w =>
+                _ := (io.buffered.writer lm).write res.utf8.as_array
```
michaellilltokiwa, line 61: I think, better check for error here?

lib/net/channel.fz
```diff
@@ -39,11 +39,6 @@ module:public channel(desc i64) : simple_effect
 is
 
 
-  # write data to channel
```
michaellilltokiwa, line 42: Cool!

michaellilltokiwa (approved):

PR #3334, opened 2024-07-08
makefile: add linter pmd

michaellilltokiwa:
see: https://pmd.github.io/

[ci skip]


michaellilltokiwa:
@maxteufel This was due to discussion with @fridis about a previous cleanup PR where I mostly removed unused variables. The javac linter can not tell you this to my knowledge. Will document this better, so converting to draft...

PR #3335, opened 2024-07-08
lib: Make `panic` and `try` inherit from `eff.fallible`

fridis:
`panic` is the standard `fallible` for quick-and-dirty handling of a severe error (like a programming bug), while `try` is the standard fallible for an expected error (like an I/O error).
    
`option.this` can apparently be used as type parameter in inheritance call of `option`, so remove the NYI-comments and commented code.
    
Also add precondition to `abort` to ensure that effect is intalled.

tests: add tests for `fallible` and calls to `try` and `catch`.

Add references to #3355. This might be improved as discussed in #3356.


michaellilltokiwa:
This patch nicely illustrates nicely how malleable the fuzion language has become. try-catch implemented with just language constructs, nice. :-)

tests/fallible/test_fallible.fz
```diff
@@ -0,0 +1,86 @@
...
+      if 3 < 4 then panic "expected failure"
+    .catch m->
+      say "11: expetcing fault, got fault: {m.0} {m.1}"
+
+# run the tests:
+test_fallible
+
+# our own child for `try` effect.
+#
+# Note that we have to declare this outside of `test_fallible` since creating an
```
michaellilltokiwa, line 82: `type.try_unit` does also not work since we need to use try_unit type? Maybe create an issue for this?

fridis: Added #3388.

PR #3367, opened 2024-07-11
util: use double underline instead of dotted underline

michaellilltokiwa:
Could barely see the underline in my terminal:

with this patch:


michaellilltokiwa:
@maxteufel Did not really consider those, which one would you suggest? Should look "good" in any terminal. This is more difficult than it sounds ;-)

michaellilltokiwa:
curly underline looks good in my terminal as well

PR #3368, opened 2024-07-11
lib: add base32 and base32hex encodings

simonvonhackewitz:
base32 and base32hex encodings as defined in [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648)

### DIFF
--- a/lib/encodings/base32.fz
+++ b/lib/encodings/base32.fz
@@ -1,0 +1,169 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature base32
+#
+# -----------------------------------------------------------------------
+
+# Base32 encoding and decoding as defined in RFC 4648
+#
+public base32 is
+
+  # allows redefinition e.g. for base32hex
+  module get_alphabet => "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567".utf8.as_array
+  module encoding_name => "base32"
+
+  # NYI: compiler does currently not optimize features without arguments to treat them like fields
+  alphabet := get_alphabet
+
+  # decode a valid base32 characters to 5 bits
+  # visibility module to allow redefinition in base32hex
+  module quintet_bits(n u8) =>
+    if n >= 65 && n <= 90        # case A-Z
+      n.as_u64 - 65
+
+    else #if n >= 50 && n <= 55  # case 2-7
+      n.as_u64 - 24
+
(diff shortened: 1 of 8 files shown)


lib/encodings/base32.fz
```diff
@@ -0,0 +1,169 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature base32
+#
+# -----------------------------------------------------------------------
+
+# Base32 encoding and decoding as defined in RFC 4648
```
michaellilltokiwa, line 24: Would be nice to have to url to the RFC here.

lib/encodings/base32.fz
```diff
@@ -0,0 +1,169 @@
...
+public base32 is
+
+  # allows redefinition e.g. for base32hex
+  module get_alphabet => "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567".utf8.as_array
+  module encoding_name => "base32"
+
+  # NYI: compiler does currently not optimize features without arguments to treat them like fields
+  alphabet := get_alphabet
+
+  # decode a valid base32 characters to 5 bits
```
michaellilltokiwa, line 35: characters->character?

lib/encodings/base32.fz
```diff
@@ -0,0 +1,169 @@
...
+  # visibility module to allow redefinition in base32hex
```
michaellilltokiwa, line 36: This comment is not necessary: visibility module to allow redefinition in base32hex.
Also it might be wrong in the future if someone decides to inherit from base32 again.

It might also be a little better design if you mark `quintet_bits` and `is_valid` as abstract and then rename base32 to e.g. rfc4648 and inherit from it twice base32/base32hex

lib/encodings/base32.fz
```diff
@@ -0,0 +1,169 @@
...
+    (c >= 65 && c <= 90) || (c >= 50 && c <= 55)
+
+  # Encodes a given byte sequence in base32, output is padded to multiple of 8
+  # returns a sequence of ascii values
+  public encode(data) =>
+
+    # extract 8 quintets from 40 bit (of an 64 bit integer)
+    enc40(n u64) =>
+      block := array i32 8 (i->((n >> (35-i*5).as_u64) & 31).as_i32)
+      block.map (x->alphabet[x])
```
michaellilltokiwa, line 56: does this work?
```
array u8 8 i->
   idx := (n.as_i32 >> 35-i*5) & 31
   alphabet[idx]
```

lib/encodings/base32.fz
```diff
@@ -0,0 +1,169 @@
...
+
+    for
+      res Sequence u8 := [], next        # the encoded input data
```
michaellilltokiwa, line 59: the encoded input data -> the encoded data

lib/encodings/base32.fz
```diff
@@ -0,0 +1,169 @@
...
+  public decode_str(data String) =>
+    decode data.utf8.as_array
+
+
+  # decodes a sequence of ASCII characters, decoding is strict as required by RFC 4648
+  # lowercase letters, non alphabet characters, line breaks, missing padding cause errors
+  # NYI: decoding does currently not reject encodings where the padding bits have not been set to zero prior to encoding
+  #      therefore in some cases multiple encodings can be decoded to the same data
+  #      See RFC4648 section 3.5: https://datatracker.ietf.org/doc/html/rfc4648#section-3.5
+  public decode(data) =>
```
michaellilltokiwa, line 96: Please add type for arg data since this is a public method.

michaellilltokiwa: I'm not even sure how the compiler manages to infer the type here...

michaellilltokiwa (approved):
Nice work!

lib/encodings/base32.fz
simonvonhackewitz: cast to i32 needs to happen at the end, but aside from that yes 👍️

PR #3387, opened 2024-07-15
ast: raise error on redefining field

### LINKED ISSUE #3108: redefining fields should be forbidden
since it would not be clear in which order the fields should be initialized. (redef of argument fields is fine)
This does not raise an error:
```
  a is
    x := 3

  b : a is
    redef x := 4 # should flag an error
```

michaellilltokiwa:
fixes #3108

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2148,6 +2148,7 @@ public class AstErrors extends ANY
   {
     error(f.pos(),
           "Redefinition of non-argument fields is forbidden.",
+          "The field being redefined: " + existing.pos().show() + System.lineSeparator() +
           "To solve this, you may want to consider converting the redefined field into a routine by replacing" + skw(":=") + " by " + skw("=>") + ".");
   }
 
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2144,6 +2144,12 @@ public class AstErrors extends ANY
        "To solve this, use the result, explicitly ignore the result " + st("_ := <expression>") + " or change " + s(e.type().feature())
               + " from constructor to routine by replacing" + skw("is") + " by " + skw("=>") + ".");
   }
+  public static void redefiningFieldsIsForbidden(AbstractFeature existing, AbstractFeature f)
+  {
+    error(f.pos(),
+          "Redefinition of non-argument fields is forbidden.",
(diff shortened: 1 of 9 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -2144,6 +2144,12 @@ public static void unusedResult(Expr e)
        "To solve this, use the result, explicitly ignore the result " + st("_ := <expression>") + " or change " + s(e.type().feature())
               + " from constructor to routine by replacing" + skw("is") + " by " + skw("=>") + ".");
   }
+  public static void redefiningFieldsIsForbidden(AbstractFeature existing, AbstractFeature f)
+  {
+    error(f.pos(),
+          "Redefinition of non-argument fields is forbidden.",
```
fridis, line 2150: Would be good to include the position `existing.pos().show()` here.

src/dev/flang/fe/SourceModule.java
```diff
@@ -915,6 +915,10 @@ else if (c._hasPost != null && c._hasPostThen == null)
               }
             if (visibleFor(existing, f.outer()))
               {
+                if (existing.isNonArgumentField())
```
fridis, line 918: I think we whould be stricter here: `existing.isNonArgumentField() || f.isNonArgumentField()`. If we redefine an argument field by a field that is not an argument field, we run into strange situations where the same field is initialized twice, once via `existing` in the inheritance call and once va `f` by the field declaration.

tests/reg_issue3108/reg_issue3108.fz
```diff
@@ -0,0 +1,38 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test
+#
+# -----------------------------------------------------------------------
+
+reg_issue3108 =>
+
+  A ref is
```
fridis, line 26: maybe add cases for redefining an inherited argument field or redefining an inherited field or feature by an argument field.

tests/inherit_postcondition/module_src/mod.fz
```diff
@@ -31,7 +31,7 @@ public mod is
   ok(s String) => say s; true
 
   public post_accessing_result is
-    public n := 333
+    public n => 333
```
fridis, line 34: This test checks that the post condition can read a field. Why does this need to be changed?

fridis: I did not see this. I think best would be to have both: The original `n` and a new function `m` that is redefined as a field in `inh_module2`.

Issue #3391, opened 2024-07-15
DFA slow, likely for effects


fridis:
Running `tests/catch_postcondition` with the JVM backend takes 24s on my machine, 93% of this time is spent during DFA. 
![pid951376-fuzion-XjavaProf-flamegraph](https://github.com/user-attachments/assets/809b4bd7-1cf5-4786-b28e-9ec0ba63cb21)

michaellilltokiwa:
This is not the case anymore:
```
~/openvscode-server-fuzion/vscode-fuzion/fuzion (main)130$ make no-java && fz -verbose=3  tests/catch_postcondition/catch_postcondition.fz
make: Nothing to be done for 'no-java'.
498 features 387 types and 1 source files included in fum file.
DFA pre iteration #1: --------------------------------------------------calls:0,values:2,envs:0
DFA pre iteration #2: --------------------------------------------------calls:1835,values:2076,envs:0
DFA pre iteration #3: --------------------------------------------------calls:1925,values:2144,envs:0
DFA pre iteration #4: --------------------------------------------------calls:1972,values:2189,envs:0
DFA pre iteration #5: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #6: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #7: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #8: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #9: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #10: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #11: --------------------------------------------------calls:1977,values:2190,envs:0
DFA pre iteration #12: --------------------------------------------------calls:1977,values:2190,envs:0
DFA real iteration #1: --------------------------------------------------calls:0,values:2190,envs:0
DFA real iteration #2: --------------------------------------------------calls:1715,values:4006,envs:0
DFA real iteration #3: --------------------------------------------------calls:1737,values:4010,envs:0
DFA real iteration #4: --------------------------------------------------calls:1742,values:4010,envs:0
DFA real iteration #5: --------------------------------------------------calls:1742,values:4010,envs:0
DFA needed 17 iterations (pre/real) (12/5).
[...]

PR #3403, opened 2024-07-17
Speed up of DFA Analysis, improve #3391

fridis:
This is a first step towards a faster DFA analysis. Speedup is achieved mostly
by more efficient data structures and in part by creating fewer Calls and Values.

In Deail:

- Added `LongMap` and `IntMap` as efficient data structure for `Call`s and
  `Value`s.

- Equipped `Call`s with a `int` `_uniqueCallId` to avoid comparison and to be
  used as key in `LongMap` or `IntMap`.

- Used `List`, `LongMap` and `IntMap` to cache fequently calculated or looked up
  objects:

  - `Instances` created at a given site
  - tagged values
  - joined values or value sets
  - constant numeric values
  - unknown numeric values
  - array internal data (`SysArray`)
  - `Call`s
  - Effect `Env`ironments
  - default effects
  - EmbeddedValue (used for escape analysis)

  For this, calls like `new SysArray` were replaced by calls `dfa.newSysArray`

- When comparing `Call`s, no longer treat calls with different arguments as
  different. Calls are instead identified by

  - the inner clazz of the call
  - the target value
  - the site
  - the effect environmen

- Add a set of _hot calls_ which are calls that should be re-analyzed since
  their code uses values that have changed in the current iteration. This avoids
  re-running a whole iteration over all calls if a changed value is only
  propagated by one call.

Minor performance improvements

- Replace iterations `value.forAll(t -> f(t))` by the more verbose

      if (value instanceof ValueSet vs)
        {
          for (var t : vs._componentsArray)
            {
              f(t);
            }
        }
      else
        {
          f(value);
        }

- Replaced `TreeSet<Integer>` by `BitSet` for `DFA._writtenFields` and
  `DF._readFields`.

- Increased max depth for recursive DFA analysis on new calls to 40.

- Moved code to print DFA analysis detail for `-verbose=4` our of the main
  iteration loop.

Debugging infrastructure

- Added support for env vars or properties to control DFA

  - boolean `dev.flang.fuir.analysis.dfa.DFA.SHOW_STACK_ON_CHANGE` now also
    usable as env var (using `_` instead of `.`).

  - boolean `dev.flang.fuir.analysis.dfa.DFA.SHOW_CALLS` to show statistics on
    `Call`s created

  - boolean `dev.flang.fuir.analysis.dfa.DFA.SITE_SENSITIVE` to disable call
    site sensitivity, which is much faster, but fails some tests.

  - boolean `dev.flang.fuir.analysis.dfa.DFA.USE_EMBEDDED_VALUES` to disable
    embeddded values resulting in escape analysis not working. This breaks the C
    backend, but is much faster.

Smaller cleanup etc.:

- renamed `DfaErrors.usedEffectNeverInstantiated` as
  `DfaErrors.usedEffectNotInstalled`.

- Added abstract methods `Context.uniqueCallId()` and `Context.env()` to avoid
  the need for instanceof-checks on context values.

- For debugging purposes, added `Call.compareToWhy` in addition to `compareTo`
  to output why a call is considered different to another one.


src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -664,7 +732,52 @@ public Pair<Val, Unit> env(int s, int ecl)
    *   FUZION_JAVA_OPTIONS=-Ddev.flang.fuir.analysis.dfa.DFA.SHOW_STACK_ON_CHANGE=true
    */
   static final boolean SHOW_STACK_ON_CHANGE =
-    Boolean.getBoolean("dev.flang.fuir.analysis.dfa.DFA.SHOW_STACK_ON_CHANGE");
+    FuzionOptions.boolPropertyOrEnv("dev.flang.fuir.analysis.dfa.DFA.SHOW_STACK_ON_CHANGE");
+
+
+  /**
+   * Set this tho show statistics on the clazzes that caused the most Calls that
```
michaellilltokiwa, line 739:
```suggestion
   * Set this to show statistics on the clazzes that caused the most Calls that
```

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -664,7 +732,52 @@ public Pair<Val, Unit> env(int s, int ecl)
...
+   * overhead. However, some tests like fuzion/tests/transducers require this to
+   * work properly.
+   */
+  static final boolean SITE_SENSITIVE = FuzionOptions.boolPropertyOrEnv("dev.flang.fuir.analysis.dfa.DFA.SITE_SENSITIVE", true);
+
+
+  /**
+   * Should the DFA analysis use embedded values?  This is required for proper
+   * escape analysis of instances that contain value types.  Disabling this is
+   * onyl for experimental purposes and will break the C backend since it relies
```
michaellilltokiwa, line 773:
```suggestion
   * only for experimental purposes and will break the C backend since it relies
```

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -2014,31 +2332,279 @@ else if(_fuir.clazzIs(cl, SpecialClazzes.c_bool))
...
+      }
+    else
+      {
+        var cid = _fuir.clazzId2num(nc);
+        var vid = original._id;
+        if (cid >= 0 && cid <= 0xFFFffff &&
+            vid >= 0 && vid <= 0xFFFffff &&
+            tag >= 0 && tag <= 0xff)
+          {
+            var k =
```
michaellilltokiwa, line 2428: These shifts need some comment, imho.

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -2014,31 +2332,279 @@ else if(_fuir.clazzIs(cl, SpecialClazzes.c_bool))
...
+      }
+    return res;
+  }
+
+
+  /**
+   * Create new SysArray instance of the given element values and element clazz.
+   *
+   * NYI: We currently do not distinguish SysArrays by the array instance that
+   * contains.  We problably should do this to make sure that, e.g.,
```
michaellilltokiwa, line 2482:
```suggestion
   * contains it.  We probably should do this to make sure that, e.g.,
```

michaellilltokiwa (approved):

PR #3414, opened 2024-07-19
ast: numliteral, remove type inference to i64

michaellilltokiwa:
If a numliteral did not have a dot and did not fit in i32, i64 was inferred as type. I think this magic does more harm than good. So this PR is proposing to remove this logic.


fridis (approved):
ok. 
I am still unsure if we should go to `i64` as the default integer type anyways. What weighs more, the additional memory use or the inconvenience of not being able to work with huge arrays...

PR #3418, opened 2024-07-19
fe: fix accessing field of other instance in field init

michaellilltokiwa:
this example was broken previous to this PR
```
  ring(r ring) is
    last ring := r.last
```


michaellilltokiwa:
@maxteufel I currently do not see a reason not to. `r.last` is guaranteed to be initialized so no problems there. So, why disallow it?

fridis (approved):
Yes, this should be allowed!

PR #3422, opened 2024-07-22
lib: add qualifiers to pre- and postconditions

### LINKED ISSUE #3410: `pre`- and `post`-condition qualification missing in base lib
I had to make this change to while debugging DFA analysis for effects (I expected my code not to cause any pre/postfault if debug and safety is turned off):
```
> git diff lib/integer.fz 
diff --git a/lib/integer.fz b/lib/integer.fz
index dbd028f75..213b2cd9f 100644
--- a/lib/integer.fz
+++ b/lib/integer.fz
@@ -46,7 +46,7 @@ module:public integer : numeric is
   # bitwise NOT
   public prefix ~ integer.this
     pre
-      is_bounded
+      debug: is_bounded
   => abstract
 
   # bitwise NOT (Unicode alias)
@@ -107,7 +107,8 @@ module:public integer : numeric is
 
     # this digit as an UTF-8 byte
     digit_as_utf8_byte(d u8) u8
-    pre d ≤ 35 # 35 would be a Z
+    pre
+      debug: d ≤ 35 # 35 would be a 'Z'
     =>
       if d < 10 then String.zero_char + d else String.a_char + d - 10
```
Similar qualifiers `debug` and `safety` are missing in other features in the base lib and should be added. As a general guideline: `debug` if the result is undefined or wrong but harmless, `safety` if a crash/segfault/illegal memory access/etc. would be caused in case the condition is not respected.

simonvonhackewitz:
unsure about some of them, e.g.
* int.fz as_uXX
* integer.fz prefix ~
* ...

resolve #3410

### DIFF
--- a/lib/container/Circular_Buffer.fz
+++ b/lib/container/Circular_Buffer.fz
@@ -82,7 +82,7 @@ is
   #
   public flush (n i64) array T
   pre
-    n ≤ buffered
+    safety: n ≤ buffered
   => abstract
 
 
--- a/lib/container/Circular_Buffer.fz
+++ b/lib/container/Circular_Buffer.fz
@@ -82,7 +82,7 @@ is
   #
   public flush (n i64) array T
   pre
-    safety: n ≤ buffered
+    debug: n ≤ buffered
   => abstract
 
 
--- a/lib/container/Set.fz
+++ b/lib/container/Set.fz
@@ -127,8 +127,8 @@ public Set(public E type : property.equatable) ref : Sequence E is
   # equality
   #
   public fixed type.equality(a, b container.Set E) bool
-  pre a.finite
-      b.finite
+  pre safety: a.finite
+      safety: b.finite
   =>
     # NYI: a bit expensive
     (a ∪ b).size_option.val = a.size_option.val
--- a/lib/container/Set.fz
+++ b/lib/container/Set.fz
@@ -127,8 +127,8 @@ public Set(public E type : property.equatable) ref : Sequence E is
   # equality
   #
   public fixed type.equality(a, b container.Set E) bool
-  pre safety: a.finite
-      safety: b.finite
+  pre debug: a.finite
+      debug: b.finite
   =>
     # NYI: a bit expensive
     (a ∪ b).size_option.val = a.size_option.val
--- a/lib/io/buffered/reader.fz
+++ b/lib/io/buffered/reader.fz
@@ -62,7 +62,7 @@ public reader(LM type : mutate, rp Read_Provider, buf_size i32) : effect effect_
   # with up to max_n bytes before returning it.
   #
   public read(max_n i32) array u8 | io.end_of_file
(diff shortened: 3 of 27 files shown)


lib/container/Circular_Buffer.fz
```diff
@@ -82,7 +82,7 @@ is
   #
   public flush (n i64) array T
   pre
-    n ≤ buffered
+    safety: n ≤ buffered
```
michaellilltokiwa, line 85: Also debug

lib/container/Set.fz
```diff
@@ -127,8 +127,8 @@ public Set(public E type : property.equatable) ref : Sequence E is
   # equality
   #
   public fixed type.equality(a, b container.Set E) bool
-  pre a.finite
-      b.finite
+  pre safety: a.finite
```
michaellilltokiwa, line 130: debug

lib/io/buffered/reader.fz
```diff
@@ -76,9 +76,9 @@ public reader(LM type : mutate, rp Read_Provider, buf_size i32) : effect effect_
   # fill the currently empty buffer with up to max_n bytes
   #
   fill_buffer(max_n i32) unit
-  pre match buffer.get
-        a array => a.is_empty
-        io.end_of_file => false
+  pre safety: match buffer.get
```
michaellilltokiwa, line 79: These here also, rather debug than safety, I think?

lib/process/spawn.fz
```diff
@@ -90,7 +90,7 @@ private:public Process(id i64, std_in option i64, std_out, std_err i64) ref is
   # helper function
   #
   write_bytes (desc i64, bytes Sequence u8) outcome i32
-  pre bytes.count > 0
+  pre safety: bytes.count > 0
```
michaellilltokiwa, line 93: debug

michaellilltokiwa (changes requested):

michaellilltokiwa:
@simonvonhackewitz There are now conflicts unfortunately... Can you merge main, please?

lib/io/buffered/reader.fz
fridis: `safety` should IMHO really protect the system from a crash, so I would go for `debug` here. The same for the following 2x`safety`

lib/uint.fz
```diff
@@ -110,7 +110,7 @@ is
   # returns the quotient and the remainder
   # NYI performance: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/math/big/natdiv.go
   fixed divide_with_remainder (divisor uint) tuple uint uint
-  pre divisor > uint.zero
+  pre safety: divisor > uint.zero
```
fridis, line 113: here `safety` is probably justified, the code might run into an recursion or loop if we divide by zero.

fridis (changes requested):
Just minor clarifications on `safety` vs `debug`.

michaellilltokiwa (approved):

PR #3427, opened 2024-07-23
docs: sort features names in api docs case insensitively

simonvonhackewitz:
I'll have to get used to `String` and `Sequence` not being at the top of the list, but generally it's easier to find a name without having to think about capitalization.

### DIFF
--- a/src/dev/flang/tools/docs/Html.java
+++ b/src/dev/flang/tools/docs/Html.java
@@ -238,6 +238,7 @@ public class Html extends ANY
   {
     return set
       .stream()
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
Hmm, I think we want to keep case sensitivity since features started with uppercase are supposed to be refs and lowercase features are values. So quite different. We may want to even separate those in the docs.

simonvonhackewitz:
In the Types category there are currently `Const_String`, `Type`, `FALSE` and `TRUE`, so sorting would still be an advantage there.

michaellilltokiwa (approved):

PR #3439, opened 2024-07-24
docs: improve representation of type features

simonvonhackewitz:
The "type." prefix for type features is now rendered as a keyword. It appears only in front of features which are actual type features from user perspective.

Fixes issue of type appearing twice.

previously
> lock_free.#type.type.map.#type.type.empty

now
> lock_free.map.type.empty

### DIFF
--- a/src/dev/flang/ast/FeatureName.java
+++ b/src/dev/flang/ast/FeatureName.java
@@ -272,7 +272,7 @@ public class FeatureName extends ANY implements Comparable<FeatureName>
       n.startsWith(FuzionConstants.PREANDCALLCONDITION_FEATURE_PREFIX) ? n.replaceFirst(FuzionConstants.PREANDCALLCONDITION_FEATURE_PREFIX + "[0-9]+_", "precall ") :
       n.startsWith(FuzionConstants.PRECONDITION_FEATURE_PREFIX       ) ? n.replaceFirst(FuzionConstants.PRECONDITION_FEATURE_PREFIX        + "[0-9]+_", "pre "    ) :
       n.startsWith(FuzionConstants.POSTCONDITION_FEATURE_PREFIX      ) ? n.replaceFirst(FuzionConstants.POSTCONDITION_FEATURE_PREFIX       + "[0-9]+_", "post "   ) :
-      n.endsWith(FuzionConstants.TYPE_NAME)                            ? n.replace("."+FuzionConstants.TYPE_NAME, "") :
+      n.endsWith(FuzionConstants.TYPE_NAME)                            ? n.replace("." + FuzionConstants.TYPE_NAME, "") :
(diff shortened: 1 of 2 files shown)


src/dev/flang/ast/FeatureName.java
```diff
@@ -272,7 +272,7 @@ public String baseNameHuman()
       n.startsWith(FuzionConstants.PREANDCALLCONDITION_FEATURE_PREFIX) ? n.replaceFirst(FuzionConstants.PREANDCALLCONDITION_FEATURE_PREFIX + "[0-9]+_", "precall ") :
       n.startsWith(FuzionConstants.PRECONDITION_FEATURE_PREFIX       ) ? n.replaceFirst(FuzionConstants.PRECONDITION_FEATURE_PREFIX        + "[0-9]+_", "pre "    ) :
       n.startsWith(FuzionConstants.POSTCONDITION_FEATURE_PREFIX      ) ? n.replaceFirst(FuzionConstants.POSTCONDITION_FEATURE_PREFIX       + "[0-9]+_", "post "   ) :
-      n.endsWith(FuzionConstants.TYPE_NAME)                            ? n.replace(FuzionConstants.TYPE_NAME, "type") :
+      n.endsWith(FuzionConstants.TYPE_NAME)                            ? n.replace("."+FuzionConstants.TYPE_NAME, "") :
```
michaellilltokiwa, line 275:
```suggestion
      n.endsWith("." + FuzionConstants.TYPE_NAME)                            ? n.replace("." + FuzionConstants.TYPE_NAME, "") :
```

src/dev/flang/tools/docs/Html.java
```diff
@@ -182,9 +193,10 @@ private String summary(AbstractFeature af, boolean printArgs)
 
 
   private String anchor(AbstractFeature af) {
-    return "<a class='fd-feature font-weight-600' href='" + featureAbsoluteURL(af) + "'>"
+    return "<span class='font-weight-600'>" + typePrfx(af)
```
michaellilltokiwa, line 196: General rule of thumb, use `div` instead of `span` if not 100% sure that all children are inline elements.

michaellilltokiwa (changes requested):

michaellilltokiwa:
`lock_free.map.type.empty` looks much nicer :-)

src/dev/flang/tools/docs/Html.java
simonvonhackewitz: Hm I thought I tried this and it caused problems, but it works now. I guess I had the type prefix outside the div before.

simonvonhackewitz:
One thing I noticed, in the list of type features, clicking on type does not open the page for the type feature but just unfold like clicking the triangle does. E.g. type.concat on the String page.

Maybe it would be better to move the type inside the link in this case?

michaellilltokiwa (approved):

PR #3448, opened 2024-07-25
ast: cache result of `ResolvedNormalType.isRef()`

fridis:
This reduces the time to run

    > time ./build/bin/fz tests/catch_postcondition/catch_postcondition.fz

by about 1s or 4% on my machine (24.517s instead of 25.472s).

### DIFF
--- a/src/dev/flang/ast/ResolvedNormalType.java
+++ b/src/dev/flang/ast/ResolvedNormalType.java
@@ -91,6 +91,13 @@ public class ResolvedNormalType extends ResolvedType
   AbstractFeature _feature;
 
 
+  /**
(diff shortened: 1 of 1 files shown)


src/dev/flang/ast/ResolvedNormalType.java
```diff
@@ -91,6 +91,13 @@ public class ResolvedNormalType extends ResolvedType
   AbstractFeature _feature;
 
 
+  /**
+   * Cached result of isRef(). Even though this function looks harmless, it is
+   * suprisinging performance critical.
```
michaellilltokiwa, line 96:
```suggestion
   * surprisingly performance critical.
```

michaellilltokiwa (approved):

PR #3451, opened 2024-07-26
docs: include inherited features in the docs

simonvonhackewitz:
Include inherited features in the API documentation. Now all features that a feature provides can be found on one page, inherited features are annotated with their parent.

Current version uses the square brackets (second line)
inherited_example

Any suggestions for the design? Like bigger/smaller font, drop the square brackets, different color...


michaellilltokiwa:
@maxteufel I think features that are redefined are not considered to be inherited, but I think yes, we should probably mark them or add something like `[redefines x]`. (Not in this PR)

michaellilltokiwa (approved):

PR #3463, opened 2024-07-29
be/jvm: always do tail call optimization no matter the lifetime

michaellilltokiwa:


fridis (commented):
This  is correct, but it results in re-allocation of the instance even if it does not escape from a tail recursive feature.

fridis (approved):
Nice!. Looks good!

PR #3464, opened 2024-07-30
parser: remove rule for `dotTypeSuffix`

fridis:
And remove `type_of_lazy`, which is no longer needed.


fridis:
> You forgot to stage the removal of `type_of_lazy`.

That was only in the PR description, unclear to me how it got there.

Issue #3406, opened 2024-07-18
`(#` results in parser crash


fridis:
See this: 
```
 > ./build/bin/fz -e '(#' 

command line:2:1: error 1: Syntax error: expected term (lbrace, lparen, lcrochet, fun, string, integer, old, match, or name), found end-of-file

error 2: java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
	at dev.flang.util.SourceFile.decodeCodePointAndSize(SourceFile.java:307)
	at dev.flang.util.SourceFile.asString(SourceFile.java:849)
	at dev.flang.util.SourcePosition.showInSource(SourcePosition.java:198)
	at dev.flang.util.SourcePosition.show(SourcePosition.java:156)
	at dev.flang.util.Errors.print(Errors.java:473)
	at dev.flang.util.Errors.error(Errors.java:412)
	at dev.flang.util.Errors.syntaxError(Errors.java:441)
	at dev.flang.util.Errors.syntax(Errors.java:796)
	at dev.flang.parser.Lexer.syntaxError(Lexer.java:2377)
	at dev.flang.parser.Parser.term(Parser.java:2115)
	at dev.flang.parser.Parser.opTail(Parser.java:1904)
	at dev.flang.parser.Parser.opExpr(Parser.java:1856)
	at dev.flang.parser.Parser.operatorExpr(Parser.java:1810)
	at dev.flang.parser.Parser.lambda$klammer$8(Parser.java:1935)
	at dev.flang.parser.Lexer.relaxLineAndSpaceLimit(Lexer.java:923)
	at dev.flang.parser.Lexer.bracketTermWithNLs(Lexer.java:979)
	at dev.flang.parser.Parser.klammer(Parser.java:1931)
	at dev.flang.parser.Parser.bracketTerm(Parser.java:1759)
	at dev.flang.parser.Parser.term(Parser.java:2088)
	at dev.flang.parser.Parser.opTail(Parser.java:1904)
	at dev.flang.parser.Parser.opExpr(Parser.java:1856)
	at dev.flang.parser.Parser.operatorExpr(Parser.java:1810)
	at dev.flang.parser.Parser.expr(Parser.java:2654)
	at dev.flang.parser.Parser.exprs(Parser.java:2495)
	at dev.flang.parser.Parser.lambda$block$16(Parser.java:2439)
	at dev.flang.parser.Lexer.optionalBrackets(Lexer.java:939)
[...]

PR #3468, opened 2024-07-30
tests: add regression test for #3406

simonvonhackewitz:


simonvonhackewitz:
test should be fixed with #3474

michaellilltokiwa (approved):

PR #3477, opened 2024-07-31
be/c: initialize all envInstalled to false

michaellilltokiwa:
this only worked by chance, since
stdbool defines:
```
#define true 1
#define false 0
```
and we use memset(..., 0) to initialize complete effect env to all zero


fridis (approved):

fridis:
I actually think this is not by chance and using `memset` should be sufficient, but it does not hurt much to clear these properly.

PR #3496, opened 2024-08-02
be/c: add fzE_memcpy, fzE_memset

michaellilltokiwa:
these should eventually use _secure_ versions of memset, memcpy that the compiler may not optimize away like it is allowed for e.g. `memset`

### DIFF
--- a/include/shared.c
+++ b/include/shared.c
@@ -94,6 +94,16 @@ void * fzE_malloc_safe(size_t size) {
   return p;
 }
 
+void fzE_memset(void *dest, int ch, size_t sz){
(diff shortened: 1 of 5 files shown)


include/shared.c
```diff
@@ -256,8 +266,9 @@ void fzE_destroy_jvm()
 // helper function to replace char `find`
 // by `replace` in string `str`.
 char* fzE_replace_char(const char* str, char find, char replace){
-    char * result = fzE_malloc_safe(strlen(str));
-    strcpy(result, str);
+    size_t len = strlen(str)
```
michaellilltokiwa, line 269:
```suggestion
    size_t len = strlen(str);
```

PR #3497, opened 2024-08-02
be/c: pthread_mutex_unlock, assert we are not unlocking from another …

michaellilltokiwa:
…thread


fridis (changes requested):
Do not put relevant code into an `assert`, it is quite comment to disable assertions by
```
  #define assert(x) ((void)0)
```
instead, use a local var to assign the result and fail if there was a problem

PR #3501, opened 2024-08-02
parser: add semicolon limit for parsing lambda expressions

### LINKED ISSUE #2251: Space separation of arguments does not work if lambda argument ends in a semicolon
This one-liner using a `;` after a lambda
```
m := (1..10).map x->2*x; say m
```
produces an error
```
 > ./build/bin/fz test_parse_lambda.fz

/home/fridi/fuzion/work/test_parse_lambda.fz:1:14: error 1: Could not find called feature
m := (1..10).map x->2*x; say m
-------------^^^
Feature not found: 'map' (3 arguments)
Target feature: 'has_interval.infix ..'
In call: '(1..10).map x->2*x; say m'
To solve this, you might change the actual number of arguments to match the feature 'map' (2 arguments) at $FUZION/lib/has_interval.fz:122:12:
    public map(B type, f Unary B has_interval.this)  => map_sequence f
-----------^

one error.
```
while adding a space before the `;` 
```
m := (1..10).map x->2*x ; say m
```
makes it work
```
 > ./build/bin/fz test_parse_lambda2.fz
[2,4,6,8,10,12,14,16,18,20]
```
Apparently, a `;` causes to terminate the space separation of arguments.

simonvonhackewitz:
For parsing lambda expressions which are not enclosed in parentheses. With this the semicolon will not be parsed as part of the lambda expression.

fix #2251


src/dev/flang/parser/Parser.java
```diff
@@ -2074,14 +2077,14 @@ Expr inlineArray()
                          var s = sep;
                          var p1 = tokenPos();
                          boolean reportedMixed = false;
-                         while ((s == Token.t_comma || s == Token.t_semicolon) && skip(s))
+                         while ((s == Token.t_comma || s == Token.t_semicolon || s == Token.t_semicolonLimit) && skip(s))
```
fridis, line 2080: This should not be needed.

src/dev/flang/parser/Parser.java
```diff
@@ -2074,14 +2077,14 @@ Expr inlineArray()
...
                            {
                              if (current() != Token.t_rcrochet)
                                {
                                  elements.add(operatorExpr());
                                }
                              s = current();
-                             if ((s == Token.t_comma || s == Token.t_semicolon) && s != sep && !reportedMixed)
+                             if ((s == Token.t_comma || s == Token.t_semicolon || s == Token.t_semicolonLimit) && s != sep && !reportedMixed)
```
fridis, line 2087: dito.

fridis (changes requested):
Nice, you seem to have understood well how the parser works. 

However, we should try to avoid specific solutions as much as possible, and your patch introduces special handling in InlineArray that IMHO should not be needed. 

The root problem, as far as I see, is that parsing lambda/block/exprs uses class Indentation, which in `startIndent` calls `endAtSpace(Integer.MAX_VALUE)`. For some reason, this seems to be called after the semicolon in `f ()->x; y`, while this should only happen when finding the indented line `b` in code like
```
  f ()->a
     b
     c
```
So, can you try to find a solution by changing class `Indentation` to disable `endAtSpace` only if we have a new, more deeply indented line?

It would be good if this could be done without `endAtSemicolon`, but that might be impossible. If `endAtSemicolon` is needed, the code you added to relax this in `Lexer.relaxLineAndSpaceLimit` should avoid the special handling in `inlineArray` etc. If it does not, we need to understand why.

Also, if `endAtSemicolon` is needed, please add this fact to the EBNF grammar comment for `Parser.actualSpace()`, which currently reads
```
actualSpace :  operatorExpr         // no white space except enclosed in { }, [ ], or ( ).
```
and should then be
```
actualSpace :  operatorExpr         // no white space or semicolon except enclosed in { }, [ ], or ( ).
```

simonvonhackewitz:
wow that reduced my 35 line commit to one line

fridis (changes requested):
That is nice. 

But please update the comment for the EBNF grammar rule `actualSpace` and in relevant Lexer methods like `endAtSpace`, maybe rename some, e.g. to `endAtSpaceOrSemi`.

fridis (approved):

PR #3503, opened 2024-08-05
parser: end expression at keyword 'then'

### LINKED ISSUE #2370: AoC issue: operator expression followed by lambda is parsed too far resulting in syntax error
This example 
```
if [1] ∃ x->true then
```
produces
```
> ./build/bin/fz test_syntax.fz 

/home/fridi/fuzion/advent_of_code/fuzion_aoc/2023/13/test_syntax.fz:1:18: error 1: Syntax error: expected semicolon ';', found keyword 'then'
if [1] ∃ x->true then
-----------------^
While parsing: semicolon or flat line break, parse stack: semiOrFlatLF, exprs, block, lambda, plainLambda, callOrFeatOrThis, term, opTail (2 times), opExpr, operatorExpr, exprInLine, ifexpr, term, opTail, opExpr, operatorExpr, expr, exprs, unit

one error.
```
while 
```
f(g i32->bool) => true
if f x->true then
```
works
```
 > ./build/bin/fz test_syntax.fz 
```
The reason is that the letter stops parsing arguments at white space while the former does not, but the former should stop at keywords that belong to the outer expression.

simonvonhackewitz:
fix #2370

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -2500,6 +2500,7 @@ brblock     : BRACEL exprs BRACER
         t_rparen,
         t_rcrochet,
         t_until,
+        t_then,
         t_else,
         t_eof -> true;
       default -> isContinuedString(currentNoLimit());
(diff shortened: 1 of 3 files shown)


fridis (approved):
good.

PR #3504, opened 2024-08-05
fuir: add method `isConstructor` to `FUIR`

### LINKED ISSUE #3499: Intrinsic functions are not constructors even though they have no result field.
@michaellilltokiwa is right. Intrinsic functions are not constructors even though they have no result field. 
We should better have `_fuir.isConstructor()` implemented like this! Also for another PR.

_Originally posted by @fridis in https://github.com/tokiwa-software/fuzion/pull/3434#discussion_r1701862400_

simonvonhackewitz:
resolve #3499

### DIFF
--- a/src/dev/flang/fuir/FUIR.java
+++ b/src/dev/flang/fuir/FUIR.java
@@ -2417,6 +2417,19 @@ public class FUIR extends IR
     return result;
   }
 
+  /**
+   * Check if the given clazz is a constructor.
+   *
+   * @param clazz a clazz id
+   *
+   * @return true if the clazz is a constructor, false otherwise
+   */
+  public boolean isConstructor(int clazz)
+  {
+    // Intrinsic functions are not constructors even though they have no result field.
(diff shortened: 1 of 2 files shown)


fridis (changes requested):
This is just the first step. The next step is to search for all source code places that should use this method and use it. E.g., in `be/c/C.java`, there is
```
        var rf = _fuir.clazzResultField(cl);
        l.add(rf != -1 ? current(_fuir.clazzCode(cl)).field(_names.fieldName(rf)).ret()  // a routine, return result field
                       : current(_fuir.clazzCode(cl)).ret()                              // a constructor, return current instance
              );
```
which should use `FUIR.isConstructor` instead of `clazzResultField`. There is similar code in `JVM.java`. 
The main point of encapsulating logic in a function is to re-use it at different places and to avoid bugs in case that logic is wrong (as it was originally for intrinsics).

fridis (approved):

PR #3510, opened 2024-08-08
parser: flag error for ambiguous use of semicolons

### LINKED ISSUE #2702: Surprising difference between lambda and partial application
This results in `f` initialized with a lambda that calls `say` once, followed by a call to `say "ho"` that is executed when this is run
```
 > ./build//bin/fz -e 'f ()->unit :=     say "hi"; say "ho"'
ho
 >
```
while this results in `f` initialized with a lambda that calls `say` twice and no calls to `say` being executed:
```
 > ./build//bin/fz -e 'f ()->unit := ()->say "hi"; say "hi"'
 > 
```
Would be nice to avoid this surprise...

simonvonhackewitz:
fix #2702

Examples for ambiguous use of semicolon
```
my_feat => say "Hi"; say "ambiguous"

if true if true then say "OK"; say "ambiguous"
```

Whats still OK
```
a := 1; say "declared a"

if true then yak "this"; say " is fine";
```


simonvonhackewitz:
NYI: code for the suggested corrections in the error message is still missing

lib/log.fz
```diff
@@ -26,4 +26,4 @@
 # In contrast to `say`, `log` returns its argument
 # which makes `log` compose better in same cases.
 #
-public log(s T) => say s; s
+public log(s T) => {say s; s}
```
michaellilltokiwa, line 29: Why is this semicolon ambiguous?

simonvonhackewitz: Because Fridtjof said so. The second s could be interpreted as part of the feature declaration `public log(s T) => {say s; s}` or something that is executed afterwards `public log(s T) => {say s}; s`

michaellilltokiwa: Hmm, okay. But when are semicolons not ambiguous then? Only when adding braces?

simonvonhackewitz: For feature declarations in one line this is correct.

michaellilltokiwa: Ah, okay. I understand now.

michaellilltokiwa (approved):

src/dev/flang/util/Errors.java
```diff
@@ -827,6 +827,19 @@ public static void trailingWhiteSpaceInMultiLineString(SourcePosition sourcePos)
...
   }
 
+  public static void ambiguousSemicolon(SourcePosition sourcePos)
+  {
+    syntaxError(sourcePos,
+                "Ambiguous semicolon in nested blocks.",
+                "It is unclear whether this semicolon terminates the inner block or not. " +
+                "To solve this, add braces { }.\n");
+
+                // NYI: give examples on how to resolve the ambiguity, like shown in the template below
```
michaellilltokiwa, line 837:
```suggestion
                // NYI: UNDER DEVELOPMENT: give examples on how to resolve the ambiguity, like shown in the template below
```

PR #3514, opened 2024-08-12
io/buffered/reader: switch to circular buffer

maxteufel:


michaellilltokiwa (approved):
Thanks for cleaning up my mess ;-)

lib/io/buffered/reader.fz
```diff
@@ -61,60 +66,66 @@ public reader(LM type : mutate, rp Read_Provider, buf_size i32) : effect effect_
   # in case the buffer is empty it fills the buffer
   # with up to max_n bytes before returning it.
```
michaellilltokiwa, line 67:
```suggestion
  # with up to n bytes before returning it.
```

PR #3522, opened 2024-08-13
lib: move `/!` and `%!` from `wrap_around` to `integer`

michaellilltokiwa:
Maybe I'm missing something but I thought these two are better redefined in integer instead of wrap_around.


fridis (approved):
Probably makes sense. There might be a child of `integer` that provides `∞` as a result of division by `zero`, but that could redefine `/!` as needed.

Issue #3520, opened 2024-08-13
dächle missing at eol


michaellilltokiwa:
```
/home/not_synced/flang_dev (main)1$ cat content/tutorial/examples/square_root9.fz
sqrt(a i32) i32
  pre
    a >= 0
  post
    result * result <= a
    (result + 1) > a / (result + 1)
    result >= 0
=>
  // NYI -- not yet implemented
```
```
/home/not_synced/flang_dev (main)$ FUZION_DISABLE_ANSI_ESCAPES=true fz content/tutorial/examples/square_root9.fz

/home/not_synced/flang_dev/content/tutorial/examples/square_root9.fz:8:3: error 1: Incompatible types in assignment
=>
--
assignment to field : 'sqrt.result'
expected formal type: 'i32'
actual type found   : 'unit'
assignable to       : 'unit'
for value assigned  : 

To solve this, you could change the type of the target 'sqrt.result' to 'unit' or convert the type of the assigned value to 'i32'.

one error.
```
It looked like this before:
```

/home/not_synced/flang_dev/content/tutorial/examples/square_root9.fz:8:3: error 1: Incompatible types in assignment
=>
--^
assignment to field : 'sqrt.result'
expected formal type: 'i32'
actual type found   : 'unit'
assignable to       : 'unit'
for value assigned  : 

To solve this, you could change the type of the target 'sqrt.result' to 'unit' or convert the type of the assigned value to 'i32'.

one error.
```

PR #3529, opened 2024-08-13
fix dächle missing at eol

simonvonhackewitz:
fix #3502


simonvonhackewitz:
Sorry issue number in branch name and commit is wrong. Correct one is:
fix #3520

PR #3530, opened 2024-08-13
util: underline error position at line end when ANSI escapes are enabled

### LINKED ISSUE #3461: Source position in error message is not highlighted in some cases
### With ANSI escapes enabled position at line end is not highlighted
For all errors which only have a position but no range, the position is only highlighted if ANSI escapes are disabled, if I am not mistaken.

Note this example needs change from #3457.

The fuzion file contains no final newline.

```
> cat /tmp/bug.fz 
(#%
```
```
> FUZION_DISABLE_ANSI_ESCAPES=true fz /tmp/bug.fz

/tmp/bug.fz:1:3: error 1: Syntax error: expected term (lbrace, lparen, lcrochet, fun, string, integer, old, match, or name), found end-of-file
(#
--^
While parsing: term, parse stack: term, opTail, opExpr, operatorExpr, klammer, bracketTerm, term, opTail, opExpr, operatorExpr, expr, exprs, block, unit

one error.
```
```
> FUZION_DISABLE_ANSI_ESCAPES=false fz /tmp/bug.fz 

/tmp/bug.fz:1:3: error 1: Syntax error: expected term (lbrace, lparen, lcrochet, fun, string, integer, old, match, or name), found end-of-file
(#

While parsing: term, parse stack: term, opTail, opExpr, operatorExpr, klammer, bracketTerm, term, opTail, opExpr, operatorExpr, expr, exprs, block, unit

one error.
```

### No Position highlighted in error for missing then block
This happens for both enabled and disabled ANSI escapes.
```
> cat /tmp/test.fz
f(x i32) i32 =>
  if x<0 then
  else x
```

```
> FUZION_DISABLE_ANSI_ESCAPES=true fz /tmp/test.fz

/tmp/test.fz:2:14: error 1: Block must end with a result expression
  if x<0 then
-------------
This block must produce a value since its result is used by the enclosing expression.
Expected type of value: 'i32'

one error.
```

simonvonhackewitz:
fix #3461

That's how it looks
image

### DIFF
--- a/src/dev/flang/util/SourcePosition.java
+++ b/src/dev/flang/util/SourcePosition.java
@@ -195,21 +195,23 @@ public class SourcePosition extends ANY implements Comparable<SourcePosition>, H
             if (bytePos() == byteEndPos())
               /* not a SourceRange! */
               {
+                var underlined = _sourceFile.asString(p, Math.min(p+1, _sourceFile.byteLength()-1));
                 sb.append(_sourceFile.asString(_sourceFile.lineStartPos(l), p))
                   .append(Terminal.CURLY_UNDERLINE)
                   .append(Terminal.UNDERLINE_LINE_RED)
-                  .append(_sourceFile.asString(p, Math.min(p+1, _sourceFile.byteLength()-1)))
+                  .append(underlined.length() == 0 && Terminal.ENABLED ? " " : underlined)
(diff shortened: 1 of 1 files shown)


simonvonhackewitz:
good idea

michaellilltokiwa (approved):

PR #3554, opened 2024-08-15
lib/net/server: add experimental feature handling connections in threads

maxteufel:


lib/net/server.fz
```diff
@@ -102,6 +102,27 @@ is
         e
 
 
+  # accept a TCP connection that will be handled in a separate thread
+  #
+  public accept_threaded(ch Connection_Handler unit) outcome unit =>
```
michaellilltokiwa, line 107: I think eventually this should should not use concur.thread.spawn directly, but an some effect that allows executing tasks in the background (thread pool).

michaellilltokiwa (approved):

PR #3564, opened 2024-08-16
lib: option/outcome, change panic messages

michaellilltokiwa:
also removed some unneeded parentheses.


michaellilltokiwa:
> I personally don't like this at all, the new panic messages are completely non-descriptive compared to the old ones.

I think this is okay in this case, since this should be unreachable code. This is unless safety is turned off. Also in the future an uncaught panic may also show a stack trace, which then shows the location of the error.

lib/num_option.fz
```diff
@@ -67,10 +67,10 @@ is
   #
   public val T
     pre
-      safety: (num_option.this??)
+      safety: num_option.this??
   =>
     num_option.this ? v T => v
-                    | nil => fuzion.std.panic "num_option.val called on nil"
+                    | nil => fuzion.std.panic "You either turned off safety or you discovered a severe compiler bug! This code should be unreachable."
```
fridis, line 73: The new message is helpful, but please keep the original cause `num_option.val called on nil` as part of the message. Otherwise it might be hard to debug if this ever shows up (and it will show up, there will be configurations that favor speed over safety). 
Same for the other messages.

fridis (changes requested):
see comment in code, please leave the original message as part of the error.

lib/num_option.fz
michaellilltokiwa: @fridis: okay, but we should also really improve our tooling to make this easy to debug.

lib/num_option.fz
```diff
@@ -67,10 +67,10 @@ is
   #
   public val T
     pre
-      safety: (num_option.this??)
+      safety: num_option.this??
   =>
     num_option.this ? v T => v
-                    | nil => fuzion.std.panic "num_option.val called on nil"
+                    | nil => fuzion.std.panic "num_option.val called on nil. You either turned off safety or you discovered a severe compiler bug! This code should be unreachable."
```
fridis, line 73: I would not ever speculate we have a compiler bug, developers would do this anyway. My suggestion for this and the other messages is something like:

```suggestion
                    | nil => fuzion.std.panic "num_option.val called on nil. Enable `safety` to obtain a precondition failure for debugging."
```

fridis (changes requested):

lib/num_option.fz
michaellilltokiwa: @fridis Yes, that makes more sense.

fridis (approved):

PR #3570, opened 2024-08-16
ast: improve error message when inheritance is used without `ref`

### LINKED ISSUE #3563: Improve "incompatible types" error message when inheritance with missing `ref` keyword is used
For people coming from other languages it can be easy to forget that in fuzion the `ref` keyword is necessary to use references and have inheritance behave like it does e.g. in Java.

Therefore it would be helpful, if in this case, the use of the `ref` keyword would be suggested in the error message.

**Example**
```
> cat /tmp/impr_err_msg.fz
parent is
  func String => abstract

child : parent is
  redef func => "not working because ref keyword missing"


main(x parent) =>
  say (x.func)

main child

> FUZION_DISABLE_ANSI_ESCAPES=true fz /tmp/impr_err_msg.fz

/tmp/impr_err_msg.fz:11:6: error 1: Incompatible types when passing argument in a call
main child
-----^^^^^
Actual type for argument #1 'x' does not match expected type.
In call to          : 'main'
expected formal type: 'parent'
actual type found   : 'child'
assignable to       : 'child'
for value assigned  : 'child'
To solve this, you could change the type of the target 'x' to 'child' or convert the type of the assigned value to 'parent'.

one error.
```

simonvonhackewitz:
fix #3563

How it looks now
```
> cat /tmp/impr_err_msg.fz
parent is
  func String => abstract

child : parent is
  redef func => "not working because ref keyword missing"


main(x parent) =>
  say (x.func)

main child
> FUZION_DISABLE_ANSI_ESCAPES=true fz /tmp/impr_err_msg.fz

/tmp/impr_err_msg.fz:11:6: error 1: Incompatible types when passing argument in a call
main child
-----^^^^^
Actual type for argument #1 'x' does not match expected type.
In call to          : 'main'
expected formal type: 'parent'
actual type found   : 'child'
assignable to       : 'child'
for value assigned  : 'child'
To solve this you could:
  • make  'parent' a reference by adding the 'ref' keyword, so all its heirs can be used in place of it
  • change the type of the target 'x' to 'child'
  • convert the type of the assigned value to 'parent'.

one error.

```

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -424,7 +424,12 @@ public class AstErrors extends ANY
           }
         else
           {
-            remedy = "To solve this, you could change the type of the target " + ss(target) + " to " + s(actlT) + " or convert the type of the assigned value to " + s(frmlT) + ".\n";
+            remedy = !frmlT.isRef() && actlT.feature().inheritsFrom(frmlT.feature()) ?
+                        "To solve this you could:\n" + //
+                            "  • make  " + s(frmlT) + " a reference by adding the " + st("ref")+ " keyword, so all its heirs can be used in place of it\n" + //
+                            "  • change the type of the target " + ss(target) + " to " + s(actlT) + "\n" + //
+                            "  • convert the type of the assigned value to " + s(frmlT) + "."
+                        : "To solve this, you could change the type of the target " + ss(target) + " to " + s(actlT) + " or convert the type of the assigned value to " + s(frmlT) + ".\n";
           }
         actlFound   = "actual type found   : " + s(actlT);
         valAssigned = "for value assigned  : " + s(value) + "\n";
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -424,7 +424,7 @@ public class AstErrors extends ANY
           }
         else
           {
-            remedy = !frmlT.isRef() && actlT.feature().inheritsFrom(frmlT.feature()) ?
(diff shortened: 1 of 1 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -424,7 +424,12 @@ else if (frmlT.compareTo(Types.resolved.t_unit) == 0)
           }
         else
           {
-            remedy = "To solve this, you could change the type of the target " + ss(target) + " to " + s(actlT) + " or convert the type of the assigned value to " + s(frmlT) + ".\n";
+            remedy = !frmlT.isRef() && actlT.feature().inheritsFrom(frmlT.feature()) ?
+                        "To solve this you could:\n" + //
+                            "  • make  " + s(frmlT) + " a reference by adding the " + st("ref")+ " keyword, so all its heirs can be used in place of it\n" + //
+                            "  • change the type of the target " + ss(target) + " to " + s(actlT) + "\n" + //
+                            "  • convert the type of the assigned value to " + s(frmlT) + "."
```
fridis, line 431: It is not clear if you have to do all three of these things or pick just one. Add `or` and commas for clarity:
```suggestion
                            "  • make  " + s(frmlT) + " a reference by adding the " + st("ref")+ " keyword, so all its heirs can be used in place of it,\n" + //
                            "  • change the type of the target " + ss(target) + " to " + s(actlT) + ", or\n" + //
                            "  • convert the type of the assigned value to " + s(frmlT) + "."
```

michaellilltokiwa (approved):

PR #3581, opened 2024-08-19
ast: suppress irrelevant errors

### LINKED ISSUE #3541: Irrelevant error messages `Different count of arguments` in `tests/semicolon_ambiguity`
This test's output contains errors like these
```
--CURDIR--/semicolon_ambiguity.fz:35:34: error 8: Different count of arguments needed when calling feature
  if true then if true then else say "inner"; say "ambiguous"
---------------------------------^^^
Feature not found: 'say' (2 arguments)
Target feature: 'semicolon_ambiguity'
In call: 'say "inner"'
To solve this, you might change the actual number of arguments to match the feature 'say' (one argument) at $MODULE/say.fz:31:8:
public say(s Any) => io.out.println s
-------^^^
```
These should not occur, the parser apparently creates an AST with a `Call` to `say` with an extra argument, which is not there in the sources.

simonvonhackewitz:
suppress irrelevant errors if previous error occured

fix  #3541


michaellilltokiwa (approved):
One line fix, nice! :-)

PR #3587, opened 2024-08-20
lib: use new effects syntax for internationalization

lisaw0:
this makes using the internationalization effect easier


lib/internationalization/int_handler.fz
```diff
@@ -25,3 +25,6 @@
 #
 public int_handler ref is
     public date(d i32,m i32,y i32) String => abstract
+
+
+#unnecessary
```
michaellilltokiwa, line 30: left accidentally?

fridis (approved):

PR #3601, opened 2024-08-20
tests: uncomment test cases in `visibility_negative` which are working by now

simonvonhackewitz:


michaellilltokiwa (changes requested):
I think a lot of those are not fixed (yet). But test needs to be updated since every `use ...` line produces an error: "Expression produces result of type 'visibility_negative.this.visi2.this.use' but result is not used."

PR #3607, opened 2024-08-21
C: When `-XkeepGeneratedCode=on` is set, do not use `File.createTempFile`

fridis:
I was confused then using `-verbose` and `-XkeepGeneratedCode=on` and I could not find the C files anymore.

With this patch, `-XkeepGeneratedCode=on` will cause not to use temp files at all but generate C files using the main clazz name.


michaellilltokiwa:
@fridis I was annoyed by this in the past but haven't come around to fixing it...

PR #3616, opened 2024-08-21
lib: replace int_eff with provide

lisaw0:
to simplify using internationalization effect, provide is used instead of int_eff


simonvonhackewitz:
I think you forgot to include `provide.fz` in the commit.

fridis (approved):

PR #3618, opened 2024-08-21
ast: improve error message when redefining hidden feature

### LINKED ISSUE #3312: Improve error message when redefining hidden feature
By default features from other files are not visible, if visibility is not set explicitly, so trying to redefine such a feature results in this error 
`Feature declared using modifier 'redef' does not redefine another feature`

This can be a bit unintuitive, if not familiar with visibility in Fuzion.

simonvonhackewitz:
fix #3312

example
```
FUZION_DISABLE_ANSI_ESCAPES=true fz -sourceDirs=/tmp/ /tmp/3312_child.fz

/tmp/3312_child.fz:2:9: error 1: Feature declared using modifier 'redef' does not redefine another feature
  redef get => "I'm child"
--------^^^
Redefining feature: 'child.get'
To solve this, check spelling and argument count against the feature you want to redefine and make sure the feature to be redefined is visible where it is redefined or remove 'redef' modifier in the declaration of 'child.get'.

one error.

```

### DIFF
--- a/src/dev/flang/fe/SourceModule.java
+++ b/src/dev/flang/fe/SourceModule.java
@@ -40,6 +40,7 @@ import java.util.SortedMap;
 import java.util.Stack;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import dev.flang.ast.AbstractAssign;
 import dev.flang.ast.AbstractBlock;
@@ -945,7 +946,9 @@ A post-condition of a feature that redefines one or several inherited features m
 A feature that does not redefine an inherited feature must not use the `redef` modifier.
     // end::fuzion_rule_PARS_NO_REDEF[]
             */
-            AstErrors.redefineModifierDoesNotRedefine(f);
+            List<FeatureAndOuter> hiddenFeaturesSameSignature = lookup(outer, f.featureName().baseName(), null, true, true);
+            hiddenFeaturesSameSignature = hiddenFeaturesSameSignature.stream().filter(fan->fan._feature.featureName().equals(f.featureName())).collect(List.collector());
+            AstErrors.redefineModifierDoesNotRedefine(f, hiddenFeaturesSameSignature);
           }
         else if (c._hasPreElse != null)
           {
--- a/src/dev/flang/fe/SourceModule.java
+++ b/src/dev/flang/fe/SourceModule.java
@@ -946,8 +946,8 @@ A post-condition of a feature that redefines one or several inherited features m
 A feature that does not redefine an inherited feature must not use the `redef` modifier.
     // end::fuzion_rule_PARS_NO_REDEF[]
             */
-            List<FeatureAndOuter> hiddenFeaturesSameSignature = lookup(outer, f.featureName().baseName(), null, true, true);
-            hiddenFeaturesSameSignature = hiddenFeaturesSameSignature.stream().filter(fan->fan._feature.featureName().equals(f.featureName())).collect(List.collector());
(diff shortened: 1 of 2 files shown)


michaellilltokiwa (changes requested):
We should probably go one step further and check if there actually is a hidden feature we are trying to redefine.

simonvonhackewitz:
It now looks like this
```
> cat /tmp/3312_parent.fz
public parent is
  get => "I'm parent"

> cat /tmp/3312_child.fz
child : parent is
  redef get => "I'm child"

say child.get

> FUZION_DISABLE_ANSI_ESCAPES=true fz -sourceDirs=/tmp/ /tmp/3312_child.fz

/tmp/3312_child.fz:2:9: error 1: Feature declared using modifier 'redef' does not redefine another feature
  redef get => "I'm child"
--------^^^
Redefining feature: 'child.get'
To solve this, check spelling and argument count against the feature you want to redefine or remove 'redef' modifier in the declaration of 'child.get'.
Also make sure that the feature to be redefined is visible where it is redefined. There is the feature 'parent.get' that could be made public to allow redefinition in 'child'.

one error.
```
````
> cat /tmp/3312_parent.fz
public parent is
  get(a, b, c i32) => "Feature with wrong signature but correct name"
  get(a i32) => "Another feature with wrong signature but correct name"

> cat /tmp/3312_child.fz
child : parent is
  redef get => "I'm child"

say child.get

> FUZION_DISABLE_ANSI_ESCAPES=true fz -sourceDirs=/tmp/ /tmp/3312_child.fz

/tmp/3312_child.fz:2:9: error 1: Feature declared using modifier 'redef' does not redefine another feature
  redef get => "I'm child"
--------^^^
Redefining feature: 'child.get'
To solve this, check spelling and argument count against the feature you want to redefine or remove 'redef' modifier in the declaration of 'child.get'.

one error.
```

src/dev/flang/fe/SourceModule.java
```diff
@@ -945,7 +946,9 @@ else if (c._hasPost != null && c._hasPostThen == null)
 A feature that does not redefine an inherited feature must not use the `redef` modifier.
     // end::fuzion_rule_PARS_NO_REDEF[]
             */
-            AstErrors.redefineModifierDoesNotRedefine(f);
+            List<FeatureAndOuter> hiddenFeaturesSameSignature = lookup(outer, f.featureName().baseName(), null, true, true);
+            hiddenFeaturesSameSignature = hiddenFeaturesSameSignature.stream().filter(fan->fan._feature.featureName().equals(f.featureName())).collect(List.collector());
```
michaellilltokiwa, line 950: Is there a reason you did not use `FeatureAndOuter.filter`?

simonvonhackewitz: I found it easier. I did not fully understand all the arguments `FeatureAndOuter.filter` requires and also it produces an error if there is more than one match.

michaellilltokiwa: @simonvonhackewitz okay, makes sense.

michaellilltokiwa, line 950:
```suggestion
            List<FeatureAndOuter> hiddenFeaturesSameSignature = lookup(outer, f.featureName().baseName(), null, true, true)
                                   .stream().filter(fan->fan._feature.featureName().equals(f.featureName())).collect(List.collector());
```

michaellilltokiwa: Instead of `fan` I think in other place `fo` (singular) is used or `fos` (plural).

michaellilltokiwa (changes requested):

Issue #3199, opened 2024-06-07
Detection of call to abstract no longer works


fridis:
This examle 
```
test_abs is
  X ref is
    f i32 => abstract

  y : X is

  x X := y
  say x.f
```
should produce an error `error 1: Used abstract feature 'test_abs.X.f' is not implemented by 'test_abs.y' during `air` phase, but this is no longer the case but it fails awfully during DFA:
```
 > ./build/bin/fz test_abs.fz

/home/fridi/fuzion/work/test_abs.fz:8:9: error 1: NYI: in test_abs(0 args) at 805306375 no targets for Call test_abs.X.f(outer test_abs.X) i32 target boxed(test_abs.ref y):test_abs.y@/home/fridi/fuzion/work/test_abs.fz:7:10:
  say x.f
--------^
Considered targets: 

program entry point
    |
    +- performs call test_abs => *** VOID *** ENV: 'NO ENV'
program entry point
    |
    +- performs call test_abs => *** VOID *** ENV: 'NO ENV'
program entry point
    |
    +- performs call test_abs => *** VOID *** ENV: 'NO ENV'
one error.

```
Problem is #3061, which changes a `isInstantiated` into  `isCalled`. The reason is the  `isInstantiated` is set for, e.,g.,  `String` by #3062. 

The  problem is: `String.isInstantiated` should never be `true`, there are no instances of `String`, only instances of children of `String`. I fear we have to go back to using `intrinsic_constructor`, because we need to distinguish an intrinsic that returns a `ref` to something that is not instantiated by the intrinsic like a `String` and intrinsics that actually instantiate the result type.

PR #3624, opened 2024-08-22
air: treat all intrinsics as intrinsic constructors

michaellilltokiwa:
i.e. mark their results as instantiated

### DIFF
--- a/src/dev/flang/air/Clazz.java
+++ b/src/dev/flang/air/Clazz.java
@@ -90,41 +90,8 @@ public class Clazz extends ANY implements Comparable<Clazz>
   public static FeatureLookup _flu;
 
 
-  /**
-   * Set of intrinsics that implicitly create an instance of their reference
-   * result value.
-   */
-  static TreeSet<String> _intrinsicConstructors_ = new TreeSet<String>();
-  static
-  {
-    _intrinsicConstructors_.add("fuzion.java.get_static_field0"     );
-    _intrinsicConstructors_.add("fuzion.java.get_field0"            );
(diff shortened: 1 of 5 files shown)


fridis (commented):
I fear we cannot do this, see my comment at the end of #3199.

michaellilltokiwa:
@fridis, please review again:

this example:
```
/home/not_synced/fuzion (air--treat-all-intrinsics-as-intrinsic-constructors)1$ cat ~/playground/test.fz
ex is
  say (type_of Any).name.utf8
  say String.utf8
```
correctly reports an error:
```
/home/not_synced/fuzion (air--treat-all-intrinsics-as-intrinsic-constructors)$ make base-only && fz ~/playground/test.fz 

$MODULE/String.fz:28:8: error 1: Used abstract feature 'String.utf8' is not implemented by 'String'
public String ref : property.equatable, property.hashable, property.orderable is

Feature 'String' instantiated at /home/sam/playground/test.fz:3:7:
  say String.utf8

inherits or declares abstract feature 'String.utf8' declared at $MODULE/String.fz:34:10:
  public utf8 Sequence u8 => abstract

which is called at /home/sam/playground/test.fz:2:26:
  say (type_of Any).name.utf8

without providing an implementation

call 'ex'
program entry point

one error.
```
while calling only the intrinsic that returns a String works:
```
/home/not_synced/fuzion (air--treat-all-intrinsics-as-intrinsic-constructors)1$ fz -e 'say (type_of Any).name.utf8'
[65, 110, 121]
```

fridis (approved):

PR #3625, opened 2024-08-22
lib: add `concur.Futur` and thread.submit that returns a `Future`

michaellilltokiwa:
example code:
```
  # sum using futures, where computation takes some time
  test_3 =>
    sum := concur.thread_pool 3 ()->

      (1..49)
        .map (i -> concur.thread_pool.env.submit i32 (() -> time.nano.sleep (time.durations.ms 100); i))
        .map x->x.val.get
        .fold i32.sum

    say sum
  test_3
```
results in 1225


simonvonhackewitz:
From a user perspective, I think, it would be nice if `submit` returned a feature instead of an option. This way unwrapping the option would not be necessary for every `and_then` (#3703) call.

So this
```
fut := (concur.thread_pool.env.submit ...
         .val.and_then ...
         .val.and_then ...
```
would become this
```
fut := (concur.thread_pool.env.submit ...
         .and_then ...
         .and_then ...
```

Then the error handling would need to be done in some other way, e.g. with preconditions or by using `panic`. I'm not sure what disadvantages this might have.

lib/concur/thread_pool.fz
```diff
@@ -118,6 +118,46 @@ is
     bq.enqueue task
 
 
+  public submit(T type, task ()->T) option (concur.Future T) =>
```
fridis, line 121: I agree with Simon. One way to achieve this would be to add a `pre !bq.is_closed` and, since `shutdown` may be done concurrently, a `panic` in case `bq.enqueue` returns `false`.

Also, please add a comment explaining the behavior in  these corner cases.

fridis (changes requested):

lib/concur/thread_pool.fz
michaellilltokiwa: If I add `pre !bq.is_closed` I have to change visibility of bq. Another option might be to introduce an `is_shutdown` feature in `thread_pool` but that is for another PR.

PR #3641, opened 2024-08-26
ast: fix precondition failure, AbstractType.asString

michaellilltokiwa:
Problem was the feature() was called on parametric type.


fridis:
Would be nice to now how this was provoked. Can you add a test?

fridis (approved):

PR #3669, opened 2024-08-29
 ast: Remove `Expr.typeForCallTarget`

fridis:
This is based on #3667 which should be merged first. 

This was only used in a tricky way for `NumLiteral` call targets, now part of `typeForInferencing`.


src/dev/flang/ast/NumLiteral.java
```diff
@@ -435,7 +439,9 @@ AbstractType typeForInferencing()
   @Override
   AbstractType typeForUnion()
   {
-    return _type;
+    // NYI: This seems to work even if we always return null here.  Need to
```
michaellilltokiwa, line 442: forgotten label on this NYI

michaellilltokiwa (approved):

PR #3671, opened 2024-08-29
ast: In `Call`, relax target type lookup to allow for partial application, fix #3658

### LINKED ISSUE #3658: Backward pipe with partial application on left argument no longer works
Forward pipe works as expected
```
 > ./build/bin/fz -e "s := [1,2,3] |> sum; say s"
6
```
backward does not
```
 > ./build/bin/fz -e "s := sum <| [1,2,3]; say s"

command line:1:6: error 1: Different count of arguments needed when calling feature
s := sum <| [1,2,3]; say s

Feature not found: 'sum' (no arguments)
Target feature: 'universe'
In call: 'sum'
To solve this, you might change the actual number of arguments to match the feature 'sum' (2 arguments) at $MODULE/sum.fz:34:8:
public sum(T type : numeric, l Sequence T) => l.fold T.sum


one error.
```
using a field without the need for partial application, however, does work
```
 > ./build/bin/fz -e "a array i32 -> i32 := sum; s := a <| [1,2,3]; say s"
6
```

fridis:
Fix partial application to call target did not work, eg. the code
```
  s := sum <| [1,2,3];
```
used to produce an error since `sum` was not automatically converted into `x -> sum x`.

To fix this, the lookup of the type of the target in a call must be lazy using Expr.typeForInferencing() instead of Expr.type().

This patch uses Expr.typeForInferencing() in Call.targetTypeOrConstraint() and returns Types.t_UNDEFINED in case there is some problem obtaining this. This is then further propagated as a null result for Call.targetFeature().

Furthermore, calls to reportPendingError were removed, they are not performed later during checkTypes and `Call.type()`.

Since `Expr.type()` may now be used to report errors, I had to add an implementation to `AbstractBlock` to avoid fallback to the default in `Expr` that produces a different error.

Simplified type inference between calls and feature results, in particular by

- clarifying the result values of `AbstractFeature`'s `resutType` and `resultTypeIfPresent`.

- removed `AbstractFeature.resultTypeIfPresent(res, generics)` and `AbstractFeature.resultTypeForTypeInference(pos, res, generics)`.

- added `AbstractFeature.resultTypeIfPresentUrgent` used only in `Feature` to force an `Expr` to report an error if a call cannot be resolved but its result type is needed.  In this case, we want the call's error to be shown and not a message that a feature's type cannot be inferred.

- moved calls to `reportedPendingError` in `Call` out of `typeForInferencing` and instead to `type` such that it does not happen before partial application on a call' s target

- restructuring of type parameter inference in a call

- Moved inference of recursive call results and cyclic type inference from `Feature` to `Call`

- Added `urgent` mode to `Impl.inferredTye` to produce error in case the type cannot be inferred from the result / initial value.

This patch also adds a regression test for #3658.

Some error output had to be updated due to different order of errors and minor changes in error output.


tests/reg_issue2365/reg_issue2365.fz.expected_err
```diff
@@ -3,7 +3,7 @@
   for _ := for j in 0.. else _ := j
 ----------------------------------^
 Feature not found: 'j' (no arguments)
-Target feature: 'reg_issue2365.loop.else'
+Target feature: 'reg_issue2365.else'
```
michaellilltokiwa, line 6: Why is the output different here? I think it should still be `loop.else`?

michaellilltokiwa (approved):

tests/reg_issue2365/reg_issue2365.fz.expected_err
fridis: I was a bit surprised here as well. The code for the `else` part is duplicated, so we have different `else`'s depending on whether the loop body is executed zero times or at least once. My assumption is that this error is now found first in the version when the loop body is never executed, but I did not verify that my assumption is correct.

michaellilltokiwa: @fridis sounds plausible

PR #3680, opened 2024-08-30
lib: cleanup resources on abort of effect

michaellilltokiwa:
[ci skip]

### DIFF
--- a/lib/container/Stack.fz
+++ b/lib/container/Stack.fz
@@ -29,6 +29,9 @@ public Stack(T type) ref is
   public pop option T => abstract
   public peek option T => abstract
 
+  public is_empty =>
+    peek.is_nil
+
 
 
 # A stack implementation using local mutation
--- a/lib/effect.fz
+++ b/lib/effect.fz
@@ -77,7 +77,10 @@ public effect is
   # determine the required effects and create code to  instate them at the beginning
   # of an application.
   #
-  public type.default(e effect.this) => default0 e
+  public type.default(e effect.this)
+    pre !e.abortable
+  =>
+    default0 e
 
 
   # convenience version of `type.default` for effect values whose type is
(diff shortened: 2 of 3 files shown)


fridis (commented):
This is an important feature that is still missing. However, my idea how this should be provided is a bit different to have `instate` wrap code into an `open_resources` effect.

What we need is a feature in `effect` like `effect.finally` that is called whenever the effect is popped from the effect stack. This must be called when 
- the effect's `instate0` intrinsic returns normally
- the effect's `instate0` intrinsic returns due to an abort
- the effect is popped from the effect stack due to an abort of a surrounding effect
With such a cleanup callback in place, effects can then provide operations to register something like what you call a `finalizer`.

This will require support from all the backends and from DFA/CFG analysis. 

The `finally`-mechanism in `effect` should be generic. One use would be an `auto_closer` (need to find a better name) effect that woudl maintain a list of resources that need to be closed. But there are probably other uses like logging termination etc.

michaellilltokiwa:
I'm struggling to understand each approaches pros and cons. Currently we do not have a means to
> determine when effect is popped from the effect stack due to an abort of a surrounding effect

but I'm not sure we really want / need this? We would need to modify every backend to support popping effect from effect stack. My approach does not need modification of backends but maybe we cannot get the same performance as with first approach? 
Tangentially related thought:
```
if e.abortable
      lm.instate_self ()->
        (open_resources (container.stack lm ()->unit)).instate_self ()->
          instate0 e cf
          match cf.res
            nil =>
              open_resources.env.abort
              def()
            x R => x
```
It may make sense to have `abortable` to default to false and also have abortable determined by the compiler?

fridis:
> but I'm not sure we really want / need this?

I think we really need this. A simple example: Some code uses a file that is closed by the surrounding effect. Not that code performs an `exit 1` at some point, which currently jumps directly to the abort handler of the surrounding `exit` handler, without caring about the file that needs to be closed.

PR #3712, opened 2024-09-04
ast: Types, remove unused `f_Any`

michaellilltokiwa:


fridis (approved):
surprised that this is unused, we might have to re-add it at some point, but good to remove it for now.

PR #3713, opened 2024-09-04
lib: add knuth morris pratt search algorithm

### LINKED ISSUE #3711: Optimized `Sequence.find` using Knuth-Morris-Pratt algorithm
Currently, our implementation of `Sequence.find` may end up requiring quadratic runtime. We should use the KMP Algorithm, a racket version is presented [here](https://www.cambridge.org/core/journals/journal-of-functional-programming/article/knuthmorrispratt-illustrated/8EFA77D663D585B68630E372BCE1EBA4). 
A test case that requires quadratic time is `("a"*2*n).find "a"*n+"b"`.

michaellilltokiwa:
fixes #3711

code ported from paper:
https://www.cambridge.org/core/services/aop-cambridge-core/content/view/8EFA77D663D585B68630E372BCE1EBA4/S0956796824000017a.pdf/knuth-morris-pratt-illustrated.pdf


fridis (changes requested):
That was quick :-).
Can you add a test case that shows the bad performance of the original implementation like `("a"*2*n).find "a"*n+"b"` with `n` selected such that the test still finishes within a few seconds using the interpreter. How much faster is that compared to the old code?

michaellilltokiwa:
@fridis code was actually broken. I misinterpreted the delay from the racket code. It is not our `Lazy` but a cached execution that I now introduced the feature `once` for.

tests/lib_knuth_morris_pratt/lib_knuth_morris_pratt.fz
```diff
@@ -0,0 +1,73 @@
...
+    len := 250
+
+
+    # for len=250 and interpreter takes ~70s on my machine
+    brute_force := time.stopwatch (()-> _:= test_brute_force len)
+    # for len=250 and interpreter takes ~1s on my machine
+    kmp := time.stopwatch (()-> _:= test_kmp len)
+
+    # check that kmp is at least four times faster
+    check (kmp*4 < brute_force)
```
fridis, line 69: You did much more than what I expected, I did not think of automating this benchmark. 
This is a nice benchmark, however, this will occasionally fail due to external workload hitting us while running the kmp benchmark, so this will result in unstable test runs. 
Better run only the kmp version, maybe with a larger `len`. The benchmark code could be disabled using, eg., `if false` or `if (envir.args.contains "-benchmark")??` or similar.

lib/once.fz
```diff
@@ -0,0 +1,36 @@
...
+#  Source code of Fuzion standard library feature once
+#
+# -----------------------------------------------------------------------
+
+# executes `f` only the first time when
+# calling `get` caching its result.
+#
+public once(T type, f Lazy T) is
+
+  cache := mut (option T) nil
```
fridis, line 29: This should use a local mutate, i.e., `LM type : mutate` argument and `LM.new (option T) nil`

fridis (changes requested):

tests/lib_knuth_morris_pratt/lib_knuth_morris_pratt.fz
michaellilltokiwa: Probably better yes, even though kmp is a lot faster than just the stated four times.

PR #3217, opened 2024-06-11
fuir: remove manual stack cleanup

michaellilltokiwa:
fixes #1875
fixes #2345 
fixes #3521

PR #3732, opened 2024-09-06
lib: introduce `switch`, parent of choices with success and failure

michaellilltokiwa:
- outcome(public T type) : switch T error,
- public option(public T type) : switch T nil,
- public num_option(public T type : numeric) : switch T nil,

moved exists, ok, postfix ??, postfix !!, prefix !, get, get#1, val, val#1 and unwrap to switch.

### DIFF
--- a/lib/num_option.fz
+++ b/lib/num_option.fz
@@ -30,54 +30,9 @@
 # numeric result or false for a boolean result.
 #
 public num_option(public T type : numeric)
-  : choice T nil,
+  : switch T nil,
     monad T (num_option T)
 is
-
-  # Does this option contain a value of type T?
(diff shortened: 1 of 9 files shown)


michaellilltokiwa:
@maxteufel Sorry, my bad. Yes they are related. 
1) (Auto)-unwrap did not work if the auto_unwrap was inherited indirectly, hence the `allInherited` that includes indirectly inherited feautures.
2) The `if (!clazzHasUnitValue(rt))` is in this manual stack cleanup code that #3217 will change again. The `drop` of a unit value triggered a check condition, in the jvm-backend. So I added this workaround.

PR #3736, opened 2024-09-06
jvm: Use special value instead of `null` for unit type effect to fix #3734

### LINKED ISSUE #3734: `effect.abort` results in `As per data flow analysis this code should be unreachable.`
This
```
e : effect is
  stop => e.type.abort e

codea1 => "hi"
codea2 => e.env.stop
defa => "ho"

codeb1 => 42
codeb2 => e.env.stop
defb => 666

say (e.instate String e codea1 defa)
say (e.instate String e codea2 defa)
say (e.instate i32 e codeb1 defb)
say (e.instate i32 e codeb2 defb)
```
produces
```
 > ./build/bin/fz test_abort.fz
hi

error 1: As per data flow analysis this code should be unreachable.
Call stack:
e.type.pre abort
e.type.precall abort
e.stop
codea2
λ.call
(e.type.instate#4 String).λ.call
(e.Effect_Call String).call
e.type.instate#4 String
universe

one error.
```

fridis:
`null` was also used to represent not installed effect, which caused the bug.

### DIFF
--- a/src/dev/flang/be/jvm/Intrinsix.java
+++ b/src/dev/flang/be/jvm/Intrinsix.java
@@ -715,7 +715,9 @@ public class Intrinsix extends ANY implements ClassFileConstants
           var arg = args.get(0);
           if (jvm._types.resultType(ecl) == ClassFileConstants.PrimitiveType.type_void)
             {
-              arg = arg.drop().andThen(Expr.ACONST_NULL);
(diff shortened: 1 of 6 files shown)


michaellilltokiwa:
Does this also fix #3352?

fridis:
> Does this also fix #3352?

No since #3352 is a C backend problem. But I have the impression that it has disappeared meanwhile.

michaellilltokiwa (approved):

Issue #3714, opened 2024-09-04
replace all "negative" tests with "simple and negative" tests, one by one


simonvonhackewitz:
There are unintended errors  in negative tests which are ignored because they don't happen in a line which is supposed to produce an error, see #3684.

I tried to check the total error count, but that does not work, because sometimes multiple lines which are marked to produce an error are reported in one error (e.g.  in `tests/abstractfeatures_negative` 12 lines are commented to produce errors in the source code but the error count when running it is four).

With "simple and negative" tests, there would still be the possibility to accidentally rerecord new errors, but at least there would be a chance to notice them. And it would still protect against accidental rerecording of missing expected errors

fridis:
probably yes. But we should do this maybe one test at a time and verify carefully if the errors are really what was intended by the test. I would like to avoid having errors in simple-and-negative tests that are actually wrong, better disable some tests and create issues if that is the case.

PR #3759, opened 2024-09-10
tests: replace negative tests with simple_and_negative tests (batch 1)

simonvonhackewitz:
part of #3714

fix broken tests


tests/argumentcount_negative/argumentcount_negative.fz
```diff
@@ -23,38 +23,38 @@
 #
 # -----------------------------------------------------------------------
 
-argumentcount_negative is
+private:public argumentcount_negative is
```
michaellilltokiwa, line 26: Why private:public?

michaellilltokiwa (approved):
Nice

simonvonhackewitz:
> why did you change the indentation of the `should flag an error` comments in some of the files?

So they match the indentation of the other lines with these comments. I find that makes it easier to see if all lines should cause the same error or if there are some different ones.

tests/argumentcount_negative/argumentcount_negative.fz
simonvonhackewitz: I think there was some error that it redefines a feature with more restrictive visibility. But now that I look at it again, I think it might not be necessary?

michaellilltokiwa: Or just use public

simonvonhackewitz: seems to no longer be necessary

PR #3762, opened 2024-09-10
tests: replace negative tests with simple_and_negative tests (batch 2)

simonvonhackewitz:
part of #3714

fix broken tests


simonvonhackewitz:
test test_asParsedType_negative fails because there is not outer feature in the test, i.e. `test_asParsedType_negative is` at the beginning is missing. Adding this alone is not the solution, because then some expected errors are missing.

michaellilltokiwa (approved):

michaellilltokiwa (approved):

PR #3768, opened 2024-09-11
lib: add `and_then` to `concur.Future`

### LINKED ISSUE #3703: lib: `concur.Future` `and_then`
Building on #3625 we could implement an `and_then` method on future. This takes a lambda from "result of future" to "new result" and returns a new future, scheduled on the same thread pool as the first future.

simonvonhackewitz:
close #3703

### DIFF
--- a/lib/concur/Future.fz
+++ b/lib/concur/Future.fz
@@ -35,3 +35,7 @@ public Future(T type) ref is
   public get T => abstract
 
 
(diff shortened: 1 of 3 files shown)


simonvonhackewitz:
> test needs to be rerecorded

🙈

michaellilltokiwa (approved):

PR #3785, opened 2024-09-13
ast: avoid duplicate error with source position 'built-in'

### LINKED ISSUE #3749: choice type with overlapping type parameters produces unnecessary second error with missing line numbers
```
fz -e "A : choice i32 i32 is"                                                                                                                                 ✔  24s  

command line:1:1: error 1: Actual type parameters to choice type must be disjoint types
A : choice i32 i32 is
^
The following types have overlapping values:
'i32'
'i32'


<built-in>: error 2: Actual type parameters to choice type must be disjoint types
The following types have overlapping values:
'i32'
'i32'

2 errors.

```

simonvonhackewitz:
fix #3749


michaellilltokiwa:
good, but please add regression test

michaellilltokiwa (approved):

PR #3804, opened 2024-09-18
docs: don't show parents of function features

### LINKED ISSUE #3802: API doc should not show parents like `: Any` for functions
See this
image
the parent features are only relevant for constructors and choice features.

simonvonhackewitz:
resolve #3802


src/dev/flang/tools/docs/Html.java
```diff
@@ -96,7 +96,7 @@ public Html(DocsOptions config, Map<AbstractFeature, Map<Kind,TreeSet<AbstractFe
    */
   private String inherited(AbstractFeature af)
   {
-    if (af.inherits().isEmpty())
+    if (af.inherits().isEmpty() || (Util.Kind.classify(af) == Util.Kind.Other)) // don't show inheritance for function features
```
fridis, line 99: `Other` seems to include `Choice`, which should have the inheritance shown. 
I think it would be better to use `AbstractFeature.Kind` or `IR.FeatureKind` and not have a new `Util.Kind` which will run out out sync with the `fz` command.

simonvonhackewitz: requires #3810

fridis: I do not see how your reply is related to my comment.

simonvonhackewitz: I will use the solution from #3810 here once it is merged. It was partly a reminder to myself why I have not resolved this yet.

fridis (approved):

fridis (approved):

PR #3810, opened 2024-09-18
docs: fix wrongly displayed 'is' where it should be '=>'

simonvonhackewitz:
Type features and intrinsics were falsely displayed with `is` and without return type.


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -396,6 +396,12 @@ public static Kind from(int ordinal)
   public boolean isTypeParameter() { return switch (kind()) { case TypeParameter, OpenTypeParameter -> true; default -> false; }; }
   public boolean isOpenTypeParameter() { return kind() == Kind.OpenTypeParameter; }
 
+  /**
+   * Does this feature has an arrow "=>" in it's signature, i.e. is a function or an intrinsic
+   * @return true if the signature contains an arrow "=>"
+   */
+  public boolean signatureWithArrow() {return (isRoutine() && !isConstructor()) || isIntrinsic();}
```
fridis, line 403: Ho about native and abstract? Are they displayed using `=>`?

simonvonhackewitz: No, good point, should have thought of that.

fridis (approved):

PR #3811, opened 2024-09-18
Frontend produces librarymodule

fridis:
Always create a fum file for the middle end to work on. The AST is thrown away.
This required the following changes
- special handling for universe which is part of every fum file 
- adding of a new `MirExprKind` for code in fum files: `Stop`. This is added after a `Match` if that match only has branches that result in `void`. This happens for code like `f => if true then f else do`
- `Types.Resolved` now has to be re-created
- error output of some tests had to be updated since the position stored for a Call in the fum file is the source range of that call. 

I measured the impact this has on the time to run `fz -e 'say "Hi!"' which is not measurable (best time out of 10 runs: 1.55s now vs. 1.585s before).


src/dev/flang/fe/LibraryOut.java
```diff
@@ -840,6 +844,7 @@ else if (e instanceof AbstractMatch m)
    */
         var cs = m.cases();
```
michaellilltokiwa, line 845: Why not `m.type().isVoid()`?

michaellilltokiwa (approved):

src/dev/flang/fe/LibraryOut.java
fridis: try this please, would be good to get rid of `Stop` by just checking the result type.

PR #3814, opened 2024-09-19
docs: show names of generic types relative to the feature they are used in

simonvonhackewitz:
`cons(A type, B type, head A, tail B) : Cons A, B is`

instead of
`cons(A type, B type, head cons.A, tail cons.B) : Cons cons.A, cons.B is`


fridis (commented):
Looks good. 
However, I fear there might be cases that are not handled correctly. Can you try to generate the doc from this code: 
```
test_generics is

  a(X, Y type,
    v X,
    w Y,
    o option X,
    a Sequence Y) is

  b(X, Y type,
    v (b X Y).X,
    w (b X Y).Y,
    o option   (b X Y).X,
    a Sequence (b X Y).Y) is

  c(X, Y type,
    v (test_generics.c X Y).X,
    w (test_generics.c X Y).Y,
    o option   (test_generics.c X Y).X,
    a Sequence (test_generics.c X Y).Y) is

  d(X, Y type,
    v (d u64 bool).X,
    w (d f32 u8  ).Y,
    o option   (d u64 bool).X,
    a Sequence (d f32 u8).Y) is

  e(X, Y type,
    v (test_generics.e TRUE FALSE).X,
    w (test_generics.e unit void ).Y,
    o option   (test_generics.e TRUE FALSE).X,
    a Sequence (test_generics.e unit void ).Y) is
```
What does it look like for `b`, `c`, `d` and `e`? These cases also do not work for `Fz` yet, so I have created #3817.

simonvonhackewitz:
image

fridis:
> image

I think in case `a` at least we should see `option X`and `Sequence Y`. Can you try to implement this by passing the context down when creating the string for the type parameters for a type?

simonvonhackewitz:
The last commit (9ad6ed1985be2ca16c179f8beb99d4c29fca660d) fixes case `a`.

fridis (approved):

Issue #3808, opened 2024-09-18
API doc feature kinds mixed up


fridis:
The API doc currently has sections for 

- reference constructors
- value constructors
- functions
- reference types
- value types
- type features

Which IMHO has a few of problems: 

- why should we distinguish ref and value types so prominently? This is IMHO not needed. 
- we risk using a different terminology within `fz` and the API docs.

My suggestion for a structure: Stick closely to AbstractFeature.Kind values:

- Routine, which should be split up in two parts
  - constructors
  - function
- Choice
- Abstract
- Field
  it might make sense to list argument fields as inner features of type field. Some of them even ave local documentation. 
- Intrinsic
- Native
- TypeParameter/OpenTypeParameter  
  these are basically like argument fields

The reason for the separate lists of types is probably the fact that type visibility is different from constructor visibility, so sometimes a type would be visible while the underlying constructor is not (I think the other way around probably does not make sense).

This can maybe be fixed by marking constructors that are not public but that define a public type somehow (e.g., `private constructor`) or adding a section for `types without public constructors`, we need to find a nicer name for that.

PR #3815, opened 2024-09-19
docs: annotate private constructors

simonvonhackewitz:
With #3808, the documentation will no longer have dedicated sections for types. Therefore, it should be clearly indicated when a feature only defines a type and cannot be called."


fridis (approved):

michaellilltokiwa:
@simonvonhackewitz Has conflicts now :-)

PR #3821, opened 2024-09-20
docs: complete restructuring of sections based on `AbstractFeature.Kind`

### LINKED ISSUE #3808: API doc feature kinds mixed up

simonvonhackewitz:
- [x] requires #3814

fix #3808


src/dev/flang/tools/docs/Util.java
```diff
@@ -155,17 +155,13 @@ static enum Kind {
 
     static Kind classify(AbstractFeature af) {
```
michaellilltokiwa, line 156: Is this still used anywhere?

src/dev/flang/tools/docs/Html.java
```diff
@@ -405,19 +434,29 @@ private String mainSection0(TreeSet<AbstractFeature> set, AbstractFeature outer)
    * @return
    */
   private String mainSection0(TreeSet<AbstractFeature> set, boolean printArgs, AbstractFeature outer)
+  {
+    return mainSection0(set.stream(), printArgs, outer);
+  }
+  /**
```
michaellilltokiwa, line 440: empty line missing

michaellilltokiwa (approved):

src/dev/flang/tools/docs/Util.java
simonvonhackewitz: Yes for the navigation that can be accessed with the hamburger menu. That might need restructuring as well, but I'm not sure what the exact purpose of that is. I will create a separate issue for that.

PR #3863, opened 2024-09-26
ast: improve type inference `flat_map` issue

michaellilltokiwa:
It does not yet work is when result of lambda requires boxing, e.g. in:
```
y.flat_map i32 (i->
      if (a % i = 0) && (b % i = 0)
        [-i, i]
      else
        [])
```
array would need to be boxed to Sequence.


fridis (changes requested):
Very nice. 
I think I have found a problem when the type parameter occurs more deeply nested in the formal argument type, this might be solved using `Call.inferGeneric`, see below.

PR #3887, opened 2024-10-01
lib: remove searchable_sequence

michaellilltokiwa:
implemented type.equality for Sequence, error, outcome, option, num_option


lib/Sequence.fz
```diff
@@ -902,6 +899,18 @@ public Sequence(public T type) ref is
   public type.AS_STRING_NON_FINITE_MAX_ELEMENTS => 10
 
 
+  # equality
+  #
+  public fixed redef type.equality(a, b Sequence T) bool
+    pre else
+      T : property.equatable
+      debug : a.finite
+      debug : b.finite
```
fridis, line 908: This does not work like that since the inherited precondition is `true`, the resulting one is `true || T : property.equaltable && debug : a.finint && debug : b. fininte`, which is `true`. 
Apparently there is a bug the implementation of type constraints, the use of `x=y` below should cause an error since the precondition does not say anything about `T` being equatable.

fridis (changes requested):
Thanks for this because it made me rethink how this could be done. 
Your solution does not work like this since `property.equatable.type.equality` has no precondition. What, however, should work, is this:
In `property.equatable`, do this
```
  public type.equality(a, b equatable.this) bool
  pre
    is_equatable
  => abstract

  # additional restrictions on when equality is permitted, 
  # e.g., `option T` might require `T : property.equatable`.
  # to implement `equality`
  #
  public type.is_equatable => true
```
and then in `Sequence`
```
  # Sequence is equatable only if T is
  #
  public redef type.is_equatable => T : property.equatable

  # equality
  #
  public fixed redef type.equality(a, b Sequence T) bool
  =>
    if T : property.equatable 
      a.count = b.count
        && ((a.zip b x,y->x=y) ∀ x->x)
    else
      panic "***"
```
We might (later) even introduce a `compile_time_panic` intrinsic for this that would refuse to create a binary if it is not eliminated when FUIR is created by the monomorphization phase.
What is a little ugly about this solution is that the constraint `T : property.equatable` occurs twice and that `Sequence` appears to always be `equatable`, which is not the case. An alternative would be to add constraint inheritance like 
```
  public Sequence(public T type) ref : (if T : property.equatable then property.equatable) is
```
but this does not make us win a beauty competition either...
Another alternative would be to extend the front end to detect that the redefines `is_equatable` implies `T: property.equatable` such that the `if T : property.equatable` is not needed. But then this analysis becomes harder to implement and the user might get confused about when compile time evaluation is possible and when not (we might even need a `comptime` modifier à la Zig for `is_equatable`...).
For now, maybe try the solution above and create issues for the bug that the constraint in the `pre else` was accepted (see below) and for the possible future enhancements I just outlined.

lib/Sequence.fz
```diff
@@ -902,6 +899,18 @@ public Sequence(public T type) ref is
   public type.AS_STRING_NON_FINITE_MAX_ELEMENTS => 10
 
 
+  # equality
+  #
+  public fixed redef type.equality(a, b Sequence T) bool
```
fridis, line 904: Does this work without `fixed`? I think it should, which would give a default implementation for `array`, `list`, etc.

michaellilltokiwa:
```
./build/lib/Sequence.fz:911:16: error 1: Failed to infer actual type parameters
        && ((a.zip b x,y->x=y) ∀ x->x)

In call to 'Sequence.zip', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'U, V'
Type inference failed for 2 type parameters 'U', 'V'

one error.
make: *** [Makefile:619: build/modules/base.fum] Error 1
```

PR #3899, opened 2024-10-07
docs: hide non public arguments in section 'Fields'

### LINKED ISSUE #3886: hide non public arguments in the docs
e.g. for [array](https://fuzion-lang.dev/docs/array+(5+arguments)/) the dummy arguments `_ unit   [src]` are shown in the (new) fields section in the docs.

simonvonhackewitz:
fix  #3886

### DIFF
--- a/src/dev/flang/tools/docs/Html.java
+++ b/src/dev/flang/tools/docs/Html.java
@@ -287,7 +287,7 @@ public class Html extends ANY
    */
   private String annotatePrivateConstructor(AbstractFeature af)
   {
-    return af.visibility() == Visi.PRIVPUB
+    return af.visibility().eraseTypeVisibility() != Visi.PUB && af.visibility().typeVisibility() == Visi.PUB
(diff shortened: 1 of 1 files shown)


src/dev/flang/tools/docs/Html.java
```diff
@@ -287,7 +287,7 @@ private String annotateAbstract(AbstractFeature af)
    */
   private String annotatePrivateConstructor(AbstractFeature af)
   {
-    return af.visibility() == Visi.PRIVPUB
+    return af.visibility().eraseTypeVisibility() != Visi.PUB && af.visibility().typeVisibility() == Visi.PUB
```
michaellilltokiwa, line 290: I don't understand this logic? imho this should just be `af.visibility().eraseTypeVisibility() != Visi.PUB`?

michaellilltokiwa (approved):

src/dev/flang/tools/docs/Html.java
simonvonhackewitz: I guess the second part is not needed, since features with a non public type should not appear in the docs at all.

PR #3906, opened 2024-10-08
makefile: use asciidoctor-web-pdf

simonvonhackewitz:
> WARN: `asciidoctor-pdf` is deprecated and will be removed in a future version, please use `asciidoctor-web-pdf` instead


michaellilltokiwa:
You would need to adjust all yml-files. But first, find out if `asciidoctor-web-pdf` is already available on brew and ubuntu-latest

Issue #2182, opened 2023-10-31
Argument types inferred from actual arguments leads to stack overflow if used in recursive feature


fridis:
This small example using inferred argument types
```
fib(a) => list a (fib a)
```
results in this
```
 > ./build/bin/fz stackoverflow.fz 

error 1: java.lang.StackOverflowError
	at dev.flang.ast.ResolvedNormalType.toString(ResolvedNormalType.java:498)
	at dev.flang.ast.AbstractType.toStringWrapped(AbstractType.java:1607)
	at dev.flang.ast.ResolvedNormalType.toString(ResolvedNormalType.java:481)
	at dev.flang.ast.ResolvedNormalType.resolveGenerics(ResolvedNormalType.java:578)
	at dev.flang.ast.ResolvedNormalType.resolve(ResolvedNormalType.java:545)
	at dev.flang.ast.Feature$ResolveTypes.action(Feature.java:1288)
	at dev.flang.ast.ResolvedNormalType.visit(ResolvedNormalType.java:523)
	at dev.flang.ast.AbstractCurrent.visit(AbstractCurrent.java:93)
	at dev.flang.ast.AbstractCurrent.visit(AbstractCurrent.java:35)
	at dev.flang.ast.Call.visit(Call.java:1151)
	at dev.flang.ast.Impl.typeFromInitialValues(Impl.java:462)
	at dev.flang.ast.Impl.inferredType(Impl.java:528)

```

PR #3917, opened 2024-10-09
ast: simplify logic of setActualResultType

michaellilltokiwa:


michaellilltokiwa:
> Can you elaborate on why this should work? I do not like to see complex logic replaced by simpler logic without explanation on why this should be okay.

Yes, good point. But. Why was this special case needed in the first place? Comment says because of #2182. But reg_issue2182 still works without it. So I think this can go.

fridis (approved):
I am happy to see this code go away!

PR #3923, opened 2024-10-10
docs: subdivide the api documentation based on library modules

### LINKED ISSUE #3630: API docs for Strings should not include APIs from terminal.fum module
See [this flang.dev](https://git.tokiwa.software/tokiwa/flang_dev/issues/344) issue. The docs for `base.fum`/`terminal.fum`/etc. should maybe be created separatedly and presented so in on fuzion-lang.dev, similar to Java APIs first giving a choice of modules, similar to [the JDK](https://docs.oracle.com/en/java/javase/21/docs/api/index.html).

simonvonhackewitz:
API documentation is now generated per module, e.g. `terminal/string/` only contains those features of `String` that are relevant to the module terminal.

The first page of the documentation now shows a list with all available modules

resolve #3630


src/dev/flang/tools/docs/Docs.java
```diff
@@ -261,27 +263,62 @@ private static boolean isDummyFeature(AbstractFeature af)
...
+
 
   private void run(DocsOptions config)
   {
-    // declared features are sorted by feature name
+    // get all modules
+    var all_modules = allInnerAndInheritedFeatures(universe)
+              .map(af->lf(af)._libModule)
+              .distinct()
+              .filter(m->!m.name().equals("main")) // NYI: CLEANUP: Don't generate page for main module. Is there a better way to do this?
```
michaellilltokiwa, line 318: I think this is good, but we should move "main" to FuzionConstants

src/dev/flang/tools/docs/Docs.java
```diff
@@ -309,31 +346,55 @@ private void run(DocsOptions config)
...
+            try
+              {
+                FileWriter writer = new FileWriter(new File(path.toFile(), "index.html"));
+                var output = htmlTool.content(af);
+                writer.write(output);
+                writer.close();
+              }
+            catch (IOException e)
+              {
+                // TODO Auto-generated catch block
```
michaellilltokiwa, line 374: handle exception: e.g. throw new Error("file not writable" + file) or similar

src/dev/flang/tools/docs/Docs.java
```diff
@@ -309,31 +346,55 @@ private void run(DocsOptions config)
...
+
+      try
+        {
+          FileWriter writer = new FileWriter(new File(path.toFile(), "index.html"));
+          writer.write(htmlTool.modulePage());
+          writer.close();
+        }
+      catch (IOException e)
+        {
+          // TODO Auto-generated catch block
```
michaellilltokiwa, line 393: see above

src/dev/flang/tools/docs/Html.java
```diff
@@ -387,16 +412,16 @@ private String redefines0(AbstractFeature af)
...
     // Type Parameters
-    var typeParameters = new TreeSet<AbstractFeature>();
+    var typeParameters = new LinkedList<AbstractFeature>();
     typeParameters.addAll(map.getOrDefault(AbstractFeature.Kind.TypeParameter, new TreeSet<AbstractFeature>()));
     typeParameters.addAll(map.getOrDefault(AbstractFeature.Kind.OpenTypeParameter, new TreeSet<AbstractFeature>()));
     typeParameters.addAll(outer.typeArguments());
 
     // Fields
-    TreeSet<AbstractFeature> fields =  new TreeSet<AbstractFeature>();
+    var fields =  new LinkedList<AbstractFeature>();
```
michaellilltokiwa, line 421: why linkedlist? why not plain list/arraylist?

simonvonhackewitz: does it make a difference?

src/dev/flang/tools/docs/Html.java
```diff
@@ -387,16 +412,16 @@ private String redefines0(AbstractFeature af)
...
     fields.addAll(map.getOrDefault(AbstractFeature.Kind.Field, new TreeSet<AbstractFeature>()));
     var normalArguments = outer.arguments().clone();
-    normalArguments.removeIf(a->a.isTypeParameter() || a.visibility().eraseTypeVisibility() != Visi.PUB);
+    normalArguments.removeIf(a->a.isTypeParameter());// || a.visibility().eraseTypeVisibility() != Visi.PUB);
```
michaellilltokiwa, line 424: comment missing, why code was comented

src/dev/flang/tools/docs/Html.java
michaellilltokiwa: Almost everything makes a difference in code unfortunately, sometimes even whitespace... :-D
We use a LinkedList when we need a queue, add at end, remove from beginning. Do you need this here? Otherwise I would suggest List.
Performance implications are irrelevant but consistency is quite important in a big code base. So this would be my main argument why to use a List instead of a LinkedList.

src/dev/flang/tools/docs/Html.java
simonvonhackewitz: good that you saw that, it shouldn't be commented out, was for debugging only

src/dev/flang/fe/LibraryFeature.java
```diff
@@ -868,7 +875,57 @@ else if (other instanceof LibraryFeature lf)
     return result;
   }
 
+  /**
+   * Union of the library features of all inner features. Checks inner features recursively.
```
michaellilltokiwa, line 879: I think in this comment library feature and library module is mixed up?

simonvonhackewitz: yes

src/dev/flang/fe/LibraryFeature.java
```diff
@@ -868,7 +875,57 @@ else if (other instanceof LibraryFeature lf)
     return result;
   }
 
+  /**
+   * Union of the library modules of all inner features. Checks inner features recursively.
+   * @return immutable set with all library features for which an inner feature exists
```
michaellilltokiwa, line 880: also here, library modules instead of library features?

simonvonhackewitz: 🙈

PR #3941, opened 2024-10-14
lib: make `list` inherit from `switch`

michaellilltokiwa:
I'm not sure if this is a good idea.
@tokiwa-software/developers What do you think?


michaellilltokiwa:
@maxteufel Maybe there is no point. :-) I currently can't think of an example where this is useful.

PR #3949, opened 2024-10-15
lib: switch, add as_outcome, as_option

### LINKED ISSUE #3938: lib/switch: `as_option` and `as_outcome`
Sometimes you want to convert between options and outcome. It might be a good idea to have that built in to the `switch` feature.

michaellilltokiwa:
fixes #3938

### DIFF
--- a/lib/switch.fz
+++ b/lib/switch.fz
@@ -111,6 +111,22 @@ is
                 | B => nil
 
 
+  # convert this switch to an option
+  #
+  public as_option option A
+  =>
+    switch.this ? a A => a
+                | B => nil
+
+
+  # convert this switch to an outcome
+  #
+  public as_outcome outcome A
+  =>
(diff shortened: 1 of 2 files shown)


lib/switch.fz
```diff
@@ -111,6 +111,22 @@ is
...
+    switch.this ? a A => a
+                | B => nil
+
+
+  # convert this switch to an outcome
+  #
+  public as_outcome outcome A
+  =>
+    switch.this ? a A => a
+                | B => error "switch.as_outcome: failure converting switch to outcome"
```
michaellilltokiwa, line 127:
```suggestion
                | B => error "switch.as_outcome: no error message provided"
```

michaellilltokiwa:
> `outcome.as_outcome` could be redefined to return itself, but that is optional.

Oh yes, we should definitely do that.

PR #3958, opened 2024-10-16
parser: end expr at comma 

### LINKED ISSUE #3761: confusing syntax error in, when loop init with `if/else`
```
for i := if true then "string" else "strong", "strung"
do
```
```

/home/sam/playground/test.fz:1:45: error 1: Syntax error: expected semicolon ';', found comma ','
for i := if true then "string" else "strong", "strung"

While parsing: semicolon or flat line break, parse stack: semiOrFlatLF, exprs, block, elseBlock, ifexpr, term, opTail, opExpr, operatorExpr, exprInLine, implFldInit, indexVar, indexVars, loop, term, opTail, opExpr, operatorExpr, expr, exprs, block, unit


/home/sam/playground/test.fz:1:37: error 2: Expression produces result of type 'String' but result is not used.
for i := if true then "string" else "strong", "strung"

To solve this, use the result, explicitly ignore the result '_ := <expression>' or change 'String' from constructor to routine by replacing'is' by '=>'.

2 errors.

```

simonvonhackewitz:
fix #3761

Example
```
for i := if true then "string" else "strong", "strung"
    j in 1..3
do
  say "$j: $i"
```

```
1: string
2: strung
3: strung
```

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -2510,6 +2510,7 @@ brblock     : BRACEL exprs BRACER
     return switch (currentAtMinIndent())
       {
       case
+        t_comma,
         t_indentationLimit,
         t_lineLimit,
         t_spaceOrSemiLimit,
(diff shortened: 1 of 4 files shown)


michaellilltokiwa (approved):
Nice, one line fixes are the best.

PR #3963, opened 2024-10-17
lib: Sequence implement sort via `pre T:property.partially_orderable`

michaellilltokiwa:


lib/container/map_of.fz
```diff
@@ -28,7 +28,7 @@
 #
 # This feature creates an instance of a map.
 #
-public map_of(K type : property.orderable, V type, ks array K, vs array V) Map K V => ordered_map ks vs
+public map_of(K type : property.orderable, V type, ks array K, vs array V) container.Map K V => ordered_map ks vs
```
fridis, line 31: is this change needed?

lib/container/map_of.fz
```diff
@@ -37,5 +37,5 @@ public map_of(K type : property.orderable, V type, ks array K, vs array V) Map K
 #
 # example: map_of [(key1, value1), (key2, value2)]
 #
-public map_of(K type : property.orderable, V type, kvs array (tuple K V)) Map K V =>
+public map_of(K type : property.orderable, V type, kvs array (tuple K V)) container.Map K V =>
```
fridis, line 40: dto.

fridis (commented):
very nice. Just wondering if a `container.` prefix is required for `Map`, see below.

lib/container/map_of.fz
michaellilltokiwa: Currently, yes:
```
./build/lib/container/map_of.fz:31:87: error 1: Incompatible types in assignment
public map_of(K type : property.orderable, V type, ks array K, vs array V) Map K V => ordered_map ks vs

assignment to field : 'container.map_of.result'
expected formal type: 'container.this.Map container.map_of.K container.map_of.V'
actual type found   : 'container.this.ordered_map container.map_of.K container.map_of.V'
assignable to       : 'container.this.ordered_map container.map_of.K container.map_of.V'
for value assigned  : 'ordered_map ks vs'
To solve this, you could change the type of the target 'container.map_of.result' to 'container.this.ordered_map container.map_of.K container.map_of.V' or convert the type of the assigned value to 'container.this.Map container.map_of.K container.map_of.V'.

one error.
make: *** [Makefile:619: build/modules/base.fum] Error 1
```

PR #3966, opened 2024-10-17
ast: box after resolving sugar2

michaellilltokiwa:
actuals of call to precondition where not properly boxed since call to precondition is inserted during resolving sugar2 and resolving sugar2 happened after boxing.


fridis (approved):
Looks good. 
Getting rid of the CHECKING__TYPES phase is also ok, I have introduced these two phases just as a safety net such that syntax sugar does not break the code. But it is fine to check this later only.

Issue #3388, opened 2024-07-15
Types declared in type features cannot be referenced. Maybe forbid them?


fridis:
This code creates two type features that themselves declare types.
```
test_type is

  type.point(x, y i32) is

  type.opt : choice i32 bool is
```
This currently compiles without problem, but it is useless since there is no syntax to describe the type of `point` or `opt`. We should probably forbid this.

michaellilltokiwa:
we currently have lots of those in stdlib:
```
  type.default_stat_handler : io.file.Stat_Handler is
    redef stats(path fuzion.sys.Pointer, meta_data_arr fuzion.sys.Pointer) =>
      fuzion.sys.fileio.stats path meta_data_arr
```
``
  public type.all : Monoid bool is
    public redef infix ∙ (a, b bool) => a & b
    public redef e => true
```

fridis:
This needs some for thought, the main problem I see is that we should avoid cotypes of cotypes, and these would occur if a type feature `c` is a constructor that defines a type `c` whose outer type is a cotype `o`. If `c.type` is the cotype of  of `c`, what is the outer feature of `c.type`? I see a few possibilities

- the co type of `o`, i.e., the cocotype of `c`, which would open the way to create an endless chain of cotypes.
- just `o`, i.e., the cotype of `c`, which is maybe not really correct
- something else like `Type`, which brutally cuts of the outer features

Since cotypes are all unit type values, the main (only?) effect of all of this is to define what type features should be visibible.

PR #3975, opened 2024-10-21
fe: forbid types in type features

michaellilltokiwa:
This is a suggestion on how to deal with #3388

This does not forbid types that can not be referred to in source code anyway, like lambdas and anonymous features.


src/dev/flang/fe/SourceModule.java
```diff
@@ -1655,6 +1655,22 @@ else if (!t1.isDirectlyAssignableFrom(t2, context) &&  // we (currently) do not
...
+  private void checkLegalDefinesType(Feature f)
+  {
+    if (f.definesUsableType() && InTypeFeature(f))
+      {
+        AstErrors.illegalFeatureDefiningType(f);
+      }
+  }
+
+
+  private boolean InTypeFeature(AbstractFeature f)
```
fridis, line 1671: method names should start in lower case: `inTypeFeature`:
```suggestion
    if (f.definesUsableType() && inTypeFeature(f))
      {
        AstErrors.illegalFeatureDefiningType(f);
      }
  }


  private boolean inTypeFeature(AbstractFeature f)
```

fridis (approved):
I think this avoids a lot of problems.

PR #3981, opened 2024-10-21
lib: rename TRUE/FALSE to right/wrong

michaellilltokiwa:
This PR is a suggestion.

Benefits of the rename, I see:
- right/wrong adhere to naming convention
- right/wrong have same amount of letters, leads to easier indentation in match cases.


fridis (commented):
I fear this is even more confusing than `TRUE`/`FALSE`. The reason for the existence of the `TRUE`/`true` duality is type inference, but this came from a time when Fuzion still had mutable fields. 
Since we have a larger codebase now, it might be time to check if the effect on the code is actually significant, i.e., maybe we could remove the lower case versions and rename `TRUE`/`FALSE` as `true`/`false` with only  a few places where we need to add `bool` as in `effect.fz`: 
```
public abortable => true
```
which should then be
```
public abortable bool => true
```

PR #3992, opened 2024-10-23
lib: fix dead-lock in thread_pool

michaellilltokiwa:
Admittedly I'm not yet 100% sure if there no other possible dead lock scenario.
Previous to this change the test erred no later than ~7 tries, whereas now it runs a 100 times without problems.
old:
```
if !is_closed.read
```
new:
```
if !is_closed.read && data.read.is_empty
```


fridis (approved):
I have opened #3997 to collect ideas for a hight-level monitor / synchronization mechanism and am playing around with APIs now. The result might be that `concur.sync` might become module-visible and used to implement that new mechanism. In any case, will need to check in detail at some point. 
Approve this for now.

PR #3993, opened 2024-10-23
lib: make `infix =` work for list, array etc.

### LINKED ISSUE #3991: `infix =` not working for subtypes of `Sequence`
For the exact type `Sequence` it is working
```
fz -e 'say ([0,0].as_seq = [0,0].as_seq)'
true
```

But not for types inheriting from `Sequence`
```
FUZION_DISABLE_ANSI_ESCAPES=true fz -e 'say ([0,0] = [0,0])'

$MODULE/array.fz:34:15: error 1: Used abstract feature 'property.equatable.type.equality' is not implemented by 'array.type'
module:public array(
--------------^^^^^
Feature 'array.type' instantiated at <source position not available>:

inherits or declares abstract feature 'property.equatable.type.equality' declared at $MODULE/property/equatable.fz:42:15:
  public type.equality(a, b equatable.this) bool
--------------^^^^^^^^
which is called at $MODULE/property/equatable.fz:43:3:
  pre
--^^^
    is_equatable
----^^^^^^^^^^^^
without providing an implementation

call '(array.type i32).precall equality' at $MODULE/equals.fz:29:55:
public equals(T type : property.equatable, a, b T) => T.equality a b
------------------------------------------------------^^^^^^^^^^^^^^
call 'equals#3 (array i32)' at $MODULE/equals.fz:39:56:
public infix =(T type : property.equatable, a, b T) => equals a b
-------------------------------------------------------^^^^^^^^^^
call 'infix =#3 (array i32)' at command line:1:5:
say ([0,0] = [0,0])
----^^^^^^^^^^^^^^^
call 'universe'
program entry point

one error.
```

michaellilltokiwa:
fixes #3991

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -902,10 +902,17 @@ public Sequence(public T type) ref : property.equatable is
 
   # equality
   #
-  public fixed redef type.equality(a, b Sequence T) =>
+  public redef type.equality(a, b Sequence.this) =>
     if T : property.equatable
-      a.count = b.count
-        && ((a.zip b x,y->x=y) ∀ x->x)
+      if a.count != b.count
+        false
+      else
+        for x in a
+            y in b
+        until x != y
+          false
+        else
+          true
     else
       panic "***"
 


fridis (changes requested):
Does the `zip`-based implementation not work when inherited, or is it just too slow?
I think it would be better to keep this fixed since any child of Sequence that adds additional data that would need to be compared as well will just be compared like a `Sequence`. 
Alternatively, for cases where there is no addditional data like `list` or `array`, we can have implementations like
```
public fixed redef type.equality(a, b list T) => (Sequence T).type.equality a b
```

michaellilltokiwa:
> Does the `zip`-based implementation not work when inherited, or is it just too slow? I think it would be better to keep this fixed since any child of Sequence that adds additional data that would need to be compared as well will just be compared like a `Sequence`. Alternatively, for cases where there is no addditional data like `list` or `array`, we can have implementations like
> 
> ```
> public fixed redef type.equality(a, b list T) => (Sequence T).type.equality a b
> ```

I created an issue for the _zip_-problem

PR #3995, opened 2024-10-23
fe: allow _ as a placeholder for a type parameter

michaellilltokiwa:


src/dev/flang/ast/Block.java
```diff
@@ -170,6 +170,18 @@ public UnresolvedType asParsedType()
...
+   * Returns either result of asParsedType()
+   * or Types.t_UNDEFINED in case types name is '_'.
+   */
+  @Override
+  public AbstractType asType()
+  {
+    return _expressions.size() == 1 ? _expressions.getFirst().asType() : null;
+  }
+
+
```
fridis, line 184: This does not seem to be needed since `Expr.asType()` already calles `asParsedType`.

fridis (changes requested):
Nice. I think I found one redundant overridden method.

fridis (approved):

PR #4010, opened 2024-10-24
tests: ctrie, remove: NYI Set should have `except()` feature

michaellilltokiwa:


fridis (approved):
The standard term for `except` is set difference, maybe we should rename this? Or, at least. add `infix -` as an alias! (in a different PR, of course!)

PR #4014, opened 2024-10-25
lib: remove FALSE TRUE

michaellilltokiwa:
This PR to see the impact of removing TRUE  / FALSE.
Admittedly it is peculiar when you first see it but it might be more clear for new learner?


michaellilltokiwa:
@maxteufel Yes, you are probably right.

PR #4018, opened 2024-10-25
lib: change Sequence.first/last to return options

michaellilltokiwa:


modules/nom/src/nom/parsers/json.fz
```diff
@@ -105,7 +105,7 @@ public json =>
         if x.is_empty
           0.0
         else
-          (String.type.from_codepoints x.first)
+          (String.type.from_codepoints x.first.get)
             .parse_u64
             .val 0
             .as_f64 / (10.0*x.count.as_f64))
```
fridis, line 111: Would be nicer to use match instead of `.get`: 
```
        match x.first
          nil        => 0.0
          f Sequence => String.type.from_codepoints f
                              .parse_u64
                              .val 0
                              .as_f64 / (10.0*x.count.as_f64))
```

fridis (changes requested):
With this change, I think code patterns of the form 
```
  if x.is_empty
    a
  else
    b x.first.get
```
should be replaced by
```
  match x.first
    nil => a
    f T => b f
```
Marked this at once source location, but should be done throughout.

PR #4020, opened 2024-10-25
lib: change Sequence.finite to return infinite, infinite, unknown.

michaellilltokiwa:


fridis (approved):
Good, that avoid mis-interpreting this!

PR #4021, opened 2024-10-25
lib: ensure `Sequence.equality` with infinite sequences terminates whenever possible

simonvonhackewitz:
I think comparisons with infinite sequences should terminate when it is possible to determine the result, i.e. when one is finite or when infinite ones differ.

Makes the following possible
```
> fz -e 'say ((1..10).as_seq = (1..).as_seq)'
false

> fz -e 'say ((1..).as_seq = (1..10).as_seq)'
false

> fz -e 'say ([0,0].as_seq = (1..10).as_seq)'
false

> fz -e 'say ((1..).as_seq = (2..).as_seq)'
false
```

### DIFF
--- a/lib/Sequence.fz
+++ b/lib/Sequence.fz
@@ -903,9 +903,18 @@ public Sequence(public T type) ref : property.equatable is
   # equality
   #
   public fixed redef type.equality(a, b Sequence T) =>
+    equality0(a1, b1 Sequence (T2 : property.equatable)) =>
+      ((a1.zip b1 x,y->x=y) ∀ x->x)
+
     if T : property.equatable
-      a.count = b.count
-        && ((a.zip b x,y->x=y) ∀ x->x)
+      if a.finite && b.finite
+        a.count = b.count && equality0 a b
+      else if a.finite
+        a.count = (b.take (a.count+1)).count && equality0 a b
+      else if b.finite
+        b.count = (a.take (b.count+1)).count && equality0 a b
+      else
+        equality0 a b
     else
       panic "***"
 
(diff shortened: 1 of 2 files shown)


michaellilltokiwa:
@tokiwa-software/developers:
I'm unsure if it is not better to error when equality is used on two sequences where one is known finite and the other known infinite?
@fridis Can you way in, please?

michaellilltokiwa:
@tokiwa-software/developers If a.finite != b.finite then equality should also return `false`, IMHO.

lib/Sequence.fz
```diff
@@ -903,9 +903,18 @@ public Sequence(public T type) ref : property.equatable is
...
-      a.count = b.count
-        && ((a.zip b x,y->x=y) ∀ x->x)
+      if a.finite && b.finite
+        a.count = b.count && equality0 a b
+      else if a.finite
+        a.count = (b.take (a.count+1)).count && equality0 a b
+      else if b.finite
+        b.count = (a.take (b.count+1)).count && equality0 a b
+      else
+        equality0 a b
```
fridis, line 917: I think this code is overly complex and not needed. It is suffcient to just use `equality0`:
```suggestion
      equality0 a b
```
or, if we want to optimize the case when both are known to be finite but might differ in their count: 
```suggestion
      !(a.finite && b.finite && a.count != b.count) && equality0 a b
```

fridis (changes requested):
Good idea. I think the code could be simplified, though.

lib/Sequence.fz
fridis: you are right, the code is needed.

PR #4023, opened 2024-10-28
fe: infer result type of `redef` via redefined feature

michaellilltokiwa:


tests/infer_result_type_via_redefined/infer_result_type_via_redefined.fz
```diff
@@ -0,0 +1,35 @@
...
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test infer_result_type_via_redefined
+#
+# -----------------------------------------------------------------------
+
+infer_result_type_via_redefined =>
+
```
fridis, line 25: I am missing some test cases: 

- redefining two or more inherited features with equal types 
- redefining result types `a` and `b` as `void` 
- redefining `a.f unit`,  `b.f void`,  `c.f String` in 
  ```
    a is
      f unit =>
    b is
      f => do
    c is
      f => "hi"
    d : a, b, c is
      redef f => panic "bla"
  ```
  should IMHO work. 
 - negative tests

fridis (changes requested):
Nice. 

I have some suggestions how to get this to work with type parameters and this type and some improvements.

PR #4024, opened 2024-10-28
lib: rename TRUE/FALSE to true_t/false_t

michaellilltokiwa:


fridis:
I think this is nicer than `TRUE`/`FALSE`, but it is unclear to me what `t` stands for. How about `true_val`/`false_val` or `true_unit`/`false_unit` or just `true_`/`false_`?

michaellilltokiwa:
> I think this is nicer than `TRUE`/`FALSE`, but it is unclear to me what `t` stands for. How about `true_val`/`false_val` or `true_unit`/`false_unit` or just `true_`/`false_`?

In my head _t was for `type` since true_t/false_t are only usable as types outside of bool.fz

fridis (approved):
good idea to use just a `_` as suffix.
When merged, can you check our tutorial?

PR #4037, opened 2024-10-29
lib: change float.type.sin/float.type.cos,... to float.sin/float.cos,...

### LINKED ISSUE #2362: type features taking only outer feature as argument, could be used as instance method with no args
example:
```
float is
   public type.sqrt (val float.this) => square_root val
```
currently has to be used like this:
```
f64.type.sqrt 5.5
```
would be nice if this would work as well
```
(5.5).sqrt
```

michaellilltokiwa:
fixes #2362


fridis (approved):
Looks good. 

Not sure if we want prefix variants like `prefix_math.sin x` in addition to `x.sin`. Maybe keep this as a new issue after #2362 was closed.

Issue #3051, opened 2024-05-14
precondition and postcondition inheritence does not work yet


fridis:
Redefined features should inherit the original feature's contract, combine it when `pre else` or `post then` is present and check it at runtime.

PR #4050, opened 2024-10-30
lib: add temporary preconditions for `/` and `%` in integer types

simonvonhackewitz:
required as long as precondition inheritance is not working, see #3051


michaellilltokiwa (approved):

michaellilltokiwa:
> I'd prefer using regular `#`-style comments over `//`-style comments, `#` seems to have taken over as the default

Ah yes, me too. I missed that.

PR #4065, opened 2024-10-31
doc: add profiling guide

maxteufel:


michaellilltokiwa:
> Does this deserve its own Makefile target or is the GitHub rendered view enough?

I would prefer if it would be built like the reference manual.

PR #4067, opened 2024-10-31
ast: AbstractType.isFunctionType, simplify isFunctionType

michaellilltokiwa:


fridis (approved):
Looks good. 

But why do you change the test in this PR?

michaellilltokiwa:
@fridis Because it does not work anymore and IMHO that is correct since F() tries executes F without any args?

PR #4086, opened 2024-11-04
ast: warn about unused fields

### LINKED ISSUE #3321: raise error on unused field
since this is likely not programmers intent.

simonvonhackewitz:
fix #3321


simonvonhackewitz:
Right now the errors result in the code not being executed, should this be changed?

michaellilltokiwa:
> Right now the errors result in the code not being executed, should this be changed?

No, errors mean the code must not be executed.

lib/String.fz
```diff
@@ -277,7 +277,7 @@ public String ref : property.equatable, property.hashable, property.orderable is
       .map x->
         match x
           c codepoint => c
-          e error     => codepoint 0xFFFD # 'REPLACEMENT CHARACTER' (U+FFFD)
+          _ error     => codepoint 0xFFFD # 'REPLACEMENT CHARACTER' (U+FFFD)
```
michaellilltokiwa, line 280:
```suggestion
          error     => codepoint 0xFFFD # 'REPLACEMENT CHARACTER' (U+FFFD)
```

tests/visibility_negative/visibility_negative.fz.expected_err
```diff
@@ -2312,69 +2312,69 @@ Target feature: 'visibility_negative.visi1.a2.b3.c6'
...
 Feature not found: 'b' (no arguments)
 Target feature: 'visibility_negative.visi8'
 In call: 'b'
 
+
+<built-in>: error 297: Incompatible types when passing argument in a call
+Actual type for argument #1 'b' does not match expected type.
+In call to          : 'visibility_negative.chck'
+expected formal type: 'bool'
+actual type found   : '**error**'
```
michaellilltokiwa, line 2375: Why is this error here now?

lib/String.fz
michaellilltokiwa: Not just here but any match case.

tests/visibility_scoping/visibility_scoping.fz
```diff
@@ -84,17 +84,17 @@ visibility_scoping is
 
   test6 is
     match true
-      f FALSE =>
+      _ FALSE =>
```
michaellilltokiwa, line 87: merge main and update

michaellilltokiwa (changes requested):

tests/visibility_negative/visibility_negative.fz.expected_err
simonvonhackewitz: What I remember this could not be avoided, I think we even talked about it. But I don't remember what it was and I could not find the code that causes it again now

michaellilltokiwa:
@simonvonhackewitz has conflicts now

simonvonhackewitz:
I don't get why some of the errors occur here, some of the files that cause problems were not touched in this PR at all.

michaellilltokiwa (approved):

PR #4095, opened 2024-11-04
lib: structured concurrency

michaellilltokiwa:
An attempt for structured concurrency in fuzion.
Main idea is that `concur.threads` keeps track of spawned threads and joins them when effect cleans up after itself.

### DIFF
--- a/lib/concur/threads.fz
+++ b/lib/concur/threads.fz
@@ -45,10 +45,6 @@ is
     st
 
 
-  type.install_default =>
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
> You say
> 
> > As of now not usable since with concur.threads can not be used with default_thread_handler and instate_self.
> 
> Is this still true even if no longer marked draft?

No, this was an old comment. I removed it.

PR #4097, opened 2024-11-04
lib: add infix version of instate_self

michaellilltokiwa:
This PR is meant as a suggestion.


fridis (approved):
I like this!

fridis:
This permits to convert `e.instate_self code` into `e ! code`.

What I am more worried about, however, is `effect.type.instate`. Would be great if we had infix operands that correspond to `T.instate e code` that would be consistent with that, e.g., `T <- e ! code`. Play around with this and check possibilities!

PR #4113, opened 2024-11-06
lib: io.file/dir remove parameter T, used for distinguishing

michaellilltokiwa:
Inheritance can instead be used to distinguish several files. Thus in fileiotests
`f => io.file` becomes `f : io.file is`

Also moved exists/write to own files.

### DIFF
--- a/lib/io/dir/open.fz
+++ b/lib/io/dir/open.fz
@@ -23,10 +23,8 @@
 
 
 # effect for manipulating open directories
-# T is used to distinguish several open directories
 #
-module:public open(public T type,
-                   fd i64,
+module:public open(fd i64,
                    public path String) : effect is
 
   # reads the next entry of this directory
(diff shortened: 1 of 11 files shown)


michaellilltokiwa:
> Good in general! This will prevent weird calls like I fixed in [tokiwa-software/fzweb@7ef5b8c#diff-e74ef0494e64c3f0c37f8b66bf9e6f002c1ebd0d11b6b1893fced5ce898b25daL102-R106](https://github.com/tokiwa-software/fzweb/commit/7ef5b8c2dc5ff7fd17b3c5b0efec675a1ac309e4#diff-e74ef0494e64c3f0c37f8b66bf9e6f002c1ebd0d11b6b1893fced5ce898b25daL102-R106).
> 
> Why are default effect handlers moved from type features to regular features?

Ah, good question. Consider this:

```
  # install default effect io.file.seek
  type.install_default unit =>
    (io.file.seek default_seek_handler).default    
```
we need `io.this.file.this` but we don't have this in the type feature, so this does not work:
```
  # install default effect io.file.seek
  type.install_default unit =>
    (seek default_seek_handler).default    
```

michaellilltokiwa:
> What's the difference between the two examples, besides the second one being a tiny bit shorter to write?

The second one does not work since in a type feature we have to fully qualify all calls since the target is basically universe.

PR #4117, opened 2024-11-06
dfa: At end of DFA, no extra iterations after `_fuir.lookupDone()` was called

fridis:
Calling `_fuir.lookupDone()` results in soem clazzes having `clazzIsUnitType(cl)` return true, which DFA turns into `Value.UNIT`.  These new values could result into several extra DFA iterations until a fixpoint is reached again.

This results in a speedup of about 20% on a Hello-World example: before

     > time ./build/bin/fz -verbose=0 -e 'say "hi"'
    hi

    real	0m1,176s
    user	0m6,102s
    sys	0m0,341s

now:

     > time ./build/bin/fz -verbose=0 -e 'say "hi"'
    hi

    real	0m0,971s
    user	0m4,805s
    sys	0m0,337s

### DIFF
--- a/src/dev/flang/fuir/analysis/dfa/DFA.java
+++ b/src/dev/flang/fuir/analysis/dfa/DFA.java
@@ -1264,7 +1264,6 @@ public class DFA extends ANY
   void findFixPoint()
   {
     var cnt = 0;
-    var lookupDone = false;
     do
       {
         cnt++;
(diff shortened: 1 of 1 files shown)


michaellilltokiwa (approved):
Cool, but IMHO Hello World example is fairly uninteresting. What about examples with more code?

fridis:
> Cool, but IMHO Hello World example is fairly uninteresting. What about examples with more code?

I am sure that larger applications will profit from the 0.2s speedup as well, likely much more. 

I currently focus on getting all redundancy out of small examples since that is much easier, then gradually increasing the size of the test cases.

PR #4123, opened 2024-11-06
docs: automatically determine modules to include

### LINKED ISSUE #3924: How to determin modules to include in the docs without manual intervention?
The modules that should be included in the documentation are currently manually specified in `Docs.java`.

I believe an automatic solution would be preferable.

Was there a specific reason why it is done like this? And what would be an easier way that does not require manual intervention when new modules are added or existing ones renamed?

simonvonhackewitz:
fix #3924


michaellilltokiwa:
Hmm, this works now but what if in the future we have modules that are not compatible? (E.g. have duplicate features)

simonvonhackewitz:
I have a new approach that restricts the changes to Docs.java and excludes all Java modules, because they caused an endless recursion when the doc generation is used on them.

see #4149

simonvonhackewitz:
> Hmm, I think this should be done differently. loadAllModules args is not necessary. Instead put the function to determine all modules into Util. Then you can use this in docs when calling constructor of universe.

Are you referring to this PR or to the new one I linked? Have you had a look at the new PR?

simonvonhackewitz:
superseded by #4149

PR #4126, opened 2024-11-07
naming: AbstractFeature.isThisRef -> isRef

michaellilltokiwa:
I find this less confusing.


fridis (approved):
Yes, confused me as well.

PR #4130, opened 2024-11-07
lib: add region-effect

michaellilltokiwa:
the idea is to have array allocations freed when effect is uninstalled.


fridis:
That would most likely require runtime checks for assignments not to create dangling references. Java's RTSJ had that for their `ScopedMemory`, but this was awful and resulted in hard to write code.

PR #4138, opened 2024-11-07
ast: flag error if type in effect signature is not an effect

### LINKED ISSUE #4084: effects signature should be checked for validity
Currently the effect signature is just parsed then thrown away.
```
public type.read_bytes(n i32) Sequence u8 ! reader =>
```
```
    var r = returnType();
    var eff = effects();
    var hasType = r instanceof FunctionReturnType;
    var inh = inherits();
    Contract c = contract(forkAtFormArgs);
    Impl p =
      a  .isEmpty()    &&
      eff == UnresolvedType.NONE &&
      inh.isEmpty()       ? implFldOrRout(hasType)
                          : implRout(hasType);
    p = handleImplKindOf(pos, p, i == 0, l, inh, v);
    l.add(new Feature(v,m,r,name,a,inh,c,p))
```

First step is to pass the effects to the feature constructor and check the list of types if they even exist.

simonvonhackewitz:
fix #4084

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2298,6 +2298,14 @@ public class AstErrors extends ANY
           - change the type to a legal"""  + " " + skw(".this") + " type.");
   }
 
+  public static void notAnEffect(AbstractType t, SourcePosition pos)
+  {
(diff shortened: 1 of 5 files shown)


michaellilltokiwa (approved):
Nice

PR #4161, opened 2024-11-12
dfa: Add infrastructure to show values produced during DFA

fridis:
This adds optional debug output for performance analysis.

Setting the env var

    dev_flang_fuir_analysis_dfa_DFA_SHOW_VALUES=

now results in output of the values produced by DFA, while setting

    dev_flang_fuir_analysis_dfa_DFA_SHOW_VALUES=u32

additionally prints all the values of a clazz given by name, `u32` in this case.

### DIFF
--- a/src/dev/flang/fuir/analysis/dfa/Call.java
+++ b/src/dev/flang/fuir/analysis/dfa/Call.java
@@ -238,7 +238,7 @@ public class Call extends ANY implements Comparable<Call>, Context
       {
         var a0 = _args.get(i);
         var a1 =  args.get(i);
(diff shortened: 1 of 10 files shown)


fridis:
> Do you need to set the environment variable `dev_flang_fuir_analysis_dfa_DFA_SHOW_VALUES` to an _empty string_ for it to show everything? This avoids any special handling of a string that prints everything, but is a bit counterintuitive to how env vars are usually used: an empty string is usually equivalent to unsetting it.

I have changed this to support `on` and `off`  and treat empty string like `off`.

Removed the unrelated and broken change to `DFA.cache`.

PR #4173, opened 2024-11-13
readme: shorten labels of badges

michaellilltokiwa:
removed text `run tests on `

[ci skip]


michaellilltokiwa:
> Do note that this will not change the text of the images, which is generated from the `name` field in the workflow's yaml file. This will change the `alt` text of the image. Is this what you wanted?

Oh, no. That is not what I want, closing :-)

PR #4189, opened 2024-11-15
turn `env` into a simple call

michaellilltokiwa:


michaellilltokiwa:
@fridis I wonder what the downsides of this approach are but can not really think of any major ones?

tests/effect_installed_negative/test_effect_neg.fz.expected_err
```diff
@@ -4,65 +4,68 @@
 ----------------^^^^^
 Callchain that lead to this point:
 
+effect environment 'test_effect_neg.b, test_effect_neg.c' for call to 'test_effect_neg.type.a.type.get' at --CURDIR--/test_effect_neg.fz:37:17:
+  use_a => _ := a.env
```
fridis, line 8: I think this kind of error is very confusing. The user did not call `get` anywhere.

src/dev/flang/fuir/analysis/dfa/DFA.java
```diff
@@ -2068,6 +2058,25 @@ static void setArrayElementsToAnything(Call cl, int argnum, String intrinsicName
...
+      var result = cl.getEffectCheck(ecl);
+      if (result == null && cl._dfa._reportResults)
+        {
+          DfaErrors.usedEffectNotInstalled(cl._dfa._fuir.sitePos(cl._site),
+                                           cl._dfa._fuir.clazzAsString(ecl),
+                                           cl);
+          cl._dfa._missingEffects.put(ecl, ecl);
+        }
+      return result;
+    });
```
fridis, line 2073: you should either usee `Call.getEffectForce` here or remove `Call.getEffectForce` which is otherwise unused.

fridis (changes requested):
Very nice simplification!

Only concern I have is that `get` is too generic as a name, maybe it should be called `from_env`, or even just `env` and we could get rid for the `env` keyword (in a separate PR).

PR #4207, opened 2024-11-19
lib: move array2/array3 instatiation to type feature (type.new)

michaellilltokiwa:


michaellilltokiwa:
Hmm, my thought was consistency with array but array has but. We could also keep both options here?

michaellilltokiwa:
Ah yes, but we should be able to make inference work for both cases. 
Then one would simply write `array.new 10 i->i`

PR #4208, opened 2024-11-19
be/c: split generated code in .h and .c file

michaellilltokiwa:


michaellilltokiwa:
Very roughly we have about 50% header file, 50% c file. The intention is to be able to use the structs from native C-code.

PR #4221, opened 2024-11-20
lib: Make sure that instance of `mutate` gets a unique id for each instate, fix #4218

### LINKED ISSUE #4218: `mutate` instance can be re-entered, permitting mutation without visible effect
In this example, feature `f` does not have any visible effects apart from `io.out`, but it performs a state change to `v`
```
m : mutate is
m1 := m

v := m1.instate_self ()->
  say "start"
  r := m.env.new 42
  say "r = $r"
  r

# f uses no effects, but mutates `v`
f ! io.out =>
  _ := m1.instate_self ()->
    v <- 4711 + v
    say "updated v via m1: $v"

f; f; f
```
which can be seen when running the code that produces different output
```
 > ./build/bin/fz repeated_mutate1.fz 
start
r = 42
updated v via m1: 4753
updated v via m1: 9464
updated v via m1: 14175
```

fridis:
This is then used to check that mutations are performed within the same `instate` call, i.e., we cannot hide a mutation by just temporarily instating the corresponding `mutate` effect.

Also add a regression test and fix the strange behavior of `fuzion.sys.misc.uniq_id` with the JVM backend. 

This currently breaks some code using `io.buffered.write` that needs to be checked.

### DIFF
--- a/lib/effect.fz
+++ b/lib/effect.fz
@@ -160,7 +160,7 @@ public effect is
                       ) R
   =>
     x := instate_helper R effect.this e code def
-    instate0 e x.call_code x.call_def
+    instate0 e.make_unique_at_instate x.call_code x.call_def
     x.res.get
 
 
@@ -254,6 +254,11 @@ public effect is
   #
   type.unsafe_get =>
     effect.this.env
+
+
+  # make sure this effect instance has a unique value when it is instated
+  #
+  module make_unique_at_instate => effect.this
 
 
   # --------------------  replacing effect instances  --------------------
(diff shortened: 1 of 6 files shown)


lib/io/buffered/writer.fz
```diff
@@ -27,13 +27,15 @@
 #
 # note: anything in the buffer when effect is uninstalled will be discarded.
 #
-public writer(wp Write_Provider, buf_size i32) : mutate
+public writer(wp Write_Provider, buf_size i32) => writer wp buf_size nil
+private:public writer(wp Write_Provider, buf_size i32, buff option (container.Circular_Buffer u8 LM)) : effect # : mutate
```
michaellilltokiwa, line 31: I think `LM` should be `writer.this`?

PR #4223, opened 2024-11-20
ast: remove checkConstraings

fridis:
Seems to be redundant


fridis:
This is not redundant, the call to `checkActualTypePars` in `Call.java` does not cover all cases.

PR #4262, opened 2024-11-25
test: add reg test #2559

### LINKED ISSUE #2559: feature declarations in lazy expressions should be possible

michaellilltokiwa:
fixes #2559

### DIFF
--- a/tests/reg_issue2559/Makefile
+++ b/tests/reg_issue2559/Makefile
@@ -1,0 +1,25 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


michaellilltokiwa:
depends on #4282

PR #4265, opened 2024-11-25
ast: fix issue in remove_type_parameter_used_for_this_type_in_cotype …

### LINKED ISSUE #4247: Boxing not done properly when passing `this`-type in a ref feature to argument type `Any`
This example
```
e ref is
  type.x(a e.this) =>
    say a

f : e is

e.type.x f
```
should box `a` when it is passed to `say`, but it produces an error
```
 > ./build/bin/fz test_box.fz

/home/fridi/fuzion/clean/fuzion/test_box.fz:3:9: error 1: Incompatible types when passing argument in a call
    say a

Actual type for argument #1 's' does not match expected type.
In call to          : 'say'
expected formal type: 'Any'
actual type found   : 'e.this.type (in type feature)'
assignable to       : 'e.this.type (in type feature)'
for value assigned  : 'a'
To solve this, you could change the type of the target 's' to 'e.this.type (in type feature)' or convert the type of the assigned value to 'Any'.

one error.

```

michaellilltokiwa:
…with boxed this.type

boxing was falsely also remove when removing type parameter.

fixes #4247

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -1672,7 +1672,9 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
         var tf = tp.outer();
         if (tf.isCotype() && tp == tf.arguments().get(0))
           { // generic used for `abc.this.type` in `abc.type` by `abc.this.type`.
-            result = tf.cotypeOrigin().selfType().asThis();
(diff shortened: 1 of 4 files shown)


fridis (approved):
I am very surprised about this simple solution, while I was trying to fix this for the last three days without much success...

PR #4266, opened 2024-11-25
ast: do not suggest to make scoped fields public in unused field error

### LINKED ISSUE #4199: for scoped fields, error message should not suggest making it `public`
The new error message "Unused, non public field" shows up even in cases like this:
```
src/Session.fz:176:7: error 2: Unused, non public field 'e'
      e error => panic "error reading from path: {file}"

To solve this, either
  - use the field
  - make it 'public'
or
  - explicitly ignore the result '_ := …'
```
where (in this example) a `match` is present. For cases like this, and likewise for all cases where the field is scoped, the error message should not suggest to make the field `public` because that will not make sense.

simonvonhackewitz:
fix #4199

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2322,7 +2322,7 @@ public class AstErrors extends ANY
           To solve this, do either of the following
             - use the field
           """ +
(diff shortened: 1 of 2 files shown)


michaellilltokiwa:
Can you please change `make it public`:
mark it as public
or
set it to public 

Not sure which one is better??

michaellilltokiwa (approved):

PR #4277, opened 2024-11-25
tests: fix redef_with_type_parameters case2

michaellilltokiwa:


fridis (approved):
nice. 

For a separate PR: Would be good to avoid the code duplication in `AbstractFeature.typeCall` and `AbstractCall.typeCall`.

PR #4284, opened 2024-11-26
parser: show error for `abstract`, `intrinsic` and `native` constructors

### LINKED ISSUE #2452: specifying an abstract constructor feature should result in an error
e.g.

```
ex is
  something is abstract
```

simonvonhackewitz:
fix #2452

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2209,10 +2209,10 @@ public class AstErrors extends ANY
         " or remove the return type if you want to define a constructor.");
   }
 
-  public static void unimplementetConstructor(SourcePosition pos, String keyword)
+  public static void unimplementedConstructor(SourcePosition pos, String keyword)
   {
     error(pos, keyword + " feature must not be constructor.",
-        "A constructor feature must always be implemented. " +
+        "A constructor feature must always be implemented.\n" +
         "To solve this, either implement the constructor feature or change it to a function feature by replacing " +
         skw(keyword) + " with " + skw("=>") + ".");
   }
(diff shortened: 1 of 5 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -2209,6 +2209,14 @@ public static void constructorWithReturnType(SourcePosition pos)
         " or remove the return type if you want to define a constructor.");
   }
 
+  public static void unimplementetConstructor(SourcePosition pos, String keyword)
```
michaellilltokiwa, line 2212: unimplementetConstructor => unimplementedConstructor

src/dev/flang/ast/AstErrors.java
```diff
@@ -2209,6 +2209,14 @@ public static void constructorWithReturnType(SourcePosition pos)
...
+  {
+    error(pos, keyword + " feature must not be constructor.",
+        "A constructor feature must always be implemented. " +
```
michaellilltokiwa, line 2215:
```suggestion
        "A constructor feature must always be implemented.\n " +
```

src/dev/flang/parser/Parser.java
```diff
@@ -3389,11 +3389,8 @@ List<Cond> invariant(boolean atMinIndent)
   /**
    * Parse implRout
    *
-implRout    : "is" "abstract"
-            | ARROW "abstract"
-            | "is" "intrinsic"
+implRout    : ARROW "abstract"
```
michaellilltokiwa, line 3392: I think we should leave the grammar as is. Parser does parse e.g. `is abstract` even though it then emits an error.

fridis:
please change the title, the parser should never `throw` any errors but produce / display / show errors.

src/dev/flang/parser/Parser.java
fridis: I think we shoud do both: change the grammar (which becomes simpler) and change the parser to not accept `is abstract`, etc. Then, we might not need new error messages and everything becomes easier...

PR #4298, opened 2024-11-26
parser: require dots `...` after open types

### LINKED ISSUE #768: `...` syntax and open generics
from `lib/Function.fz`:

```
Function(R type, A type...) ref is
  call(a A) R is abstract
```

It is a bit confusing that we have `A type...` but no `...` in `call(a A)`. It might be good to require some kind of `...`-syntax in the latter as well, you can easily miss this if there were 100 lines of code between the first and second line of the example.

simonvonhackewitz:
fix #768

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -32,7 +32,6 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
 
-import dev.flang.ast.AstErrors.PreOrPost;
 import dev.flang.util.ANY;
 import static dev.flang.util.Errors.*;
 import dev.flang.util.Errors;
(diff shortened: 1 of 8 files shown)


src/dev/flang/ast/AstErrors.java
```diff
@@ -2351,6 +2350,20 @@ public static void ambiguousResultType(Feature f)
        + "  return the type of the reference itself (instead of " + s(f.resultType()) + " return " + s(f.resultType().asRef()) + ").");
   }
 
+  public static void openGenericMissingDots(SourcePosition pos, AbstractType t)
+  {
+    error(pos, "open type is not followed by " + skw("..."),
+          "An open type must be followed by " + skw("...") + ".\n"
+          + "To solve this add " + skw("...") + " after the highlighted error.");
```
michaellilltokiwa, line 2357: To solve this comma

PR #4312, opened 2024-11-27
lib: remove intrinsics mapped_buffer_get/set

michaellilltokiwa:
We already have getel/setel from internal_array.
Only ugliness is that index is i64 in one case i32 in the other but that should be changed in the future.


michaellilltokiwa:
these are now _natives_

PR #4314, opened 2024-11-27
ast: remove code duplication in `AbstractFeature.typeCall` and `AbstractCall.typeCall`

### LINKED ISSUE #4297: Would be good to avoid the code duplication in AbstractFeature.typeCall and AbstractCall.typeCall
from #4277

simonvonhackewitz:
fix #4297

### DIFF
--- a/src/dev/flang/ast/AbstractCall.java
+++ b/src/dev/flang/ast/AbstractCall.java
@@ -233,11 +233,7 @@ public abstract class AbstractCall extends Expr
           }
       }
 
-    var typeCallResult = (target() instanceof AbstractCall ac && !ac.isCallToOuterRef())
-                            ? ac.typeCall(p, res, that)
-                            :  null;
-
-    return calledFeature().typeCall(p, typeParameters, res, that, typeCallResult);
+    return calledFeature().typeCall(p, typeParameters, res, that, target());
   }
 
 
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -809,34 +809,19 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
    *
    * @param that the original feature that is used to lookup types.
    *
-   * @param abstractCallTypeCall inheritance call for a parent feature in case this was called from an
-   *                             AbstractCall.typeCall that is not a call to an outer ref, might be null otherwise
+   * @param target the target of this typeCall, null for recursive calls for
+   * outer typeCalls of to this method.
    *
    * @return instance of Call to be used for the parent call in cotype().
    */
-  Call typeCall(SourcePosition p, List<AbstractType> typeParameters, Resolution res, AbstractFeature that, Call abstractCallTypeCall)
+  Call typeCall(SourcePosition p, List<AbstractType> typeParameters, Resolution res, AbstractFeature that, Expr target)
   {
     var o = outer();
-    Expr oc;
-    if (o == null || o.isUniverse())
-      {
-        oc = new Universe();
-      }
-    else if (abstractCallTypeCall != null)
-      {
-        oc = abstractCallTypeCall;
-      }
-    else
-      {
-        var typeParameters2 = new List<AbstractType>();
-        for (var tp : new List<>(o.selfType(), o.generics().asActuals()))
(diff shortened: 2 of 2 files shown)


simonvonhackewitz:
I think there should be a better solution, but I don't quite understand the code.

fridis (changes requested):

src/dev/flang/ast/AbstractCall.java
```diff
@@ -233,21 +233,11 @@ Call typeCall(SourcePosition p, Resolution res, AbstractFeature that)
...
-
-    var tf = calledFeature().cotype(res);
-
-    return new Call(p,
-                    oc,
-                    typeParameters,
-                    Expr.NO_EXPRS,
-                    tf,
-                    tf.selfType());
+      var typeCallResult = (target() instanceof AbstractCall ac && !ac.isCallToOuterRef())
```
michaellilltokiwa, line 236: I think the indentation here is a bit wrong?

fridis (changes requested):
The patch seems correct, but the code does not really get simpler. 

Can you try the following patch instead. I made no semantic changes, just refactoring after first putting the logic fo `oc` from `AbstractCall` to `typeCall` and then moving the logic for `typeParameters2` out of `typeCall` to the recursive call to `typeCall` using `map`.

```
diff --git a/src/dev/flang/ast/AbstractCall.java b/src/dev/flang/ast/AbstractCall.java
index 16d9b3ec7..eed39d497 100644
--- a/src/dev/flang/ast/AbstractCall.java
+++ b/src/dev/flang/ast/AbstractCall.java
@@ -231,21 +231,7 @@ public abstract class AbstractCall extends Expr
           }
       }
 
-    var o = calledFeature().outer();
-    Expr oc = o == null || o.isUniverse()
-      ? new Universe()
-      : (target() instanceof AbstractCall ac && !ac.isCallToOuterRef())
-      ? ac.typeCall(p, res, that)
-      : o.typeCall(p, new List<>(o.selfType(), o.generics().asActuals()), res, that);
-
-    var tf = calledFeature().cotype(res);
-
-    return new Call(p,
-                    oc,
-                    typeParameters,
-                    Expr.NO_EXPRS,
-                    tf,
-                    tf.selfType());
+    return calledFeature().typeCall(p, typeParameters, res, that, target());
   }
 
 }
diff --git a/src/dev/flang/ast/AbstractFeature.java b/src/dev/flang/ast/AbstractFeature.java
index cfab7c9f4..92da4a6b1 100644
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -806,25 +806,24 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
    *
    * @param that the original feature that is used to lookup types.
    *
+   * @param target the target of this typeCall, null for recursive calls for
+   * outer typeCalls of to this method.
+   *
    * @return instance of Call to be used for the parent call in cotype().
    */
-  Call typeCall(SourcePosition p, List<AbstractType> typeParameters, Resolution res, AbstractFeature that)
+  Call typeCall(SourcePosition p, List<AbstractType> typeParameters, Resolution res, AbstractFeature that, Expr target)
   {
     var o = outer();
-    var oc = o == null || o.isUniverse()
-      ? new Universe()
-      : outer().typeCall(p, new List<>(o.selfType(), o.generics().asActuals()), res, that);
+    var oc = o == null || o.isUniverse()                            ? new Universe()
+      : target instanceof AbstractCall ac && !ac.isCallToOuterRef() ? ac.typeCall(p, res, that)
+      : o.typeCall(p, new List<>(o.selfType(),
+                                 o.generics().asActuals().map(that::rebaseTypeForCotype)),
+                   res, that, null);
+
     var tf = cotype(res);
-    var typeParameters2 = new List<AbstractType>();
-    for (var tp : typeParameters)
-      {
-        typeParameters2.add(typeParameters2.size() == 0
-                              ? tp
-                              : that.rebaseTypeForCotype(tp));
-      }
     return new Call(p,
                     oc,
-                    typeParameters2,
+                    typeParameters,
                     Expr.NO_EXPRS,
                     tf,
                     tf.selfType());
```

michaellilltokiwa (approved):

PR #4317, opened 2024-11-27
lib: add `replace` returning old value, change `put` to return nothing

### LINKED ISSUE #3124: > This looks terribly complicated to me, especially in the case of `for` loops and similar. Is there any motivation behind this change?
> This looks terribly complicated to me, especially in the case of `for` loops and similar. Is there any motivation behind this change?

The motivation is that ignoring a result accidentally is usually a bug that indicates that error handling code is missing or dead code was written. Of course, the APIs and expressions like loops should be designed in a way that this does not create code bloat, e.g. instead of having a mutable map with `put` that returns the old value
```
  _ := m.put k v1
  old_v := m.put k v2
  _ := m.put k v3
```
we should better have two APIs calls, e.g., 
```
  m.put k v1
  old_v := m.replace k v2
  m.put k v3
```

_Originally posted by @fridis in https://github.com/tokiwa-software/fuzion/issues/3094#issuecomment-2129637307_

simonvonhackewitz:
add `replace` that returns the old value in places where there is a `put` and change the `put` to return nothing.

fix #3124

### DIFF
--- a/modules/base/src/container/mutable_tree_map.fz
+++ b/modules/base/src/container/mutable_tree_map.fz
@@ -93,10 +93,15 @@ is
 
   # add the mapping k -> v as a new entry to this map
   #
+  public put(k KEY, v VAL) =>
+    _ := replace k v
+
+  # add the mapping k -> v as a new entry to this map
+  #
   # returns the value that k previously mapped to, or nil if
   # k was not yet contained in this map
   #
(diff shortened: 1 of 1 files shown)


simonvonhackewitz:
Is there only one or did I miss some?

Are there features other than `put` that should have the same change?

PR #4322, opened 2024-11-27
tests: in `negative.fz` use `$(FILE)` instead of `$(NAME)`

fridis:
This makes it clear which source file is loaded and avoids searching for source files.

In particular, this also permits executable code in the universe of that source file, otherwise, it would be ambiguous when to execute that code.

This revealed an error in `reg_issue4273` where the negative tests were ignored due to a typo in the comments resulting in the lines that must fail not being found by `grep`.


fridis:
This is held back until #5545, #5547 and #5549 are merged.

michaellilltokiwa (approved):

PR #4378, opened 2024-12-02
jvm: Update dependencies for -jar backend

fridis:
There are a few new classes in package `dev.flang.be.jvm.runtime` or required by that package.


michaellilltokiwa:
@fridis  Runtime$5.class does not seem to exist anymore

PR #4382, opened 2024-12-03
lib: add Sequence.count_by

### LINKED ISSUE #4379: AoC: Add feature to `Sequence` that count matches to a given predicate
It seems to be a common pattern to count the number of elements in a Sequence that match a certain predicate, an example is [this](https://github.com/fridis/fuzion_aoc/blob/0b4f1b560dbf29c6189992febe2adc5a85e89614/2024/02/part2.fz#L28-L30),

Might be nice to have a feature in `Sequence` that combines `filter` and `count`. Maybe overloading `count` by adding
```
Sequence.count(f T->bool) => (filter f).count
```
might do the trick. Or is there a better name? Do other language provide this?

michaellilltokiwa:
fixes #4379

### DIFF
--- a/modules/base/src/Sequence.fz
+++ b/modules/base/src/Sequence.fz
@@ -77,6 +77,16 @@ public Sequence(public T type) ref : property.equatable is
   => (map i32 (_ -> 1)).fold i32.sum
 
 
+  # count the number of elements in this Sequence that match the
+  # given predicate. Note that this typically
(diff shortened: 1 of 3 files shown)


michaellilltokiwa:
> Why `count_by` instead of `count` (overloaded)?

I liked the name better. It is more idiomatic. 
@tokiwa-software/developers What do the other think? count or count_by?

PR #4398, opened 2024-12-04
dfa: Experimental implementation of three possible optimizations

fridis:
- For certain clazzes, do not create ValueSet but, in case of a set, use one global set for all sets.

- For cetain clazzes, do not create separate Instances depending on context, but create only one single instance

- Do not create ValueSets of RefValue of the same clazz, but instead create RefValue of ValueSet, i.e., instead of
```
    {boxed(v1), boxed(v2)}
```
  create
```
    boxed({v1,v2})
```

All these can be enabled by static fields in DFA
```
  static boolean ONLY_ONE_VALUE_SET = false; 
  static boolean ONLY_ONE_INSTANCE  = false;
  static boolean NO_SET_OF_REFS     = false;
```


fridis:
The somewhat successful part of this was merged with #5333, so closing this.

PR #4399, opened 2024-12-04
Dfa optimizations only one valueset

fridis:


fridis:
Giviing up on this for now.

PR #4400, opened 2024-12-04
Dfa optimizations only one instance

fridis:


fridis:
This was essentially merged with #5333, so I am closing this.

PR #4401, opened 2024-12-04
Dfa optimizations no set of refs

fridis:


fridis:
giving up on this for now...

PR #4403, opened 2024-12-04
tests: add effects test

michaellilltokiwa:


fridis (approved):
nice, we needed this since long!

PR #4417, opened 2024-12-05
base/net: add `ip_address` feature

maxteufel:
To be extended with more features in the future…

### DIFF
--- a/modules/base/src/net/ip_address.fz
+++ b/modules/base/src/net/ip_address.fz
@@ -1,0 +1,50 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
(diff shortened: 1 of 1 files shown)


modules/base/src/net/ip_address.fz
```diff
@@ -0,0 +1,50 @@
...
+# in the case of an IPv4 address, this is the canonical mapping of that
+# address to an IPv6 address.
+#
+public ip_address (val u128) : property.equatable is
+
+  # is the stored IP address an IPv4 or IPv6 address?
+  #
+  public version ip_version =>
+    ipv4_prefix_netmask := u128 0x0000_0000_0000_0000 0x0000_ffff_0000_0000
+    if (val & ipv4_prefix_netmask) = (ipv4_prefix_netmask & ipv4_prefix_netmask)
```
simonvonhackewitz, line 36: bit late... but that does not make much sense to me

michaellilltokiwa: @simonvonhackewitz please explain why you think that? ;-)

simonvonhackewitz: Why `ipv4_prefix_netmask & ipv4_prefix_netmask`? Isn't that just the same as `ipv4_prefix_netmask`?

PR #4488, opened 2024-12-12
native: move fileio features to native

michaellilltokiwa:
see also: #4309


michaellilltokiwa:
@fridis No the intrinsics can go. But we need an io.out that uses `System.out.println` via fuzion.java.

PR #4492, opened 2024-12-12
lib: move type parameter from `io.buffered.reader` to `io.buffered`

fridis:
This avoids the need to repeat this type parameter on every call to to read something, e.g., instead of
```
  b := io.buffered
  s1 := b.read_string LM
  s2 := b.read_string LM
```
one can now write
```
  b := io.buffered LM
  s1 := b.read_string
  s2 := b.read_string
```
Also, this should enable change `io.buffed.writer` to no longer inherit `mutate` but to use the mutate provided via `(io.buffered M).writer`.  We for this to be convenient, we need to implement a default `stdout_buffer_mutate`, which I prefer to do only after DFA supports adding code to install default effects.


michaellilltokiwa:
@fridis This PR was failing only in the examples.

PR #4496, opened 2024-12-12
lib/io: don't return line for final newline in `io.buffered.read_lines`

### LINKED ISSUE #4486: `read_line` vs. `read_lines` behave differently wrt last line
Take this example
```
LM : mutate is
LM ! ()->
  io.stdin LM ! ()->
    if envir.args.count = 1 then

      for l := io.buffered.read_line LM
          i in 1..
      while l??
        say "using read_line: $i: $l"

    else

      for l in io.buffered.read_lines LM
          i in 1..
      do
        say "using read_lines: $i: $l"
```
using `read_line`, I get two lines
```
 > echo -e "abc\ndef" | ./build/bin/fz test.fz 
using read_line: 1: abc
using read_line: 2: def
```
and using `read_lines`, I get three lines
```
 > echo -e "abc\ndef" | ./build/bin/fz test.fz 1
using read_lines: 1: abc
using read_lines: 2: def
using read_lines: 3: 
```

simonvonhackewitz:
fix #4486

### DIFF
--- a/modules/base/src/io/buffered/reader.fz
+++ b/modules/base/src/io/buffered/reader.fz
@@ -244,11 +244,14 @@ public read_fully (LM type : mutate) array u8 ! (reader LM) =>
 # any trailing carriage returns (ASCII 13) from the resulting strings.
 #
 public read_delimiter_full (LM type : mutate, delim String, strip_cr bool) array String ! (reader LM) =>
-  (String.from_bytes (read_fully LM))
-    .split delim
-    .map (s -> if strip_cr && s.ends_with "\r" then s.substring 0 s.byte_length-1 else s)
-    .as_array
-
+  str := (String.from_bytes (read_fully LM))
+  if str.byte_length > 0
+    (str.ends_with "\n" ? str.substring 0 str.byte_length-1 : str)
+      .split delim
+      .map (s -> if strip_cr && s.ends_with "\r" then s.substring 0 s.byte_length-1 else s)
+      .as_array
+  else
+    []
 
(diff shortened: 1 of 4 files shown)


michaellilltokiwa:
@simonvonhackewitz please add test, to prove that the features now return the same result.

modules/base/src/io/buffered/reader.fz
```diff
@@ -244,11 +244,14 @@ public read_fully (LM type : mutate) array u8 ! (reader LM) =>
 # any trailing carriage returns (ASCII 13) from the resulting strings.
 #
 public read_delimiter_full (LM type : mutate, delim String, strip_cr bool) array String ! (reader LM) =>
-  (String.from_bytes (read_fully LM))
-    .split delim
-    .map (s -> if strip_cr && s.ends_with "\r" then s.substring 0 s.byte_length-1 else s)
-    .as_array
-
+  str := (String.from_bytes (read_fully LM))
+  if str.byte_length > 0
```
michaellilltokiwa, line 248: The `if` should not be necessary, I think?

simonvonhackewitz: The `if` is necessary so that an empty input/file does not return an empty line.

Output of the example code from the issue without the `if`:
```
❯ printf "" | fz /tmp/read_test.fz

❯ printf "" | fz /tmp/read_test.fz 1
using read_lines: 1: 

```

michaellilltokiwa (approved):

PR #4499, opened 2024-12-13
ast: replace `$MODULE` in error message with name of fum file

### LINKED ISSUE #4495: `$MODULE` in reported errors should be `base.fum` or similar
We currentyl get errors like this
```
$MODULE/pipes.fz:63:3: error 1: Call has an ambiguous result type since target of the call is a 'ref' type.
```
It would be more helpful if that was
```
base.fum:pipes.fz:63:3: error 1: Call has an ambiguous result type since target of the call is a 'ref' type.
```
or similar.

simonvonhackewitz:
fix #4495

### DIFF
--- a/src/dev/flang/fe/LibraryModule.java
+++ b/src/dev/flang/fe/LibraryModule.java
@@ -2382,7 +2382,7 @@ SourceFile
             var bb = sourceFileBytes(at);
             var ba = new byte[bb.limit()]; // NYI: Would be better if SourceFile could use bb directly.
             bb.get(0, ba);
-            sf = new SourceFile(Path.of(sourceFileName(at)), ba);
+            sf = new SourceFile(Path.of("{" + name() + FuzionConstants.MODULE_FILE_SUFFIX + "}").resolve(Path.of(sourceFileName(at))), ba);
             _sourceFiles.set(i, sf);
           }
         return new SourceRange(sf, pos - sourceFileBytesPos(at), posEnd - sourceFileBytesPos(at));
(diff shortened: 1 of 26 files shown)


src/dev/flang/fe/LibraryModule.java
```diff
@@ -2382,7 +2382,7 @@ else if (pos == 0)
             var bb = sourceFileBytes(at);
             var ba = new byte[bb.limit()]; // NYI: Would be better if SourceFile could use bb directly.
             bb.get(0, ba);
-            sf = new SourceFile(Path.of(sourceFileName(at)), ba);
+            sf = new SourceFile(Path.of(name() + ".fum").resolve(Path.of(sourceFileName(at))), ba);
```
fridis, line 2385: Do not use `".fum"` directly, but `FuzionConstants.MODULE_FILE_SUFFIX` (also at other locations, maybe fix that in another PR).

fridis (changes requested):
two small suggestions.

fridis (commented):
You also have to update the test output. 

Not very happy about `base.fum/String.fz`, which looks like a legal path.  It might be better to use `$base/String.fz`,  `{base.fum}/String.fz` or similar.

src/dev/flang/fe/LibraryModule.java
simonvonhackewitz, line 2385:
```suggestion
            sf = new SourceFile(Path.of("module " + name() + ": " + Path.of(sourceFileName(at)).toString()), ba);
```
How about this so it doesn't look like a legal path? 

The output is
```
Original feature declared at module base: String.fz:34:10:
```

Does not seem to be the proper way of using Path though.

PR #4517, opened 2024-12-16
dfa: override accessedClazz

michaellilltokiwa:
return NO_CLAZZ if we are assigning unit type or target of assignment is unit.


fridis (approved):
Looks good. 

What is unclear to me: What effect does this have? I think it might make the back-ends easier or avoid some dead code, but I'm not sure from the patch.

michaellilltokiwa:
@fridis Yes, it avoids some unneeded assigns. I realized this when change the Interpreter to actually use the DFA-FUIR and not the normal FUIR. This change then triggered some problems and this patch avoids those problems for the interpreter. Has no observable effect in JVM/C-Backend as far as I know.

PR #4519, opened 2024-12-16
fuir: remove declarationPos

michaellilltokiwa:
was only needed by interpreter.


fridis (approved):
Good. 

We might need something similar for better error messages, but we can add exactly what we need for all backends if we actually do.

PR #4521, opened 2024-12-17
lib: change `Const_String` to `const_string`

michaellilltokiwa:
The fields `clazz_Const_String_utf8_data` outer clazz was a `ref`, this was inconsistent and is the main reason for this change. I also think it makes more sense for the const_string to be a value instead of a reference.


fridis (approved):
good!

PR #4524, opened 2024-12-17
lib: change impl of Sequence.concat for finite Sequences

michaellilltokiwa:


modules/base/src/Sequence.fz
```diff
@@ -271,11 +271,19 @@ public Sequence(public T type) ref : property.equatable is
...
   # by all the elements of s
   #
-  public concat (s Sequence T) Sequence T => as_list.concat_list s.as_list
+  # NYI: PERFORMANCE: time complexity may be as bad as O(n²), see #4465
+  #
+  public concat (s Sequence T) Sequence T =>
+    if finite = trit.yes && s.finite = trit.yes
+      as_array.concat_array s.as_array
+    else
+      as_list.concat_list s.as_list
```
fridis, line 280: If you do something like 
```
  for i in 0..n
      s := [i], s ++ [i]
  else
    say s
```
Your code will create arrays of size 0, 1, 2, 3, ..., n, so we still have quadratic performance.

fridis (changes requested):
I am happy with adding `array.concat_array` and adding the `NYI`-comment. 

However, the change to `Sequence.concat` IMHO does not improve the situation. 

Essentially, we will need something like Java's `ArrayList` that is backed by a `ref` to a mutable array that get doubled in size when it is full and that gets cloned in case the array holds more elements than the list or when the array is modified repeatedly, ie., 
```
  s Sequence i32 := []
  s1 := s ++ [1] ++ [2]  ++ [3] # backed by mutable array of size 4
  a0 := s1                      # backed by same mutable array as s1
  a1 := a0 ++ [4]               # still same mutable array
  a2 := s1 ++ [5]               # new mutable array with size 8
  b0 := s1                      # backed by same mutable array as s1, a0, a1
  b1 := b0 ++ [6]               # new mutable array with size 4
  b2 := b1 ++ [7                # new mutable array with size 8
  say a2  # 1,2,3,4,5
  say b2  # 1,2,3,6,7
```

PR #4526, opened 2024-12-17
add_simple_test: add comment explaining how to end input

michaellilltokiwa:


fridis (changes requested):
Please create an issue that we must provide a way to read `io.stdin` and react to `return` and reference this issue here.

PR #4530, opened 2024-12-18
lib/tests: use pipes operator in some places

michaellilltokiwa:


fridis (approved):
I start to like the pipes more and more, but there is still a bit limitation due to `<|` being left-associative, so you cannot write `say <| (.replace "e" "a") <| "Hello"`, but we need `()` as in `say <| ((.replace "e" "a") <| "Hello")`

PR #4568, opened 2025-01-07
actions: fix linux workflow

michaellilltokiwa:
also added shellcheck to apple workflow


michaellilltokiwa:
@maxteufel I should have described that... The linux workflow includes `make shellcheck`, hence it failed because of some warnings in the scripts. This PR fixes this and also adds shellcheck to the macOS workflow.

PR #4635, opened 2025-01-20
lib: add features like `codepoint.is_ascii_digit` for existing sets

### LINKED ISSUE #4547: AoC: `codepoint.is_ascii_digit` and similar features would be helpful
Found this during [Dec 21 Advent-of-Code](https://github.com/fridis/fuzion_aoc/blob/main/2024/21/part2.fz): There are a number of  codepoint sets defined in `codepoint`, but code to check if a codepoint is part of a set is clumsy: `codepoint.ascii_digit.contains c.val`. This could be `c.is_ascii_digit`.

simonvonhackewitz:
fix #4547

### DIFF
--- a/modules/base/src/codepoint.fz
+++ b/modules/base/src/codepoint.fz
@@ -53,6 +53,46 @@ is
   #
   public redef is_ascii => codepoint.ascii.contains val
 
+  # is this an ASCII digit i.e. 0-9
+  #
+  public is_ascii_digit => codepoint.ascii_digit.contains val
+
+  # is this an ASCII letter i.e. A-Z or a-z
+  #
+  public is_ascii_letter => codepoint.latin_alphabet.contains val
+
+  # is this an uppercase ASCII letter i.e. A-Z
+  #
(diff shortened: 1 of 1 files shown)


modules/base/src/codepoint.fz
```diff
@@ -53,6 +53,46 @@ is
...
+  #
+  public is_uppercase_ascii_letter => codepoint.A_to_Z.contains val
+
+  # is this a lowercase ASCII letter i.e. a-z
+  #
+  public is_lowercase_ascii_letter => codepoint.a_to_z.contains val
+
+  # is this codepoint encoded in one byte
+  #
+  public is_one_byte => codepoint.utf8_encoded_in_one_byte.contains val
```
michaellilltokiwa, line 74: Maybe instead of `is_one_byte`, etc. just `byte_count`

simonvonhackewitz: there is `byte_length` (inherited from `String`), so I remove the features `is_one_byte` etc.

PR #4638, opened 2025-01-20
lib: let `mutable_tree_map` inherit form `Mutable_Map` and add features

### LINKED ISSUE #4539: AoC: Add `mutable_tree_map.get_or_put`
During today's Advent-of-Code [solution](https://github.com/fridis/fuzion_aoc/blob/main/2024/19/part2.fz), I wrote code like this using a `mutable_tree_map` in field `cacheCnt`:
```
      match cacheCnt.get s
        v i64 => v
        nil   => res := s="" ? 1 : towels.filter s.starts_with
                                         .map s.suffix
                                         .map count |> sum
                 cacheCnt.put s res
                 res
```
It would be nice if one could instead write
```
      cacheCnt.get_or_put s ()->
        s="" ? 1 : towels.filter s.starts_with
                         .map s.suffix
                         .map count |> sum
```
There are some more issues `mutable_tree_map`, in particular

- it should not inherit from `Map`, but have a feature `as_map`, similar to mutable arrays
- there should be some common parent feature for mutable maps
- there should be other implementations for that parent like `mutable_hash_map`

simonvonhackewitz:
fix #4539

### DIFF
--- a/modules/base/src/container/Map.fz
+++ b/modules/base/src/container/Map.fz
@@ -70,6 +70,13 @@ public Map(public K type : property.equatable, public V type) ref is
       r
 
 
+  # create a mutable map from this
+  #
+  # NYI: maybe add default implementation when there is a naive implementation only requiring property.equatable
+  #
+  public as_mutable_map(LM type : mutate) container.Mutable_Map K V => abstract
+
+
   # initializer for this map
   #
   # every map implementation needs to implement this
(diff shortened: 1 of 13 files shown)


michaellilltokiwa:
I think you should also remove the inheritance of Map and make mutable_tree_map inherit from sth. like Mutable_Map and put those features there.

michaellilltokiwa:
I just realized there already is a `Mutable_Map` interface...

PR #4644, opened 2025-01-21
lib: add `Sequence.scan` that takes a function

### LINKED ISSUE #4546: AoC: `scanf` missing and error referring to `--UNDEFINED--`
While playing around with dec 20 solution, tried to use different versions of `Sequence.scan` just to find that there is none that takes a function, only a monoid is allowed. 

But passing a function `(+)` results in an error referring to `--UNDEFINED--`:
```
 > $FUZION/bin/fz -e 'say <| (1 :: id).scan (+)'

command line:1:23: error 1: Incompatible types when passing argument in a call
say <| (1 :: id).scan (+)

Actual type for argument #1 'm' does not match expected type.
In call to          : 'list.scan'
expected formal type: 'Monoid i32'
actual type found   : '--UNDEFINED--'
assignable to       : '--UNDEFINED--'
for value assigned  : '(+)'
To solve this, you could change the type of the target 'm' to '--UNDEFINED--' or convert the type of the assigned value to 'Monoid i32'.

one error.

```

simonvonhackewitz:
fix #4546

### DIFF
--- a/modules/base/src/Sequence.fz
+++ b/modules/base/src/Sequence.fz
@@ -457,6 +457,15 @@ public Sequence(public T type) ref : property.equatable is
   public scan (m Monoid T) => as_list.scan m.e m
 
 
+  # map this Sequence to a list that contains the result of folding
+  # all prefixes using the given function and initial value.
+  #
+  # e.g., for a Sequence s of i32, s.scan (+) 0 creates a list of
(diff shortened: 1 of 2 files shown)


michaellilltokiwa (approved):
:+1:

PR #4653, opened 2025-01-22
lib: avoid `as_list` in `Sequence.as_array` if array backed

fridis:
IMHO feels better like this and should improve performance, but I did not measure thatn.


michaellilltokiwa (approved):

michaellilltokiwa (commented):
But we also have this in array:
```
  public fixed redef as_array =>
    array.this
```

PR #4658, opened 2025-01-23
ci: enable ci-energy-estimation

michaellilltokiwa:
> Hey @michaellilltokiwa ,
>
> we add first beta support for macOS!
>
> If you are still keen on trying out we would really like you as beta tester and feedback provier. Any interest ? :)


simonvonhackewitz:
So our tests are running in San Jose?

PR #4665, opened 2025-01-23
fuir: do not create outerRef clazz when lookupDone

michaellilltokiwa:


fridis (approved):
Would be good, if you have an example where this makes a difference, to add a short comment when this happens.

michaellilltokiwa:
@fridis
When serializing FUIR we currently do not have the info if outerRef was created or not. This idea was basically just to prevent creation of new clazzes in this case.

PR #4668, opened 2025-01-23
Experimental addition to fix #4662, can be get rid of stack cleanup for calls that always result in void?

### LINKED ISSUE #4662: Stack cleanup in case of void in `AbstractInterpreter.processCode` for match
#4651 instroduces Fuzion code that provokes a case where stack cleanup is needed for a `Match` when using the JVM backend. I will implement a workaround, but this should be checked.

fridis:
Trying to find when a Call returns a value even though it always results in void.


fridis:
merged into #4667.

PR #4672, opened 2025-01-24
lib: add `linear_effect`

### LINKED ISSUE #4218: `mutate` instance can be re-entered, permitting mutation without visible effect

michaellilltokiwa:
A linear effect checks at runtime if it has already been instated and panics if this is the case.

fixes #4218


modules/base/src/effect.fz
```diff
@@ -378,6 +379,12 @@ public effect is
   static_finally => finally
 
 
+  # trigger runtime check if effect is
+  # already depleted any may not be used
```
fridis, line 383:
```suggestion
  # already depleted and may not be used any more
```

fridis (approved):
Just a small suggestion to improve a comment.

PR #4675, opened 2025-01-24
lib: orderable provide implementation for type.equality

### LINKED ISSUE #4551: AoC: `property.orderable` should provide implementation of `type.equality`
Found this during soling [23 Dec Advent-of-Code](https://github.com/fridis/fuzion_aoc/blob/main/2024/23/part2.fz): Since a total order implies an equality relation, `property.orderable` should implement `type.equality` as follows
```
  redef property.orderable.type.equality(a,b property.orderable.this) 
    post
      result = (b <= a <= b)
  =>
    a <= b <= a
```

michaellilltokiwa:
fixes #4551

### DIFF
--- a/modules/base/src/property/orderable.fz
+++ b/modules/base/src/property/orderable.fz
@@ -46,7 +46,7 @@ is
   #
   # This relation must be
   #
-  #  - reflexive (equality a b),
+  #  - reflexive (equality a a),
   #  - symmetric (equality a b = equality b a), and
   #  - transitive ((equality a b && equality b c) : equality a c).
   #
@@ -55,7 +55,7 @@ is
   #
   public redef type.equality(a,b property.orderable.this)
 
-  # NYI: BUG:
+  # NYI: BUG: see #4676
   #
   # String.fz:73:27: error 1: Incompatible types when passing argument in a call
   #   public fixed redef type.equality(a, b String) =>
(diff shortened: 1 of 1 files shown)


modules/base/src/property/orderable.fz
```diff
@@ -40,3 +40,36 @@ public orderable : partially_orderable
 */
 
 is
+
+
+  # equality implements the default equality relation for values of this type.
+  #
+  # This relation must be
+  #
+  #  - reflexive (equality a b),
```
simonvonhackewitz, line 49:
```suggestion
  #  - reflexive (equality a a),
```

modules/base/src/property/orderable.fz
```diff
@@ -40,3 +40,36 @@ public orderable : partially_orderable
...
+  #  - symmetric (equality a b = equality b a), and
+  #  - transitive ((equality a b && equality b c) : equality a c).
```
simonvonhackewitz, line 51: I'm a bit confused by the colon and would find `=>` more intuitive. Though it might lead to confusion with the fuzion syntax...

modules/base/src/property/orderable.fz
```diff
@@ -40,3 +40,36 @@ public orderable : partially_orderable
...
+  #
+  # result is true iff 'a' is considered to represent the same abstract value
+  # as 'b'.
+  #
+  public redef type.equality(a,b property.orderable.this)
+
+  # NYI: BUG:
```
simonvonhackewitz, line 58:
```suggestion
  # NYI: BUG: see #4676
```

PR #4679, opened 2025-01-27
dfa: compare_and_* mark complete value as read

### LINKED ISSUE #2131: Fields that are never read but used in compare-and-swap must not be removed
The example from #2129 uses a field `l` that is never read in a value that is used in a `concur.atomic`. Since `compare_and_swap` and `compare_and_set` behave differently depending on the value of such a field that is otherwise never read, we should not remove this field here.

michaellilltokiwa:
fixes #2131


michaellilltokiwa:
@maxteufel Ugh, yes, sorry.

PR #4691, opened 2025-01-28
tests: test starting process using unicode

### LINKED ISSUE #4565: starting process, fileio (creating file, directory etc.) needs tests using unicode
We probably pass down utf-8 bytes to the intrinics/native features but I'm almost sure that this does not work for all cases.

simonvonhackewitz:
part of #4565


michaellilltokiwa:
@simonvonhackewitz I propose this for finding out what is going wrong here:
- Delete every test but the process test (for faster iteration).
- When running the test first print all environment variables and their value.
- Print all the files in the directory where the executable is meant to be.

PR #4700, opened 2025-01-29
makefile: add debugging info to html when generating api docs

simonvonhackewitz:
I think this would make debugging easier when generating api docs


fridis (approved):
I think we could even add *visible* version information in a small font on every page of the API docs, maybe the Fuzion version string and the date of creation, similar to the output of `fz -version`.

PR #4701, opened 2025-01-30
fe: add `ast.Select`

michaellilltokiwa:
Idea is that Select contains all the logic of resolving a select-clause, hence this logic was removed from Call.

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2359,6 +2359,13 @@ public class AstErrors extends ANY
           + "To solve this, remove " + skw("...") + " after the highlighted error.");
   }
 
+  public static void selectIsNoType(SourcePosition pos)
+  {
+    error(pos,
+          "Select clause is not a valid type",
+          "To solve, this specify a valid type.");
+  }
+
 }
 
 /* end of file */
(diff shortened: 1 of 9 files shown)


fridis (commented):
I though of having `Select` created separately from `ParseCall` and then later turning the `Select` into either a `Call` when it is used as `tuple.1` (so it is ok for this to inherit from `Call`, but maybe it would be better to create the Call during `resolveTypes` and really replace the select then) or to remove the `Select` if used as `tuple.values.1`. 

Hope this makes the code easier / clearer.

michaellilltokiwa:
> I though of having `Select` created separately from `ParseCall` and then later turning the `Select` into either a `Call` when it is used as `tuple.1` (so it is ok for this to inherit from `Call`, but maybe it would be better to create the Call during `resolveTypes` and really replace the select then) or to remove the `Select` if used as `tuple.values.1`.
> 
> Hope this makes the code easier / clearer.

If I understand your suggestion correctly this would mean we set the select value on a Call after resolving it? (if it turns out to be necessary)

michaellilltokiwa:
@fridis building on this PR is  #4704

fridis (approved):

PR #4702, opened 2025-01-30
ast: remove possibility to specify types in tuple destructuring

simonvonhackewitz:
forbids this
```
❯ FUZION_DISABLE_ANSI_ESCAPES=true fz -e '(a i32, b i32) := (1, 2); _:=a; _:=b'

command line:1:1: error 1: Failed to infer open type parameter type from actual argument.
(a i32, b i32) := (1, 2); _:=a; _:=b
^
Type of second actual argument could not be inferred at command line:1:2:
(a i32, b i32) := (1, 2); _:=a; _:=b
-^


command line:1:30: error 2: Could not find called feature
(a i32, b i32) := (1, 2); _:=a; _:=b
-----------------------------^
Feature not found: 'a' (no arguments)
Target feature: 'universe'
In call: 'a'


command line:1:36: error 3: Could not find called feature
(a i32, b i32) := (1, 2); _:=a; _:=b
-----------------------------------^
Feature not found: 'b' (no arguments)
Target feature: 'universe'
In call: 'b'

3 errors.
```

### DIFF
--- a/src/dev/flang/ast/Destructure.java
+++ b/src/dev/flang/ast/Destructure.java
@@ -218,13 +218,6 @@ public class Destructure extends ExprWithPos
       {
         var newF = (Feature) fields.next();
 
-        var a = newF._returnType.functionReturnType().resolve(res, context);
-
-        if (a != Types.t_UNDEFINED && !(!newF._returnType.isConstructorType() && a.isAssignableFrom(t)) )
-          {
-            AstErrors.incompatibleType(newF._returnType.posOrNull(), "in tuple destructuring", "", "", newF._returnType.functionReturnType(), assign, t, context);
-          }
-
         newF._returnType = new FunctionReturnType(t);
         assign = new Assign(res, pos(), newF, call_f, context);
       }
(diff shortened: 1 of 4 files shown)


michaellilltokiwa:
@simonvonhackewitz cool, but please add this example you give in the comment as a simple test.

src/dev/flang/ast/Destructure.java
```diff
@@ -217,6 +217,14 @@ private void addAssign(Resolution res,
     if (fields != null && fields.hasNext())
       {
         var newF = (Feature) fields.next();
+
+        var a = newF._returnType.functionReturnType().resolve(res, context);
+
+        if (a != Types.t_UNDEFINED && !(!newF._returnType.isConstructorType() && a.isAssignableFrom(t)) )
+          {
```
fridis, line 224: This looks dangerous and too complex. `newF._returnType.functionReturnType()` should crash if `newF._returnType.isConstructorType()`, which is checked afterwards but with a `!(!(` that asks for restructuring.

fridis (changes requested):
I suggest a completely different approach: Please check our examples, tests, tutorial, idioms for cases that use the grammar rule for `destructrDcl`. If this is not used anywhere, or if those uses are not necessary, maybe we can just throw this rule away and not allow any types here. 
If, howerver, we see we need something like this, we first needs positive and negative tests for all forms at `formArgs` could have and check if they work or produce a good error.

michaellilltokiwa (approved):

PR #4704, opened 2025-01-30
fe/parser: fix short handle select parsing

### LINKED ISSUE #2913: nested tuple with short hand access gives weird error
```
/home/not_synced/fuzion (2024-04-17T11-21-11+02-00)130$ cat ~/playground/test.fz
ex =>

   x := (3.4,(3,"asdf"))
   say x.values.1.values.1
   say x.1.1
/home/not_synced/fuzion (2024-04-17T11-21-11+02-00)$ fz ~/playground/test.fz 

/home/sam/playground/test.fz:5:8: error 1: Selector must be in the range of 0..1 for 2 actual type parameters given for the open type parameter type
   say x.1.1
-------^
In call to 'tuple.values'
Selected variant 'values.11'
Number of actual type parameters: 2
Actual type parameters: 'f64, tuple i32 String'

one error.

```

### LINKED ISSUE #4408: short handle select gives syntax error.
```
ex =>

  t => (1,2)

  say (id t).1
```
```
/home/sam/playground/test.fz:5:14: error 1: Syntax error: expected identifier name, infix/prefix/postfix operator, 'ternary ? :', 'index' or 'set' name, found numeric literal '1'
  say (id t).1

While parsing: name, parse stack: name (twice), call (twice), callTail, term, opTail, opExpr, operatorExpr (twice), actualSpace, actualSpaces, actualArgs, call (twice), callOrFeatOrThis, term, opTail, opExpr, operatorExpr (twice), expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), unit

one error.
```
this works
```
ex =>

  t => (1,2)

  say (id t).values.1
```

michaellilltokiwa:
fixes #2913
fixes #4408


src/dev/flang/parser/Parser.java
```diff
@@ -1436,32 +1436,19 @@ Expr call(Expr target)
 call        : name actuals callTail
             ;
 actuals     : actualArgs
-            | dot NUM_LITERAL
+            | dot select callTail
```
fridis, line 1439: callTail is already after `actuals`, so this rule (`dot select callTail`) should be removed completely.

tests/select_clause_negative/Makefile
```diff
@@ -0,0 +1,25 @@
...
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test Makefile
+#
+# -----------------------------------------------------------------------
+
+override NAME = select_clause_negative
+include ../simple.mk
```
fridis, line 25: Better do this: 
```suggestion
include ../simple_and_negative.mk
```
To make sure that none of the errors gets lost accidentally when updating expected error output.

fridis (approved):
Nice! I particularly like the simplification of the code in the grammar and the parser!

Just a minor suggestion to use `simple_and_negative.mk`

PR #4708, opened 2025-01-31
tests: test file io using unicode

### LINKED ISSUE #4565: starting process, fileio (creating file, directory etc.) needs tests using unicode

simonvonhackewitz:
part of #4565


michaellilltokiwa:
:+1:

michaellilltokiwa (approved):

PR #4711, opened 2025-01-31
Makefile: Add missing dependency and fix building docs for work != fuzion dir

fridis:
The fz_cmd module seems to be required and a `cd $(FZ_SRC)` is needed to call git show...`.


simonvonhackewitz:
I've opened #4710 to revert the change that adds version info as a html comment from the makefile. If it get's merged this PR would no longer be necessary.

michaellilltokiwa:
superseded by #4710

PR #4712, opened 2025-01-31
ast: Fix #197 by fixing order of `applyTypePars` for call resul type

### LINKED ISSUE #197: Unexpected: Incompatible types in assignment
this example:
```fuzion
ex =>
  my_array<T> is
    my_map<B> my_array<B> is
      my_array<B>
    f (ms my_array<Monoid<T>>) my_array<T> is
      ms.my_map<T>
```
yields:

```
...fz:6:10: error 1: Incompatible types in assignment
      ms.my_map<T>
---------^
assignment to field : 'ex.my_array.f.result'
expected formal type: 'ex.my_array<ex.my_array.T>'
actual type found   : 'ex.my_array<Monoid<ex.my_array.T>>'
assignable to       : 'ex.my_array<Monoid<ex.my_array.T>>'
for value assigned  : '{
  f.this.ms.my_map<ex.my_array.T>
}'
To solve this, you could change the type of the target 'ex.my_array.f.result' to 'ex.my_array<Monoid<ex.my_array.T>>' or convert the type of the assigned value to 'ex.my_array<ex.my_array.T>'.

one error.
```

fridis:
The problem is that a call `(a X).(b Y)` where `a` and `b` are generic needs to apply the actual generics to `b`'s forma result type, and these could come from  the outer type `a X` and from the call `b Y`.  This, unfortunately, was done in the wrong order: It has to be done first taking the formal result type of `b` and replacing type parameters in there by the outer  type `a X`, and then  applying the type parameters given in the call `b Y`.

When done the other way around, i.e., first applying the type parameters given to the call to `b Y` and then those in the outer type `a X`, the result is wrong if the calls is done within `a` and `Y` contains type parameters of `a`.

To be able to do this, the code to determine the call's formal result type, which contained `t.applyTypePars(cf, _generics)`, first had to be moved into `setActualResultType`. Then the call to `applyTypePars` was moved at the right position, i.e., `var t2 = ...`. 

Finally, some cleanup was done, the call to `setActualResultType` in `ParsedCall.resolveImplicitSelect` seems not necessary, so it was removed. 

NOTE: Best reviewed with `Hide whitespace` in the diff view since indentation of quite a lot of code changed.

### DIFF
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -1460,7 +1460,7 @@ public class Call extends AbstractCall
    * unknown and urgent is false, Types.t_ERROR in case of an error (e.g. urgent
    * is true and the formal result type is still unknown).
    */
-  protected AbstractType setActualResultType(Resolution res, Context context, boolean urgent)
+  private AbstractType setActualResultType(Resolution res, Context context, boolean urgent)
   {
     AbstractType frmlT;
     if (isTailRecursive(context.outerFeature()) || _recursiveResolveType)
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -1440,8 +1440,6 @@ public class Call extends AbstractCall
   }
 
 
-
-
   /**
    * Helper function for resolveTypes to determine the static result type of
    * this call.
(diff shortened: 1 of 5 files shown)


michaellilltokiwa:
Cool, to have this fixed! Might affect some other open issues.

michaellilltokiwa (approved):

PR #4722, opened 2025-02-04
docs: reimplement generation of tree structure for navigation on left side

### LINKED ISSUE #4707: API documentation tree display strange
The tree display
Image
looks weird, should look more like the one for files (which also is not perfect, there are small gabs between the bars):
Image

Also, the use of the hamburger-icon Image is confusing. This Icon is usually used for a menu that pops up when clicking, not for an area of the display that is folded away. Should better be something like Image (as in LibreOffice), but maybe you find something better still.

simonvonhackewitz:
fix #4707

Screenshot_20250204_130732


michaellilltokiwa (commented):
Code seems to be nicer than before. One small nit pick.

src/dev/flang/tools/docs/Html.java
```diff
@@ -812,48 +810,78 @@ private String typeArgAsString(AbstractFeature f)
...
+
+        var innerFeatures = lm.declaredFeatures(f).values().stream()
+                              .filter(ft -> ft.definesType()
+                                            && ft.visibility().typeVisibility() == Visi.PUB)
+                              .sorted(Comparator.comparing(ft -> ft.featureName().baseName(), String.CASE_INSENSITIVE_ORDER))
+                              .collect(Collectors.toList());
+
+        // addition to the tree structure prefix for current feature: universe / normal element / last element
+        var featPrfx = f.isUniverse() ? ""
+                                      : iter.hasNext() ? "├─&nbsp;"      : "└─&nbsp;";
```
michaellilltokiwa, line 849: use <span class='width-1/2/3'></span> instead of nbsp

PR #4723, opened 2025-02-04
native: add `fzE_errno`

michaellilltokiwa:
This is for getting the last error number. For now only used in process.start.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1438,7 +1438,7 @@ ifeq ($(OS),Windows_NT)
 	-lMswsock -lAdvApi32 -lWs2_32
 	touch $(FUZION_RT)
 else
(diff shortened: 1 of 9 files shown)


michaellilltokiwa:
> I would highly advise caution with this. The call to `errno` returns the last error code produced by any C standard library code. If our backend happens to insert an unintentional other call to the C standard library, for example to `alloc` to `memcpy` in between the call of the function that should match this error code, and the call to `errno`, the result is wrong.

Good point, I revised the PR, i.e. I introduced a thread local `last_error` that specifically holds the error we want to pass from runtime to the user. I had to enable clang option `-fPIC` for this.

PR #4731, opened 2025-02-05
ast: suppress subsequent errors for error types

### LINKED ISSUE #4694: Too many subsequent errors in small example
Check this, in particular the `**errorr**` output should be suppressed 
```
 > ./build/bin/fz -e '0 |> x,y->container.expanding.env.put' 

command line:1:6: error 1: Wrong number of arguments in lambda expression
0 |> x,y->container.expanding.env.put

Lambda expression has 2 arguments while the target type expects one argument.
Arguments of lambda expression: 'x', 'y'
Expected function type: Unary --UNDEFINED-- i32
To solve this, remove one argument from the list 'x', 'y' before the '->' of the lambda expression.


command line:1:21: error 2: Wrong number of type parameters
0 |> x,y->container.expanding.env.put

Wrong number of actual type parameters in type:
Type: container.expanding
expected one generic argument for 'T'
found none.


command line:1:6: error 3: Wrong number of type parameters
0 |> x,y->container.expanding.env.put

Wrong number of actual type parameters in call:
Called feature: Unary
expected 2 generic arguments for 'T, U'
found 3: '**error**, i32, **error**'.


command line:1:3: error 4: Failed to infer actual type parameters
0 |> x,y->container.expanding.env.put

In call to 'infix |>', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'A, R'
Type inference failed for one type parameter 'R'


command line:1:3: error 5: Constructor code should result in type 'unit'
0 |> x,y->container.expanding.env.put

Type returned by this constructor's implementation is '--UNDEFINED--'
To solve this, you could turn this constructor into a routine by adding a matching result type compatible to '--UNDEFINED--' or by using '=>' instead of 'is' to infer the result type from the result expression.
Alternatively, you could explicitly return 'unit' as the last expression or explicitly ignore the result of the last expression by an assignment '_ := <expression>'.

5 errors.

```

simonvonhackewitz:
fix #4694

I think the remaining subsequent error is justified and should not be suppressed, but correct me if I'm wrong
```
❯ FUZION_DISABLE_ANSI_ESCAPES=true fz -e '0 |> x,y->container.expanding.env.put'

command line:1:6: error 1: Wrong number of arguments in lambda expression
0 |> x,y->container.expanding.env.put
-----^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Lambda expression has 2 arguments while the target type expects one argument.
Arguments of lambda expression: 'x', 'y'
Expected function type: Unary --UNDEFINED-- i32
To solve this, remove one argument from the list 'x', 'y' before the '->' of the lambda expression.


command line:1:21: error 2: Wrong number of type parameters
0 |> x,y->container.expanding.env.put
--------------------^^^^^^^^^
Wrong number of actual type parameters in type:
Type: container.expanding
expected one generic argument for 'T'
found none.

2 errors.
```

Also I'm not sure if it is a good idea to just hide the error in `AbstractType.java` if other errors were reported in order to suppress this subsequent error
```
[...]
command line:1:3: error 3: Failed to infer actual type parameters
0 |> x,y->container.expanding.env.put
--^^
In call to 'infix |>', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'A, R'
Type inference failed for one type parameter 'R'

3 errors.
```

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -2192,7 +2192,10 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
 
         if (a == Types.t_UNDEFINED)
           {
-            AstErrors.failedToInferActualGeneric(pos, called, new List<>(f));
+            if (!Errors.any())
+              {
+                AstErrors.failedToInferActualGeneric(pos, called, new List<>(f));
+              }
           }
         else
           {
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -2192,10 +2192,7 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
 
         if (a == Types.t_UNDEFINED)
           {
-            if (!Errors.any())
-              {
-                AstErrors.failedToInferActualGeneric(pos, called, new List<>(f));
-              }
+            AstErrors.failedToInferActualGeneric(pos, called, new List<>(f));
           }
         else
           {
--- a/src/dev/flang/ast/Call.java
+++ b/src/dev/flang/ast/Call.java
@@ -316,7 +316,7 @@ public class Call extends AbstractCall
   {
     this(pos, target, calledFeature.featureName().baseName(), -1, generics, actuals, calledFeature, type);
     if (PRECONDITIONS) check
-      (calledFeature.generics().sizeMatches(generics));
+      (calledFeature.generics().sizeMatches(generics) || generics.contains(Types.t_ERROR));
   }
 
 
--- a/src/dev/flang/ast/FormalGenerics.java
+++ b/src/dev/flang/ast/FormalGenerics.java
@@ -140,7 +140,7 @@ public class FormalGenerics extends ANY
     if (PRECONDITIONS) require
       (Errors.any() || !actualGenerics.contains(Types.t_ERROR));
 
-    var result = sizeMatches(actualGenerics);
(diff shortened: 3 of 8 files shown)


michaellilltokiwa (changes requested):
Add a regression test, please.

simonvonhackewitz:
output looks like this now
```
❯ FUZION_DISABLE_ANSI_ESCAPES=true fz -e '0 |> x,y->container.expanding.env.put'

command line:1:6: error 1: Wrong number of arguments in lambda expression
0 |> x,y->container.expanding.env.put
-----^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Lambda expression has 2 arguments while the target type expects one argument.
Arguments of lambda expression: 'x', 'y'
Expected function type: Unary --UNDEFINED-- i32
To solve this, remove one argument from the list 'x', 'y' before the '->' of the lambda expression.


command line:1:21: error 2: Wrong number of type parameters
0 |> x,y->container.expanding.env.put
--------------------^^^^^^^^^
Wrong number of actual type parameters in type:
Type: container.expanding
expected one generic argument for 'T'
found none.


command line:1:3: error 3: Failed to infer actual type parameters
0 |> x,y->container.expanding.env.put
--^^
In call to 'infix |>', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'A, R'
Type inference failed for one type parameter 'R'

3 errors.
```

michaellilltokiwa (approved):

PR #4732, opened 2025-02-05
fe: fix inheritance for contracts

michaellilltokiwa:
fixes https://github.com/tokiwa-software/fuzion/issues/3051

Root of the problem was wrong logic in `requiresPreConditionsFeature`, was
```
var fc = f.contract();

    return fc._hasPre != null &&
      (!fc._declared_preconditions.isEmpty() || !f._inheritedPre.isEmpty());
```
instead of
```
    return !f.contract()._declared_preconditions.isEmpty() ||
      f._inheritedPre.stream().anyMatch(x -> hasPreConditionsFeature(x));
```

- Had to change naming logic because of one name clash in different modules.
- several fixes/changes in tests/lib


fridis (approved):
Very nice!

PR #4733, opened 2025-02-05
lib: Add `Sequence.orderable`

fridis:
This makes it easy to use `Sequence`s as keys in maps or similar.


michaellilltokiwa:
@maxteufel I think this is about ordering the Sequence as a whole, not the elements in them.

PR #4739, opened 2025-02-06
tests: fix `tuple_negative`, remove unsupported test cases

### LINKED ISSUE #89: tuples syntax sugar result in too few or wrong error messages
This is shown be tests/tuple_negative which produces wrong error message (Incompatible type) for case 31 through 33 and no error for cases 34 and higher.

simonvonhackewitz:
fix #89


michaellilltokiwa (approved):
Cool.
Noted that some tests are deleted but they or similar test cases are in test/tuple.

PR #4748, opened 2025-02-07
ast: cleanup: use "must not" instead of "cannot" for rules or restrictions

### LINKED ISSUE #4742: Error messages, cannot vs. must not
I think in some places `must not` should be used over cannot, e.g. `Choice cannot refer to an outer value type`

See also: https://www.rfc-editor.org/rfc/rfc2119

simonvonhackewitz:
fix #4742

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -1026,7 +1026,7 @@ public class AstErrors extends ANY
 
   static void cannotRedefineChoice(AbstractFeature f, AbstractFeature existing)
   {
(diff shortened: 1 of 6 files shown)


michaellilltokiwa (approved):
Aligns better with my feeling for language.
@tokiwa-software/developers Do you feel the same or do we need a native speaker to look at this?

simonvonhackewitz:
> I would use "cannot" if something is logically impossible, and "must not" for something that we chose to not support. See also https://english.stackexchange.com/q/19684.

I followed this when I made the changes.  In some cases, however, I wasn't sure what would be more suitable.

PR #4749, opened 2025-02-07
lib: use `expanding_array.expand` in `expanding_array.concat`

fridis:
This used to cause bad performance in `tests/process` during DFA, but I currently do not see this any longer, so I try to change this.


fridis:
This slows down the JVM backend test runs by a factor 1.75 (1326000ms instead of 756000ms), so this should not be merged yet.

PR #4760, opened 2025-02-10
base: remove `sum`, add `Sequence.sum` and `Sequence.product`

maxteufel:
`sum` was added because it was not possible to further constrain features in `Sequence` to only work with sequences of a specific type.

Also remove `count`. It likewise pollutes the name space.


michaellilltokiwa (approved):

modules/base/src/Sequence.fz
```diff
@@ -1179,7 +1179,7 @@ public Sequence(public T type) ref : property.countable is
     else
       cnt := (map _->T.one).fold T.sum
       sum := fold T.sum
-      sum / cnt
+      Sequence.average.this.sum / cnt
```
fridis, line 1182: Strange that this compiles, the `sum` field is unused.

PR #4761, opened 2025-02-10
enable fuir serialization for jenkins builds

michaellilltokiwa:
[ci skip]


michaellilltokiwa:
@maxteufel I think it has no effect while building Fuzion.

PR #4768, opened 2025-02-11
base/Sequence: simplify `average` and `variance` features

maxteufel:


modules/base/src/Sequence.fz
```diff
@@ -1177,9 +1177,8 @@ public Sequence(public T type) ref : property.countable is
     if is_empty
       nil
     else
-      cnt := (map _->T.one).fold T.sum
-      sum := fold T.sum
-      Sequence.average.this.sum / cnt
+      cnt := (map _->T.one).sum
```
michaellilltokiwa, line 1180: Does `sum / count` work?

michaellilltokiwa (approved):

PR #4793, opened 2025-02-13
lib/Sequence: implement `scan1`, variant of `scan` without initial value

maxteufel:


michaellilltokiwa (changes requested):
Cool, but a test would be nice.

simonvonhackewitz (approved):

michaellilltokiwa (approved):

PR #4802, opened 2025-02-14
build: Enable building `fuzion.ebnf` in a dir different than the source dir

fridis:
It should be possible to build `fuzion` from any dir using `make -f <fuzion-dir>/Makefile`. This did not work for `fuzion.ebnf` and, what is worse, the error was silently ignored with an empty `fuzion.ebnf` file.

Now, `ebnf.fz` expects the source files that contain ebnf data as arguments. Furthermore, it produces errors in case these arguments are missing or in case the files do not exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -479,7 +479,7 @@ $(BUILD_DIR)/%.md: $(FZ_SRC)/%.md
 
 $(FUZION_EBNF): $(FUZION_BASE) $(FZ_SRC)/bin/ebnf.fz
 	mkdir -p $(@D)
-	$(FZ) $(FZ_SRC)/bin/ebnf.fz > $@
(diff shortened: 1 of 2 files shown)


michaellilltokiwa:
I think the order of rules may not be stable with this change. It is either Lexer then Parser or vice versa.
Probably okay but maybe it is better to always have Lexer rules first?

fridis:
> I think the order of rules may not be stable with this change. It is either Lexer then Parser or vice versa. Probably okay but maybe it is better to always have Lexer rules first?

I was aware of this and thought about adding a comment or listing the files explicitly, but came to the conclusion a different order is less tragic than missing rules in case the source files get rearranged or split up.

michaellilltokiwa (approved):

PR #4809, opened 2025-02-17
remove `__marker_for_make__` for fuzion rt

### LINKED ISSUE #4807: Abuse of `__marker_for_make__` for `build/lib/libfuzion.so`
The hack of using `__marker_for_make__` is used for compilation of Java source code to class files since we cannot  easily get a list of the class files created from given Java source files and just using the package directory as a target does not work since its date does not get updated and dirs are handled somewhat differently by make. 

In short, when we know what files a makefile rule produces, the rule should list these files on the left hand side of the rule. In particular, `$(FUZION_RT)` should be a list of the library files for the platform we are building for (.dll, .so, .dylib). 

I ran into this since for some reason, I got `Cannot open library: libfuzion.so`, so I did
```
 > make build/lib/libfuzion.so
```
which should be the standard way to build a file, but this did not work, so I had to search the Makefile for the corresponding rule.

michaellilltokiwa:
fixes #4807

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -300,7 +300,13 @@ MOD_FZ_CMD_DIR = $(BUILD_DIR)/modules/fz_cmd
 MOD_FZ_CMD_FZ_FILES = $(MOD_FZ_CMD_DIR)/__marker_for_make__
 MOD_FZ_CMD = $(MOD_FZ_CMD_DIR).fum
 
-FUZION_RT = $(BUILD_DIR)/lib/__marker_for_make__
+ifeq ($(OS),Windows_NT)
+	FUZION_RT = $(BUILD_DIR)/lib/fuzion.dll
+else ifeq ($(OS),Darwin)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1436,14 +1442,11 @@ ifeq ($(OS),Windows_NT)
 	-fno-trigraphs -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -std=c11 \
 	$(BUILD_DIR)/include/win.c $(BUILD_DIR)/include/shared.c -o $(BUILD_DIR)/lib/fuzion.dll \
```
fridis, line 1443:
```suggestion
	$(BUILD_DIR)/include/win.c $(BUILD_DIR)/include/shared.c -o $@ \
```

fridis (approved):
just a small suggestion for consistency.

PR #4818, opened 2025-02-19
time/date_time: remove `before`, `after`, `equal`, and related features

maxteufel:
This is fully covered by the fact that `date_time` is orderable. The canonical way for comparison thus are the operators defined for orderables. In particular, the diamond operator `<>` can be used as a replacement for the removed `before_after_equal` feature.


michaellilltokiwa (approved):
Much cleaner!

PR #4834, opened 2025-02-20
os/process: remove explicit inheritance from `property.equatable`

maxteufel:
`property.orderable` implies `property.equatable`.


michaellilltokiwa (approved):

michaellilltokiwa:
We can also remove: `public fixed redef type.equality(a, b process.this) => a.id = b.id`

PR #4873, opened 2025-03-04
lib: add `mutable_hash_map`

### LINKED ISSUE #4642: Add `mutable_hash_map` inheriting from `Mutable_Map`
Add `mutable_hash_map` as an implementation of `Mutable_Map`.

Also implement `container.hash_map.as_mutable_map`

simonvonhackewitz:
fix #4642

### DIFF
--- a/modules/base/src/container/Mutable_Hash_Map.fz
+++ b/modules/base/src/container/Mutable_Hash_Map.fz
@@ -25,6 +25,11 @@
 
 # Mutable_Hash_Map -- a mutable hash map from keys HK to values V
 #
+# Uses a mutable array internally, when upper_load_factor is reached size is multiplied
+# by resize_factor (doubled, tripled, etc.) and elements are copied to new array of this size.
+# If lower_load_factor is reached size is divided by resize_factor.
+# Insert and retrieve operations on this map are therefore in amortized O(1).
+#
 module:public Mutable_Hash_Map(
         # the mutate effect
         #
@@ -67,11 +72,6 @@ is
   #
   load := mut 0
 
-  # number of entries in this map
-  #
-  redef size =>
-    load.get
-
   # maximum number of pairs before size is increased
   #
   max_elem => (allocated_size.get.as_f32 * upper_load_factor).ceil.as_i32
@@ -84,6 +84,10 @@ is
   #
   allocated_size := mut (min_size.as_f32 / upper_load_factor).ceil.as_i32
 
+  # the contents
+  #
+  contents := mut (LM.env.new_array (option (tuple HK V)) allocated_size.get.as_i64 nil)
+
 
   # calculate the index of k within contents array in case of no conflict
   #
@@ -95,11 +99,6 @@ is
   #
   collision (at i64) =>
     (at + 1) % allocated_size.get.as_i64  # NYI: dumb collision function, check literature and improve!
-
-
-  # the contents
-  #
-  contents := mut (LM.env.new_array (option (tuple HK V)) allocated_size.get.as_i64 nil)
 
 
   # increase internal allocation if upper_load_factor is reached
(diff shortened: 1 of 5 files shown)


modules/base/src/container/Mutable_Map.fz
```diff
@@ -85,14 +85,14 @@ public Mutable_Map(public K type : property.equatable, public V type) ref is
...
-  # if key already exists, NYI what should happen?
+  # if key already exists, existing value gets updated
   #
   public put(k K, v V) unit => abstract
 
 
   # add all key-value pairs to this map
   #
-  # if key already exists, NYI what should happen?
+  # if key already exists, existing value gets updated
```
michaellilltokiwa, line 95: existing values

modules/base/src/container/mutable_hash_map.fz
```diff
@@ -0,0 +1,262 @@
...
+       t tuple =>
+        (ek, v) := t
+        if ek = k
+          contents.get[del_idx] := nil
+          load <- load.get - 1
+          fill_hole del_idx (collision del_idx)
+          v
+        else
+          del (collision del_idx)
+
```
michaellilltokiwa, line 193: insert newline?

modules/base/src/container/mutable_hash_map.fz
```diff
@@ -0,0 +1,262 @@
...
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature mutable_hash_map
+#
+# -----------------------------------------------------------------------
+
+
+
+# mutable_hash_map -- a mutable hash map from keys HK to values V
+#
```
michaellilltokiwa, line 27: Needs a few sentences explaining the underlying data structure

modules/base/src/container/mutable_hash_map.fz
```diff
@@ -0,0 +1,262 @@
...
+  # in case of a collision at given position,
+  # return the next alternative position to check
+  #
+  collision (at i64) =>
+    (at + 1) % allocated_size.get.as_i64  # NYI: dumb collision function, check literature and improve!
+
+
+  # the contents
+  #
+  contents := mut (LM.env.new_array (option (tuple HK V)) allocated_size.get.as_i64 nil)
```
michaellilltokiwa, line 102: move all the fields to the beginning of the constructor

michaellilltokiwa (commented):
Nice.

You use mut inside of a value feature. I think it should probably be a Mutable_Hash_Map?

modules/base/src/container/Mutable_Hash_Map.fz
```diff
@@ -0,0 +1,272 @@
...
+    container.hash_map HK V (pairs.map (.values.0) ).as_array (pairs.map (.values.1)).as_array
+
+
+  # empty -- convenience routine to create an empty instance of hash_map
+  #
+  public fixed redef type.empty => container.Mutable_Hash_Map LM HK V 11 0.25 0.5 2
+  public fixed type.empty(
+    # initial size, also the size of the hash map will not decrease below this
+    #
+    private initial_and_min_size i32,
```
michaellilltokiwa, line 259: private is default visibility, can be removed

michaellilltokiwa:
> Should the possibility to set a initial/minimum size and the increase factor be removed?

I don't see much value in having it exposed.

michaellilltokiwa (approved):

simonvonhackewitz:
> I don't see much value in having it exposed.

The idea was to avoid the resize operations if you know the maximum number of elements and don't care about the extra memory when it is not full.
Or if you create an empty map knowing you are going to add x elements.

But I'm not sure if it was a good idea to combine initial size and minimum size into one attribute.

modules/base/src/container/Mutable_Hash_Map.fz
simonvonhackewitz: you're right, I think `public` would not even make sense for function features...

PR #4879, opened 2025-03-05
time/date_time: move features used only by contract inside contract

maxteufel:


michaellilltokiwa (approved):
Maybe we want thinks like days_in_month, is_leap_year as a type feature in the future.

PR #4894, opened 2025-03-06
fix #4531

### LINKED ISSUE #4531: we must provide a way to read `io.stdin` and react to `return`
Please create an issue that we must provide a way to read `io.stdin` and react to `return` and reference this issue here.

_Originally posted by @fridis in https://github.com/tokiwa-software/fuzion/pull/4526#pullrequestreview-2511720032_

michaellilltokiwa:
This gist of this fix is that we set `stdin` to none blocking then use `poll` when reading to wait until input is available.
This way we can read line by line from stdin since stdin is usually line buffered. No Ctrl-D necessary anymore.
This will probably not work on Windows.

fixes #4531

### DIFF
--- a/bin/add_simple_test.fz
+++ b/bin/add_simple_test.fz
@@ -25,8 +25,7 @@
 #
 add_simple_test =>
 
-  # NYI: BUG: #4531, can not end input via return currently
-  say "Name of the new test: (press Ctrl+D to end input)"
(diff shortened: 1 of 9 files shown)


michaellilltokiwa:
@maxteufel Exactly that actually but only in the c backend.

PR #4898, opened 2025-03-07
ast: forbid setting `private` visibility on arguments of function features

### LINKED ISSUE #4886: raise error when using visibility on function args
you're right, I think `public` would not even make sense for function features...

_Originally posted by @simonvonhackewitz in https://github.com/tokiwa-software/fuzion/pull/4873#discussion_r1983022513_

simonvonhackewitz:
`public` is already forbidden, arguments are private by default

fix #4886

### DIFF
--- a/src/dev/flang/ast/Function.java
+++ b/src/dev/flang/ast/Function.java
@@ -282,7 +282,7 @@ public class Function extends AbstractLambda
         for (var n : _names)
           {
             var arg = new Feature(n._pos,
-                                  Visi.PRIV,
+                                  Visi.UNSPECIFIED,
                                   0,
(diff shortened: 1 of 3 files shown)


fridis:
I do not understand the motivation for this, I can see reasons for `public` as well as for `private` argument fields: 
For `public` argument fields
```
  public Abstract_Point is
    public x, y i32 => abstract
  private my_point (public x, y i32) : Abstract_Point is 
  get_origin Abstract_Point => my_point 0 0 
```
For `private` argument fields
```
  public LoginDetails(public name String, 
                      private password String) is
    public check_password(entered String) =>
      password = entered
```

simonvonhackewitz:
superseded by #4904

PR #4918, opened 2025-03-12
ast: `If`, change `typeForInferencing` to return `null` on `t_ERROR`

michaellilltokiwa:


michaellilltokiwa:
@maxteufel
Two reasons:
- One is consistency of `typeForInferencing`, which should return null if type is not yet known.
- And #4916 where I work on improving type inference of lambas but this current behavior led to problems.

PR #4930, opened 2025-03-13
docs: make api sources link configurable

### LINKED ISSUE #4705: API documentation tools: sources link should be configurable
I had to add a patch for publication of the API docs on [fuzion-lang.dev](https://fuzion-lang.dev) that replaces `[src]` links to `/api/modules/`  to instead go to `/api`. 
The location of the sources produced by the API documentation tool should be configurable by a command line option such that this post-processing is not needed and that we can change the location of the sources anytime we want. 
The default location, however, should be pointing to the correct location in the `build` directory such that one can browse the docs in`build/apidocs` directly.

simonvonhackewitz:
fix #4705

I'm not sure how helpful this is other than for the web server, because for `-api-src=custDir` this would generate links like the following
`custDir/base/abort.fz`
but in the fuzion repo this file is in 
`modules/base/src/abort.fz`

So should the part `base/abort.fz` be also configurable so that `src` could be added?

### DIFF
--- a/src/dev/flang/tools/docs/Html.java
+++ b/src/dev/flang/tools/docs/Html.java
@@ -558,7 +558,7 @@ public class Html extends ANY
   /**
    * the link [src] to the source file
    */
-  private static String source(AbstractFeature feature)
+  private String source(AbstractFeature feature)
   {
     return "<div class='pl-5'><a href='$1'>[src]</a></div>"
       .replace("$1", featureURL(feature));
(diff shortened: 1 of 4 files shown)


src/dev/flang/tools/docs/Html.java
```diff
@@ -558,7 +558,7 @@ private String htmlEncodedQualifiedName(AbstractFeature af)
   /**
    * the link [src] to the source file
    */
-  private static String source(AbstractFeature feature)
+  private String source(AbstractFeature feature)
```
michaellilltokiwa, line 561: What is the reason for removing static?

simonvonhackewitz: Because it calls`featureURL()`, which is no longer static, because it now calls `apiScrDir()` from DocsOption instead of the static field baseApiDir.

michaellilltokiwa (approved):

PR #4936, opened 2025-03-14
parser: fix partial application of automatic select

### LINKED ISSUE #4750: Partial application of automatic select as in `m.map (.1)` does not work yet
We can use a lambda with an automatic select on a tuple:
```
 > ./build/bin/fz -e 'say ([(3,4),(5,6)].map x->x.1 |> sum)'
10
```
Also, partial application of a dot-call works with an explicit select:
```
 > ./build/bin/fz -e 'say ([(3,4),(5,6)].map (.values.1) |> sum)'
10
```
But, partial application of an automatic select currently fails:
```
 > ./build/bin/fz -e 'say ([(3,4),(5,6)].map (.1) |> sum)'

command line:1:26: error 1: Syntax error: expected identifier name, infix/prefix/postfix operator, 'ternary ? :', 'index' or 'set' name, found numeric literal '1'
say ([(3,4),(5,6)].map (.1) |> sum)
[...]
```

simonvonhackewitz:
fix #4750


src/dev/flang/parser/Parser.java
```diff
@@ -1945,25 +1945,26 @@ Expr operatorExpr(boolean mayUseCommas)
...
+          }
+        else
+          {
+            return new Partial(oe.op(0)._pos,
+                               oe.op(0)._text);
+          }
+      }
+    else
+      {
+        return Partial.dotCall(tokenSourcePos(), a -> current() == Token.t_numliteral ? select(a, null)
```
michaellilltokiwa, line 1965: @simonvonhackewitz I think we need/want a grammar rule here: 
Sth. like:
```
partialCall :  select
            | pureCall
            ;
```

michaellilltokiwa (approved):
Great patch! I have one grammar suggestion.

tests/reg_issue4750/reg_issue4750.fz
```diff
@@ -0,0 +1,28 @@
...
+#
+#  Source code of Fuzion test reg_issue4750
+#
+# -----------------------------------------------------------------------
+
+reg_issue4750 =>
+
+  say ([(3,4),(5,6)].map (.1) |> .sum)
+
+  say <| [("Error:","This does"),("not","work.")].map (.1)
```
fridis, line 28: can you add test cases that go further and to things like `.map (.1.0)`, `.map (.1[i])`, `.map (.1.as_string)`, etc.?

fridis (commented):
Very nice and small patch! 

I agree with @michaellilltokiwa request for an EBNF rule for this. 

I added a request for some more test cases, in particular with call tails following the partial select.

michaellilltokiwa (approved):

PR #4939, opened 2025-03-17
time/calendar_duration: implement `as_string`

maxteufel:


modules/base/src/time/calendar_duration.fz
```diff
@@ -90,7 +90,48 @@ is
   # create a string representation of this duration.
   #
   public redef as_string String =>
-    "NYI"
+    lm : mutate.
+    lm ! ()->
+      is_positive := lm.env.new bool false
```
michaellilltokiwa, line 95: I would prefer not using a mut here but `is_positive := years > 0 || months >0 || ...`

modules/base/src/time/calendar_duration.fz
```diff
@@ -90,7 +90,48 @@ is
...
+        is_positive <- true
+
+      if months > 0
+        s <- "{s.get}{months}M"
+        is_positive <- true
+
+      if days > 0
+        s <- "{s.get}{days}D"
+        is_positive <- true
+
```
michaellilltokiwa, line 117: Instead of `with_time_prefix` I would prefer sth. like `needs_time_prefix` then add the prefix here.

michaellilltokiwa (approved):

PR #4948, opened 2025-03-18
Restrict possibilities for indentation of `else`

### LINKED ISSUE #4570: Indentation of `else` in `if` statement should be more restrictive
This code 
```
x =>
  if 3 = 4 then
    say "1"
    say "2"
             else if 4 =5 then
    say "3"
    say "4"
                            else
    say "5"
    say "6"
```
is accepted
```
 > ./build/bin/fz tst_indent.fz 
5
6
```
while I think it should fail and require `else` to have the same indentation level as `if`. 

Here is another example: 
```
y0 => if 3=4 then say "hi"; if 4=5 then say "ho"
      else say "he"
y1 => if 3=4 then say "hi"; if 4=5 then say "ho"
             else say "he"
y2 => if 3=4 then say "hi"; if 4=5 then say "ho"
                           else say "he"
y3 => if 3=4 then say "hi"; if 4=5 then say "ho"
                            else say "he"
y4 => if 3=4 then say "hi"; if 4=5 then say "ho"
                             else say "he"
y5 => if 3=4 then say "hi"; if 4=5 then say "ho"
                                   else say "he"
say "y0"; y1
say "y1"; y1
say "y2"; y2
say "y3"; y3
say "y4"; y4
say "y5"; y5
```
which results in
```
 > ./build/bin/fz tst_indent2.fz 

y0
he
y1
he
y2
he
y3
y4
y5
```
I think the behavior is ok, but cases `y2` and `y4` should better be forbidden and flag an error.  Cases `y1` and `y5` , i.e., alignment with `then` should maybe be allowed, but I am not sure here.
We also need more positive and negative test cases using `if .. then .. else if ..`, loops `for .. while .. do .. else if .. else ..`, etc. 
Should the following cases
```
  if cc1 then
  else if cc2 then
  else
```
and
```
  while cc1 do
  else if cc2 then 
  else
```
be handled equally? Probably yes, but I am not so sure here. 

[Here are the indentation rules for Haskell](https://en.m.wikibooks.org/wiki/Haskell/Indentation) that might be useful for us.

simonvonhackewitz:
fix #4570

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -2977,8 +2977,8 @@ ifexpr      : "if" exprInLine thenPart elseBlock
             var errPos = (ifPos != null) ? ifPos : (_outerElse != null) ? _outerElse : _then;
 
             Errors.indentationProblemEncountered(tokenSourcePos(), errPos,
-              "When it is not in the same line as " + Errors.skw("if") + ", " + Errors.skw("else")
-              + " must be aligned with " + Errors.skw("if") + ", " + Errors.skw("then")
+              "When " + Errors.skw("else") + " is not in the same line as " + Errors.skw("if")
+              + ", it must be aligned with " + Errors.skw("if") + ", " + Errors.skw("then")
               + " or " + Errors.skw("else if") + ".") ;
           }
 
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -86,6 +86,8 @@ public class Parser extends Lexer
   private boolean _nestedIf = false;
   private int _lastIfLine = -1;
   private boolean _elif = false;
+  private SourcePosition _outerElse = null;
+  private SourcePosition _then = null;
 
   /*--------------------------  constructors  ---------------------------*/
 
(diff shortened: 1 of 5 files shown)


src/dev/flang/parser/Parser.java
```diff
@@ -2958,6 +2961,27 @@ If ifexpr()
...
+        if (currentAtMinIndent() == Token.t_else && !(
+            (_then != null      && (_then.line()      == elPos.line() || _then.column()      == elPos.column())) ||
+            (_outerElse != null && (_outerElse.line() == elPos.line() || _outerElse.column() == elPos.column())) ||
+            (ifPos != null      && (ifPos.line()      == elPos.line() || ifPos.column()      == elPos.column())))
+           )
+          {
+            var errPos = (ifPos != null) ? ifPos : (_outerElse != null) ? _outerElse : _then;
+
+            Errors.indentationProblemEncountered(tokenSourcePos(), errPos,
+              "When it is not in the same line as " + Errors.skw("if") + ", " + Errors.skw("else")
```
michaellilltokiwa, line 2980: "When it..." is a little strange. We should be more specific, what is _it_?

michaellilltokiwa (changes requested):

michaellilltokiwa (approved):

PR #4958, opened 2025-03-20
ast: fix NPE in Select.toString

michaellilltokiwa:


simonvonhackewitz:
Do we need a regression test for this?

michaellilltokiwa:
@simonvonhackewitz No, this code is only used when debugging.

simonvonhackewitz:
I just created #4959 because that's probably a different error then.

michaellilltokiwa:
@simonvonhackewitz Yes it is different, that is in `resolveTypes` not `toString`.

PR #4981, opened 2025-03-25
ast: extend syntax sugar to convert numliterals to required type

### LINKED ISSUE #4754: Syntax sugar for `T.zero`?
The type of numeric literatals like `0` or `1` is already adapted automatically to match the target type as in 
```
  f(x i32, y u64) => say "$x $y"
  f 0 1
```
This could be extended to automatically convert `0` to `T.zero`, `1` to `T.one` and `2`..`4294967295` to `T.from_u32 2`..`T.from_u32 4294967295`. Then, we could write in `Sequence` (see #4753) 
```
  public sum 
    pre
      T : numeric
  => 
    reduce 0 (+)
```

simonvonhackewitz:
fix #4754


fridis (changes requested):
Very nice and clean patch!

Since `numeric.type.zero` and `numeric.type.one`  must be implemented anyway and should have better performance, it would IMHO be good to specialize the code for `0` and `1` to call `numeric.type.zero` and `numeric.type.one`, respectively. 

What error do you get if you use illegal values such as `3.14` or `1E10`? 

Would be good to add test cases for `0` and `1` and maybe some more values to the test, but avoid large values since their performance is awful. 

Eventually, the middle end or back ends should optimize the code and replace calls to `from_u32` by the corresponding constant to avoid poor performance. 

It might, however, make sense to add an issue (and later fix it) to optimize the generic implementation of `from_u32` to be in `O(log n)` by checking the argument bit-wise and using shifts to create the result.

src/dev/flang/ast/NumLiteral.java
```diff
@@ -440,7 +441,7 @@ AbstractType typeForUnion()
   {
     // NYI: UNDER DEVELOPMENT: This seems to work even if we always return null
```
michaellilltokiwa, line 442: This comment can be removed

simonvonhackewitz:
> What error do you get if you use illegal values such as `3.14` or `1E10`?

`1E10` works for `u64`


3.14:
```
❯ fz /tmp/test.fz

/tmp/test.fz:17:14: error 1: Numeric literal used for integer type is not a whole number
  say (d.foo 3.14)

Type propagation results in an integer type that cannot hold a value that is not integer.
Numeric literal: '3.14'
Assigned to type: 'u64'

one error.
```


1E100:
```
❯ fz /tmp/test.fz

/tmp/test.fz:17:14: error 1: Integer constant value outside of allowed range for target type
  say (d.foo 1E100)

Type propagation results in a type that is too small for the value represented by the given constant.
Numeric literal: '1E100'
Assigned to type: 'u64'
Acceptable range of values: '0' .. '18446744073709551615'

one error.
```

src/dev/flang/ast/NumLiteral.java
```diff
@@ -915,6 +914,46 @@ public byte[] data()
...
 
 
+  @Override
+  protected Expr resolveSyntacticSugar2(Resolution res, Context _context)
+  {
+    if (_propagatedType != null && _propagatedType.isGenericArgument())
+      {
+        Call result;
+
+        if (_originalString.equals("0") || _originalString.equals("0E0"))
```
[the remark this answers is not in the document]
simonvonhackewitz: > how will this handle "+0" and "-0"?
it will not do the optimization and instead just create a call to `.from_u32`. Do you think these cases should be added?

> since technically 0^0 = 0 is not really a natural definition
So just leave that out from the optimization?

src/dev/flang/ast/Constant.java
```diff
@@ -174,6 +173,12 @@ Expr propagateExpectedTypeForPartial(Resolution res, Context context, AbstractTy
       : this;
   }
 
+
+  protected Expr resolveSyntacticSugar2(Resolution res, Context _context)
```
michaellilltokiwa, line 177: please copy the comment from the other `resolveSyntacticSugar2` implementations

michaellilltokiwa (approved):

michaellilltokiwa (approved):

PR #4986, opened 2025-03-27
ast: avoid errors caused by additional lambda argument

### LINKED ISSUE #4457: AoC: Lambda passed as an extra argument to a call should not result in an error
Try this:
```
 > ~/fuzion/work/build/bin/fz -e 'say "hi" x->x+1' 

command line:1:10: error 1: No type information can be inferred from a lambda expression
say "hi" x->x+1

A lambda expression can only be used if assigned to a field or argument of type 'Function'
with argument count of the lambda expression equal to the number of type parameters of the type.  The type of the
assigned field must be given explicitly.
To solve this, declare an explicit type for the target field, e.g., 'f (i32, i32) -> bool := x, y -> x > y'.


command line:1:1: error 2: Different count of arguments needed when calling feature
say "hi" x->x+1

Feature not found: 'say' (2 arguments)
Target feature: 'universe'
In call: 'say "hi" x->x+1'
To solve this, you might change the actual number of arguments to match the feature 'say' (one argument) at $MODULE/say.fz:31:8:
public say(s Any) => io.out.println s


2 errors.
```
The second error that is produced is fine, the argument count for the call to `say` is wrong. However, in the context of a wrong argument count, the first error should not appear at all. 
My guess is that the problem is that the error for the call is stored in `Call._pendingError`, hoping we might still solve this through partial application, but then the argument types are requested even though there is a pending error.

simonvonhackewitz:
fix #4457

### DIFF
--- a/src/dev/flang/ast/Feature.java
+++ b/src/dev/flang/ast/Feature.java
@@ -2023,7 +2023,7 @@ A ((Choice)) declaration must not contain a result type.
         @Override public AbstractType action(AbstractType   t) { return t.checkConstraints(_context);           }
         @Override public void         action(Cond           c) {        c.checkTypes();                         }
         @Override public void         actionBefore(Block    b) {        b.checkTypes();                         }
-        @Override public Expr action(Function f) { return f.checkTypes(); }
+        @Override public Expr         action(Function       f) { return f.checkTypes();                         }
       });
 
     res._module.checkTypes(this);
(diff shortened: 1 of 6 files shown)


src/dev/flang/ast/Feature.java
```diff
@@ -2023,6 +2023,7 @@ void checkTypes(Resolution res)
         @Override public AbstractType action(AbstractType   t) { return t.checkConstraints(_context);           }
         @Override public void         action(Cond           c) {        c.checkTypes();                         }
         @Override public void         actionBefore(Block    b) {        b.checkTypes();                         }
+        @Override public Expr action(Function f) { return f.checkTypes(); }
```
michaellilltokiwa, line 2026: please fix code alignment

src/dev/flang/ast/Function.java
```diff
@@ -540,6 +536,13 @@ public String toString()
     return _names + " -> " + _expr;
   }
 
+
+  public Expr checkTypes()
```
michaellilltokiwa, line 540: needs a comment

michaellilltokiwa (approved):

PR #4989, opened 2025-03-31
Dangling else error issue3508

### LINKED ISSUE #3508: dangling else should cause error if everything is in one line
> This is a variant of the [dangling else](https://en.wikipedia.org/wiki/Dangling_else) problem, which we also have and should solve similarly by producing an error.

_Originally posted by @fridis in https://github.com/tokiwa-software/fuzion/issues/2702#issuecomment-2270891573_

Currently the dangling else belongs to the inner if.

The error could look something like this
```
> fz -e 'if true if false then say "inner true" else say "dangling else"'

command line:1:47: error 1: Ambiguous else block.
if true if false then say "inner true" else say "dangling else"
---------------------------------------^^^^^^^^^^^^^^^^^^^^^^^^
It is unclear whether the else block belongs to the outer or inner if.
To solve this, add braces { } or use line breaks and indentation.

```

simonvonhackewitz:
fix #3508


src/dev/flang/parser/Parser.java
```diff
@@ -2971,15 +2983,25 @@ Block elseBlock()
...
+            // reached unambiguous `else`, therefore a new if-else-block would be unambiguous
+            _danglingElseState--;
             result = new Block(false, new List<Expr>(ifexpr(true)));
           }
         else
           {
             _outerElse = null;
+
+        _danglingElseState--;
+
```
fridis, line 3004: looks strange to decrement this both in the `if` and in the `else ` part.

src/dev/flang/parser/Parser.java
```diff
@@ -2994,6 +3016,9 @@ Block elseBlock()
       (result == null          ||
        result instanceof Block    );
 
+    // reached unambiguous `else`, therefore a new if-else-block would be unambiguous
+    _danglingElseState--;
+
```
fridis, line 3021: Is this correct to decrement this again after it was alread decremented after `if (skip(true, Tokwn.t_else))`?

src/dev/flang/ast/AstErrors.java
```diff
@@ -2428,6 +2428,14 @@ public static void mustNotDeclareFieldInModulesUniverse(AbstractFeature f)
           "To solve this, remove the field or move its declaration into a parent feature.");
   }
 
+  public static void danglingElse(SourcePosition pos)
+  {
+    error(pos,
+          "Ambiguous dangling else",
+          "It is unclear to which " + skw("if") + " the " + skw("else") + " block belongs to."
```
fridis, line 2435:
```suggestion
          "It is unclear to which " + skw("if") + " the " + skw("else") + " block belongs."
```

src/dev/flang/parser/Parser.java
```diff
@@ -89,6 +89,14 @@ static enum FormalOrActual
    */
   private int _lastIfLine = -1;             // NYI: CLEANUP: state in parser should be avoided see #4991
 
+  /**
+   * To detect and forbid dangling else when everything is in one line.
+   *
+   * 1 for first `if` in line, increased with every `if` in same line
+   * reset by newline or a block with braces
+   */
+  private int _danglingElseState = 0;       // NYI: CLEANUP: state in parser should be avoided see #4991
```
fridis, line 98: I would very much prefer not to add any more state to the parser, but instead to remove the existing state!

Please do not include an initial value since then you forget to copy this in `Parser.clone()`/`Parser(Parser original)`.

fridis (changes requested):
Would be good to avoid the extra state in the parser!

I expect similar problems with loops as in 
```
  while true do while false do else say "dangling"
```
or also
```
  while true do while false do until (true)  say "dangling"
```
or 
```
  while true while false until (true)  say "dangling"
```
Can you check these?

src/dev/flang/parser/Parser.java
simonvonhackewitz: I don't think so. I forgot this PR in draft state a while ago and don't remember why I did it. But, because of how it is incremented, I think it did not cause any errors.

src/dev/flang/parser/Parser.java
simonvonhackewitz: Do you have an idea on how to avoid the state in the parser?

PR #4990, opened 2025-03-31
parser: remove states of parser

simonvonhackewitz:


src/dev/flang/parser/Parser.java
```diff
@@ -83,11 +83,23 @@ static enum FormalOrActual
...
-  private boolean _nestedIf = false;
-  private int _lastIfLine = -1;
-  private boolean _elif = false;
-  private SourcePosition _outerElse = null;
-  private SourcePosition _then = null;
+  /**
+   * contains the last line number in which an `if` was found,
+   * required to set the semicolon state for the ambiguous semicolon error
+   */
+  private int _lastIfLine = -1;             // NYI: state in parser should be avoided see #4991
```
michaellilltokiwa, line 90: NYI should always have a label. In this case NYI: CLEANUP: seems appropriate.

src/dev/flang/parser/Parser.java
```diff
@@ -2913,30 +2925,35 @@ boolean isIndexVarPrefix()
   /**
    * Parse ifexpr
    *
+   * @param elif is this part of an `else if`
```
michaellilltokiwa, line 2928: This comment seems to belong to the other if method?
I wonder why javadoc does not complain?

michaellilltokiwa (commented):
Definitely nicer than before. :-)

src/dev/flang/parser/Parser.java
simonvonhackewitz: it probably does, but they are only warnings and don't cause make to fail

PR #5027, opened 2025-04-08
tests: file/dir, use unicode symbols

michaellilltokiwa:


simonvonhackewitz:
I think this is already tested in `lib_io_utf8.fz`

PR #5033, opened 2025-04-09
create link libclang.dll, clang.dll

michaellilltokiwa:
lib prefix is unusual on windows, so creating a softlink, such that System.mapLibraryName("clang") works as expected.


michaellilltokiwa:
I cancelled the tests since window.yml is no relevant change.

PR #5041, opened 2025-04-10
rename libfuzion -> libfuzion_rt

michaellilltokiwa:
I think this might be the better name in the long run...


fridis (commented):
Why do you think this is better. I prefer a less cryptic name as long as this does not cause a conflict. So I would stay with `libfuzion` as long as there is no conflict and no plan to use it for something else.

michaellilltokiwa:
@fridis If I would read libfuzion I'd think this is a library for the compiler or for calling/interacting with fuzion code.

PR #5049, opened 2025-04-11
fe: remove `If`, create `Match` early

michaellilltokiwa:
This also changes the type inference for matches to consider _nested_ cases as well.
Also this has the side effect that if and match are resolved in the exact same way except for some error messages.


src/dev/flang/ast/AbstractMatch.java
```diff
@@ -168,7 +168,9 @@ public AbstractType type()
           {
             new IncompatibleResultsOnBranches(
               pos(),
-              "Incompatible types in cases of match expression",
+              "Incompatible types in cases of " +
+                (kind() == Kind.Plain ? "match" : "if") +
```
fridis, line 172:
```suggestion
              "Incompatible types in " +
                (kind() == Kind.Plain ? "cases of match" : "branches of if") +
```

src/dev/flang/ast/Match.java
```diff
@@ -258,6 +259,65 @@ public boolean producesResult()
...
+   *
+   * @param c then condition
+   *
+   * @param b the "true"-block, must not be null
+   *
+   * @param elseB else block, may be null
+   *
+   * @param fromContract is this an if generated from a contract? (for error messages)
+   */
+  public static Match If(SourcePosition pos, Expr c, Expr b, Expr elseB, boolean fromContract)
```
fridis, line 275: Should start with lower case, maybe call this `createIf`.

fridis (changes requested):
Very nice code size reduction!

Just minor comments.

modules/base/src/net/server.fz
```diff
@@ -107,7 +107,7 @@ is
 
     # bind to port
     (fuzion.sys.net.bind f.as_num s.as_num p.as_num addr port)
-      .bind sd->
+      .bind i32 sd->
```
fridis, line 110: Is this an accidental commit or does your PR worsen the type inference?

fridis (commented):
Looks good, just wondering if the type inference somehow got worse due to the change in `server.fz`.

modules/base/src/net/server.fz
michaellilltokiwa: I created an issue: #5054

PR #5065, opened 2025-04-15
ast: universe, return type use NoType

michaellilltokiwa:
Was surprised that universe.isContructor returned true.


michaellilltokiwa:
hmm, maybe it is fine that universe is considered a constructor...

Issue #5085, opened 2025-04-16
Reading unitialized field from lambda not detected


fridis:
Take this code
```
x i32 := [1,2].fold1 a,b->x
say x
```
which runs without causing an error, showing `-555555555` when using `-interpreter`, which is the interpreters dummy value for uninitialized `i32` values:
```
 > ../../bin/fz -interpreter test7.fz 
-555555555
```

PR #5086, opened 2025-04-16
tests: cleanup in visibility scoping

fridis:
`test2` actually works, correctly complains about ambiguity of `b` and if `A.this.` or `C.this.` is added to solve this, correctly complains about uninitialized field in the first case.

`test7` is now issue #5085.

For `test8`, `fz` correctly complains about reading an uninitialized field.


michaellilltokiwa (changes requested):
I think we should add test8 as a regression test.

michaellilltokiwa (approved):

PR #5118, opened 2025-04-23
lib: add `order`, choice of less/greater/equal

michaellilltokiwa:
I think this is more readable than returning -1/0/1. Also it complicates misuse of this comparison operation.

### DIFF
--- a/modules/base/src/equals.fz
+++ b/modules/base/src/equals.fz
@@ -109,7 +109,7 @@ public infix ⩻(T type : property.orderable, a, b T) bool =>
 # result is > 0 if this > other
 # result is = 0 if this = other
 #
-public infix ⋄(T type : property.orderable, a, b T) lgoe =>
+public infix ⋄(T type : property.orderable, a, b T) order =>
   a <> b
 
 
@@ -142,8 +142,8 @@ public infix <(T type : property.orderable, a, b T) bool =>
 # result is > 0 if this > other
 # result is = 0 if this = other
 #
-public infix <>(T type : property.orderable, a, b T) lgoe =>
-  if      a < b then lower
+public infix <>(T type : property.orderable, a, b T) order =>
+  if      a < b then less
   else if a > b then greater
   else               equal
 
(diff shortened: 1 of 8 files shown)


modules/base/src/lgoe.fz
```diff
@@ -0,0 +1,79 @@
...
+
+
+  # human readable string of this choice
+  #
+  public redef as_string =>
+    match lgoe.this
+      lower => "lower"
+      greater => "greater"
+      equal => "equal"
+
```
fridis, line 78: Maybe a set of predicates `is_less`, `is_less_or_equal`, etc. for all 6 non-trivial combinations would be nice here.

modules/base/src/equals.fz
```diff
@@ -144,10 +142,10 @@ public infix <(T type : property.orderable, a, b T) bool =>
 # result is > 0 if this > other
 # result is = 0 if this = other
```
simonvonhackewitz, line 143: this comment is now outdated

fridis (commented):
Nice idea. A few suggestions.

modules/base/src/equals.fz
```diff
@@ -109,10 +109,8 @@ public infix ⩻(T type : property.orderable, a, b T) bool =>
...
```
simonvonhackewitz, line 110: this comment is now outdated

modules/base/src/lgoe.fz
```diff
@@ -0,0 +1,79 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature lgoe
+#
+# -----------------------------------------------------------------------
+
+# lower -- unit feature to denote the result of a comparion
+#
+public lower is
```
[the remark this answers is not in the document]
simonvonhackewitz: I thought the same. AI says that lower greater and equal are commonly used together, but it also mentioned less as an alternative option.

fridis (approved):
looks nice, would like to see #5122 added, though.

PR #5138, opened 2025-04-24
remove (un)setenv

michaellilltokiwa:
I don't see the point for this anymore, java does not have this either.

When starting a process we can give it a custom map of environment variables.


fridis (approved):
agreed.

PR #5154, opened 2025-04-28
lib: add String.lines, splitting the String at \n

michaellilltokiwa:


michaellilltokiwa:
@maxteufel Hmm, yes but I think 
```
   "some string".lines
```
is more readable than
```
   "some string".split encodings.ascii.lf_str
```
or
```
    "some string".split "\n"
```

PR #5177, opened 2025-05-02
ast: add error for adding precondition when implicit `pre true` is inherited

### LINKED ISSUE #5095: pre else not working as I would expect
the example:
```
ex =>
  A ref is
    b unit => abstract

  B ref : A is
    redef b
    pre else false
    => unit

  say B.b
```
output:
```
unit
```

### LINKED ISSUE #5149: What should happen if a feature redefines multiple features of which only some have an explicit precondition?
A feature redefines two features, because it inherits that feature from two parent features.
What should happen when only one has an explicit precondition?

Forbid to add a precondition because it is implicitly `pre true`?
Allow adding a precondition with `pre else`?
Enforce explicit `pre else true` to clarify what actually happens?

```
parent_1 is
  foo(x i32) unit
    pre x > 0
  => abstract

parent_2 is
  foo(x i32) unit => abstract

child : parent_1, parent_2 is
  redef foo(x i32)
    pre else true  # what should happen?
  => unit
```

@fridis what do you think?

simonvonhackewitz:
fix #5095 
close #5149

### DIFF
--- a/doc/ref_manual/fuzion_reference_manual.adoc
+++ b/doc/ref_manual/fuzion_reference_manual.adoc
@@ -391,7 +391,7 @@ include::rule.adoc[]
 Pre- and postconditions of features are introduced using the `pre` or `post`
 keyword if those features do not redefine an inherited feature:
 
-:RULE_SRC: src/dev/flang/fe/SourceModule.java
+:RULE_SRC: src/dev/flang/ast/Feature.java
 :RULE_ID: PARS_CONTR_PRE_NO_ELSE
 include::rule.adoc[]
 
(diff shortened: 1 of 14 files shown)


simonvonhackewitz:
Why is it allowed to use the keyword `pre` without defining a precondition

e.g.
```
foo
pre
=> unit
```

fridis (changes requested):
Sorry for the very late review of this!

Yes, if any inherited precondition is an implicit `pre true`, flagging an error is a good idea. 

I only have some minor suggestions to improve the test cases.

fridis (approved):
sorry again for the slow review...

fridis (approved):
sorry again for the late review.

PR #5192, opened 2025-05-07
docs: right align src button in api docs

### LINKED ISSUE #5191: right align the src button in the API docs?
In the past, the `[src]` button in the API docs was right aligned, I think this makes a simpler and clearer look.

I noticed that when comparing an [outdated version](https://fuzion-lang.dev/docs/String+(no+arguments)) of String with the [current one](https://fuzion-lang.dev/docs/base/String+(no+arguments)).

@tokiwa-software/developers what do you think?

simonvonhackewitz:
fix #5191

### DIFF
--- a/assets/docs/style.css
+++ b/assets/docs/style.css
@@ -215,3 +215,9 @@ Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
   display: inline-block;
   width: 3ch;
 }
+
+.heading-summary{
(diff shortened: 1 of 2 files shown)


michaellilltokiwa (commented):
This is hard for me to review. Can you refactor the code in `summary`, break it up into smaller parts, etc.? Also indentation might help.

michaellilltokiwa (approved):

PR #5200, opened 2025-05-09
lib: add uuid

### LINKED ISSUE #5164: add feature to generate UUID, rfc9562
The feature should generate a UUID according to https://www.ietf.org/rfc/rfc4122.txt
It should use the currently installed random effect.

simonvonhackewitz:
fix #5164


modules/base/src/uuid.fz
```diff
@@ -0,0 +1,202 @@
...
+# -----------------------------------------------------------------------
+
+
+# UUID as defined in RFC9562
+#
+# https://datatracker.ietf.org/doc/html/rfc9562
+#
+# NYI: UNDER DEVELOPMENT: some versions are missing
+#
+public uuid(private val u128) : property.orderable is
```
michaellilltokiwa, line 31:
```suggestion
private:public uuid(val u128) : property.orderable is
```

modules/base/src/uuid.fz
```diff
@@ -0,0 +1,202 @@
...
+
+  # Create a string representation of this UUID
+  # using lowercase letters
+  #
+  # e.g. f81d4fae-7dec-11d0-a765-00a0c91e6bf6
+  #
+  public redef as_string() String
```
michaellilltokiwa, line 38:
```suggestion
  public redef as_string String
```

michaellilltokiwa (commented):
This looks like a great starting point.
On general thing. There are a lot of _magic_ numbers in the source code. Would be good to write the code in a way that it is immediately clear what the number represents/ good for, e.g. by introducing names for each number, adding comments, adding references to the spec.

michaellilltokiwa (approved):

PR #5202, opened 2025-05-09
docs: shorten the navtitle in the api docs

simonvonhackewitz:
This is just a suggestion.

Changes
`API-Documentation: module base` -> `base`
`Fuzion Library Modules` -> `Library Modules`

e.g.
`Fuzion • Fuzion Library Modules • API-Documentation: module base` into
`Fuzion • Library Modules • base`
or
`Fuzion • Fuzion Library Modules • base • Any` into
`Fuzion • Library Modules • base • Any`

### DIFF
--- a/src/dev/flang/tools/docs/Html.java
+++ b/src/dev/flang/tools/docs/Html.java
@@ -518,12 +518,13 @@ public class Html extends ANY
    */
   private String headingSection(AbstractFeature f)
   {
-    return "<h1 class='$5'>$0</h1><h2>$3</h2><div class='heading-summary'>$1</div><div class='fd-comment'>$2</div>$6"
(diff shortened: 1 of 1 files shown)


src/dev/flang/tools/docs/Html.java
```diff
@@ -997,7 +998,7 @@ String modulePage()
   <nav style="display: none">$0</nav>
 </div>
 <div class="container">
-  <section><h1>Fuzion Library Modules</h1>
+  <section><h1 hidden>Library Modules</h1><h1>Fuzion Library Modules</h1>
```
michaellilltokiwa, line 1001: Why do you add these hidden h1s?

simonvonhackewitz: Because the first h1 from a page is used for the navtitle, so this seems like a simple way to have a shorter version in the navtitle without changing the headline

michaellilltokiwa (approved):

PR #5210, opened 2025-05-12
lib: on cleanup error trigger runtime fault

michaellilltokiwa:


modules/base/src/io/file/mapped_buffer.fz
```diff
@@ -54,5 +54,5 @@ module:public mapped_buffer
   # cleanup
   #
   public redef finally unit =>
-    # NYI: what to do on error?
-    _ := fzE_munmap memory length
+    if fzE_munmap memory length = -1
```
simonvonhackewitz, line 57: Is it intentional that you check for `=-1` here but for `!=0` in the other place?

modules/base/src/io/dir/open.fz
```diff
@@ -41,7 +41,8 @@ module:public open(dd Directory_Descriptor,
   # close dir when de-instating effect
   #
   public redef finally unit =>
-    _ := fzE_dir_close dd
+    if fzE_dir_close dd != 0
```
simonvonhackewitz, line 44: here

modules/base/src/io/file/mapped_buffer.fz
michaellilltokiwa: Yes, I agree we should make this more consistent. But as of now there are some inconsistencies left in the run time.

PR #5215, opened 2025-05-12
ast: add error when calling effect.finally from user code

michaellilltokiwa:
This is meant to ensures that `finally` is called exactly once when effect is deinstated.


tests/must_not_call_effect_finally/must_not_call_effect_finally.fz.expected_err
```diff
@@ -0,0 +1,13 @@
+
+--CURDIR--/must_not_call_effect_finally.fz:27:6: error 1: Must not call '<effect>.finally'.
+  lm.finally
+-----^^^^^^^
+'<effect>.finally' is called automatically when deinstating the effect.
```
simonvonhackewitz, line 5: We should probably decide on the terminology that we want to use for instating effects. I think the English word "deinstate" does not exist.

simonvonhackewitz (approved):

PR #5216, opened 2025-05-12
lib: simplify circular_buffer modded_mod

michaellilltokiwa:


modules/base/src/mutate/circular_buffer.fz
```diff
@@ -56,14 +56,10 @@ is
...
+  post debug: result >= 0
   =>
     b := c + 1
-    m := a % b
-
-    if m < 0
-      m + b
-    else
-      m
+    (a+b) % b
```
simonvonhackewitz, line 62: Can you be sure that `a + x` is not negative, i.e. that `a >= -b` ?

michaellilltokiwa: @simonvonhackewitz The post condition would catch this misuse.

PR #5234, opened 2025-05-14
ast: improve qualified name of abstract feature

### LINKED ISSUE #5224: Is the information ` (in type feature)` still necessary/useful?
It is printed in `toString` of the type. What I find confusing, is that it gets highlighted like it belongs to the type in the error messages.

```
❯ fz tests/reg_issue1739_this_type_in_type_feature_choice/issue1739.fz 

/home/simon/fuzion/tests/reg_issue1739_this_type_in_type_feature_choice/issue1739.fz:28:7: error 1: Incompatible types in assignment
      nil  # should flag an error: Incompatible types in assignment

assignment to field : 'issue1739.a.type.empty.result'
expected formal type: 'issue1739.a.this.type (in type feature)'
actual type found   : 'nil'
assignable to       : 'nil'
for value assigned  : 'nil'
To solve this, you could change the type of the target 'issue1739.a.type.empty.result' to 'nil' or convert the type of the assigned value to 'issue1739.a.this.type (in type feature)'.

one error.
```

michaellilltokiwa:
fixes #5224

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -467,15 +467,14 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
   {
     var tfo = state().atLeast(State.FINDING_DECLARATIONS) && outer() != null && outer().isCotype() ? outer().cotypeOrigin() : null;
     return
-      /* special type parameter used for this.type in type features */
-      isCoTypesThisType() ? (tfo != null ? tfo.qualifiedName(context) : "null") + ".this.type (in type feature)" :
-
-      /* cotype: use original name and add ".type": */
-      isCotype()             &&
-      cotypeOrigin() != null                 ? cotypeOrigin().qualifiedName(context) + ".type"
-
-      /* a normal feature name */
-                                                  : qualifiedName0(context);
+      isCoTypesThisType()
+        /* special type parameter used for this.type in type features */
+        ? (tfo != null ? tfo.qualifiedName(context) : "null") + ".this.type"
+        : isCotype() && cotypeOrigin() != null
+        /* cotype: use original name and add ".type": */
+        ? cotypeOrigin().qualifiedName(context) + ".type"
+        /* a normal feature name */
+        : qualifiedName0(context);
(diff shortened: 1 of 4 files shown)


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -467,15 +467,14 @@ String qualifiedName(AbstractFeature context)
...
-      /* a normal feature name */
-                                                  : qualifiedName0(context);
+      isCoTypesThisType()
+        /* special type parameter used for this.type in type features */
+        ? (tfo != null ? tfo.qualifiedName(context) : "null") + ".this.type"
+        : isCotype() && cotypeOrigin() != null
+        /* cotype: use original name and add ".type": */
+        ? cotypeOrigin().qualifiedName(context) + ".type"
+        /* a normal feature name */
+        : qualifiedName0(context);
```
simonvonhackewitz, line 477: I find the nested ternary operators a bit confusing. I think the readability could be improved by either changing the formatting or avoiding the nesting.

```suggestion
    return
      isCoTypesThisType()
        /* special type parameter used for this.type in type features */
        ? (tfo != null ? tfo.qualifiedName(context) : "null")
          + ".this.type"
        : isCotype() && cotypeOrigin() != null
          /* cotype: use original name and add ".type": */
          ? cotypeOrigin().qualifiedName(context) + ".type"
          /* a normal feature name */
          : qualifiedName0(context);
```

```suggestion
    if (isCoTypesThisType())
      {
        /* special type parameter used for this.type in type features */
        return (tfo != null ? tfo.qualifiedName(context) : "null") + ".this.type";
      }
    else
      {
        return isCotype() && cotypeOrigin() != null
          /* cotype: use original name and add ".type": */
          ? cotypeOrigin().qualifiedName(context) + ".type"
          /* a normal feature name */
          : qualifiedName0(context);
      }
```

michaellilltokiwa: @fridis does not like this style of having multiple returns. you would need to introduce a result field.

simonvonhackewitz (approved):

PR #5248, opened 2025-05-15
ast: isAssignableFrom return true even if boxing is required

michaellilltokiwa:
we have variants `isAssignableFromWithoutTagging/Boxing`


fridis:
would be good to document (or even check with a precondition) which of the `isAssignableFrom*` methods should be called in which phase, but this might require an additional arguments like the feature that contains the assignment whose state could be checked.

PR #5254, opened 2025-05-15
lib: unify syntax `t(arg)` => `(t arg)`

michaellilltokiwa:


modules/base/src/Sequence.fz
```diff
@@ -642,7 +642,7 @@ public Sequence(public T type) ref : property.countable is
   # ages := map (Sequence i32) human i32 (x -> x.age)
   # gt_ten := filter (Sequence i32) i32 (x -> x > 10)
   # xf := ages ∘ gt_ten
-  # say ([human(4), human(12), human(30)].into xf) # [12,30]
+  # say ([(human 4), (human 12), (human 30)].into xf) # [12,30]
```
fridis, line 645:
```suggestion
  # say ([human 4, human 12, human 30].into xf) # [12,30]
```

fridis (changes requested):
Good, but found some parentheses that can be removed.

fridis (approved):

PR #5273, opened 2025-05-19
lib: add `concur.Channel`

michaellilltokiwa:
a Channel can be used to send values to another thread

### DIFF
--- a/modules/base/src/concur/Receiving_Channel.fz
+++ b/modules/base/src/concur/Receiving_Channel.fz
@@ -27,5 +27,5 @@ public Receiving_Channel(T type) ref is
 
   public has_next option T => abstract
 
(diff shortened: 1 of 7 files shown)


modules/base/src/concur/Receiving_Channel.fz
```diff
@@ -0,0 +1,31 @@
...
+# -----------------------------------------------------------------------
+
+# an interface for a receive only Channel
+#
+public Receiving_Channel(T type) ref is
+
+  public has_next option T => abstract
+
+  public type.select(ch array (Channel T)) option T =>
+    panic "NYI"
```
michaellilltokiwa, line 31:
```suggestion
  # public type.select(ch array (Channel T)) option T =>
  #   panic "NYI"
```

PR #5280, opened 2025-05-20
ast: fix #5263 by creating an error message that explains partial application

### LINKED ISSUE #5263: Confusing error when calling `infix <|`
Try this
```
 > ./build/bin/fz -e 'say <| .byte_length <| "hi"'

command line:1:5: error 1: Different count of arguments needed when calling feature
say <| .byte_length <| "hi"

Feature not found: 'infix <|' (3 arguments)
Target feature: 'universe'
In call: '<|'
To solve this, you might change the actual number of arguments to match the feature 'infix <|'  (2 type arguments, 2 value arguments) at {base.fum}/pipes.fz:162:8:
public infix <| (A, R type, f A->R, a A) R =>
To call 'infix <|' you must provide 4 arguments. The type arguments may be omitted or `_` may be used in place of a type argument.


one error.
```
It is clear that this currently does not work due to `<|` being left-associative (#4253), but I do not see how this small example calls `infix <|` with 3 arguments. Maybe the error should list the arguments that where found to clarify this.

fridis:
Also added a regression test for this error.

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -157,7 +157,7 @@ public class AstErrors extends ANY
   }
   static String s(Expr e)
   {
-    return expr(e.sourceRange().sourceText());
+    return expr(e.sourceText());
   }
   static String s(AbstractAssign a)
   {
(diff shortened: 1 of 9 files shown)


maxteufel (changes requested):
Please update the failing tests accordingly.

maxteufel (approved):

PR #5291, opened 2025-05-22
lib: net, move read/write-handler from channel to connection

michaellilltokiwa:


michaellilltokiwa:
> Looks good. Do you have any rationale for this?

It seemed cleaner to me to define these handlers closer to their actual usage.

PR #5293, opened 2025-05-22
ast: AbstractType replace refOrVal by mode

michaellilltokiwa:
`mode` of a type is one of `ThisType`, `RefType`, `ValueType`.
`AbstractType.mode` replaces the methods `isRef` and `isThisType`.
For an unresolved type ValueType is returned by mode().

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -699,7 +699,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
 
     if (POSTCONDITIONS) ensure
       (selfType != null,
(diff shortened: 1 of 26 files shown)


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -1895,6 +1895,12 @@ && valueArguments()
   }
 
 
+  public TypeMode defaultTypeMode()
+  {
```
michaellilltokiwa, line 1899:
```suggestion

  /**
   * @return RefType if Feature is a reference otherwise ValueType
   */
  public TypeMode defaultTypeMode()
  {
```

PR #5300, opened 2025-05-22
ast, base: Support for right-associative operators

fridis:
A new keyword `infix_right` now permits the definition of right-assiciative operators.

First, the parser now always parsed infix operator calls as if they were right-assciative. During `Call.resolveTypes()`, this is fixed by a new method `fixAssociativity` that is redefined in `ParsedOperatorCall` to do its magic.

Code for chained booleans `a < b < c < d` had to be changed to work with the changed associativity created by the parser.

Also improved `AbstractCall.toString()` to use parentheses for better debug output.

`infix <|`, `infix <||`, etc. are now right-associative such that you can write

    say <| +"!" <| "Hello, "+ <| "World"

instead of

    say <| (+"!" <| ("Hello, "+ <| "World"))


maxteufel (approved):
Looks nice! `fixAssociativity` still feels like a hack to me, but I don't see any better solution either.

PR #5307, opened 2025-05-27
base: rename `infix :` in `list.fz` as `infix_right :`

fridis:
This permits creating lists like `1 : 2 : 3 : nil` without parentheses


maxteufel (approved):

modules/base/src/list.fz
```diff
@@ -713,12 +713,16 @@ public list(T type, h T, t Lazy (list T)) list T =>
 #
 #   0 : [1,2,3,4].as_list
 #
+# or
+#
+#   0 : 1 : 2 : 3 : 4 : nil
+#
 # or to create lists by recursion, e.g., a an endless list containing
 # integer 1 repeatedly is
```
fridis, line 721:
```suggestion
# or to create lists by recursion, e.g., an endless list containing
# integer 1 repeatedly can be created as follows
```

PR #5312, opened 2025-05-30
Fix 1015a, experimental using `Nullary` instead of `Lazy`

fridis:


fridis:
merged this into #5308

PR #5325, opened 2025-06-04
lib: add Effect interface and Sequence.instate_all

michaellilltokiwa:


michaellilltokiwa:
approach not feasible since DFA can not know about which effects are instated

PR #5333, opened 2025-06-04
dfa: Two phase DFA

fridis:
Run DFA in two phases: First, to collect calls in `CallGroups` with common call destination, call site and call target to find all environments required by this call group.

Second, to run the analysis tracing effect environments as well, but limiting this to the environments that are actually required as found in the first phase for the group each call belongs to.

For this to work, the following changes were needed:

- `effect.instate0` is now called by `instate_helper` such that distinguishing the `instate_helper` instances is no longer needed to distinguish the effect environments
- `Java_String.utf8` no longer uses `fuzion.jvm.env` since the analysis is no longer accurate to trace the presence of the `fuzion.jvm` effect. This may lead to errors at runtime if a `Java_String` is used outside of the jvm effect
- added `FUIR.clazzDeclarationPos()` for better error output for Environments, implemented `FUIR.clazzSrcFile` using `FUIR.clazzDeclarationPos()`.
- added new class `CallGroup` to goup calls with same environment
- In a first phase, all the `CallGroup` and the effects that are used or possibly available in each `CallGroup` are determined. Then, DFA is restarted from scratch adding effect environments to calls, but only for those effects that are both used and possibly available by the corresponding `CallGroup`. 
- A new env var / property `dev.flang.fuir.analysis.dfa.TRACE_ALL_EFFECT_ENVS` can be used to enforce full effect environments for all calls independent on what was found in the `CallGroup`
- A test `tests/tricky_dfa_cases` was added that contains cases that required full effect environments.


maxteufel (changes requested):
Putting a blocker on this, as discussed per internal mailing list. This pull request currently increases the duration of the DFA for the webserver.

maxteufel (approved):
Performance is now fixed in the webserver again!

PR #5339, opened 2025-06-05
port check_simple_example to fuzion

michaellilltokiwa:
caveat currently: setting ulimit is not done yet.

### DIFF
--- a/tests/check_simple_example.fz
+++ b/tests/check_simple_example.fz
@@ -229,7 +229,7 @@ delete_tmp_files =>
 
 be     := envir.args[1]
 # can not execute bash script fz on windows with CreateProcess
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
@maxteufel Can you also try this please:
```diff
diff --git a/tests/check_simple_example.fz b/tests/check_simple_example.fz
index 5e6b8e092..58795ac6c 100644
--- a/tests/check_simple_example.fz
+++ b/tests/check_simple_example.fz
@@ -80,6 +80,7 @@ envir_vars :=
       .map x->x.1
       .as_array))
   .add "FUZION_DISABLE_ANSI_ESCAPES" "true"
+  .add "LC_CTYPE" "en_GB.UTF-8"
 
 # helper to get out, err, exit_code from started process
 #
```

michaellilltokiwa:
@maxteufel I think this might work now since I replaced calling `fz` by calling `java`directly. Reason being that we cannot execute `fz` on windows via `CreateProcess` which is used by `fzE_process_create`

michaellilltokiwa:
@maxteufel I think this is due to java doing some detection as to what exactly is stdout and then does some charset conversions. In the fz/bash case it is  the one from shell/terminal, in the direct case it is a standard pipe.

PR #5386, opened 2025-06-17
lib: fix precondition of `container.ordered_map`

simonvonhackewitz:
Without this it was possible to create a map, that when printed would show
`(1 => one), (2 => two), (3 => drei), (3 => three), (4 => four), (5 => five)
`

### DIFF
--- a/modules/base/src/container/ordered_map.fz
+++ b/modules/base/src/container/ordered_map.fz
@@ -38,7 +38,7 @@ public ordered_map(
            k0 Sequence OK,
            v0 Sequence V) : container.Map OK V
 pre
-  k0.count = v0.count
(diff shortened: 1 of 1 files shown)


modules/base/src/container/ordered_map.fz
```diff
@@ -39,6 +39,7 @@ public ordered_map(
            v0 Sequence V) : container.Map OK V
 pre
   k0.count = v0.count
+  debug: k0.unique.count = k0.count
```
michaellilltokiwa:
```suggestion
  debug: k0.count = v0.count
  debug: k0.unique.count = k0.count
```

michaellilltokiwa (approved):

PR #5395, opened 2025-06-18
cleanup: move boxing/tagging to IR

michaellilltokiwa:
Removes boxing/tagging from frontend, also boxing/tagging is not written to fum-file anymore.


src/dev/flang/ast/Expr.java
```diff
@@ -881,6 +696,149 @@ boolean isTypeAsValueCall()
...
+
+
+  /**
+   * @param expr the expr to be boxed
+   *
+   * @param frmlT the formal type
+   *
+   * @return the potentially boxed type
+   */
+  private static AbstractType boxedType(Expr expr, AbstractType frmlT)
```
fridis, line 727: This should IMHO not be `static`, the `expr` argument should be the target `this`.

src/dev/flang/ast/Expr.java
```diff
@@ -881,6 +696,149 @@ boolean isTypeAsValueCall()
...
+        if (frmlT.isChoice() && frmlT.isAssignableFrom(boxedType).yes())
+          {
+            checkTagging(this, boxedType, frmlT);
+          }
+      }
+  }
+
+
+  /**
+   * @param expr the expr to be boxed
```
fridis, line 721: This needs some comment, what is `frmlT` used for?

src/dev/flang/ir/IR.java
```diff
@@ -337,6 +340,204 @@ else if (e instanceof Universe)
...
+          }
+        if (frmlT.isChoice() && frmlT.isAssignableFrom(result.type()).yes())
+          {
+            result = tag(result, frmlT);
+            if (CHECKS) check
+              (needsBoxing(result, frmlT) == null);
+          }
+      }
+    // NYI: ugly special case: currently needed for code like
+    // because isAssignableFrom does not return yes without correct Context...
```
fridis, line 376: I guess the `code like` refers to the code underneath this comment. Should be rearranged.

src/dev/flang/ir/IR.java
```diff
@@ -337,6 +340,204 @@ else if (e instanceof Universe)
...
+
+
+  /**
+   * Is boxing needed when we assign to frmlT?
+   *
+   * @param frmlT the formal type we are assigning to.
+   *
+   * @return the type after boxing or null if boxing is not needed
+   */
+  private static AbstractType needsBoxing(Expr expr, AbstractType frmlT)
```
fridis, line 500: This appears to do the same as `Expr.boxedType()`, but slightly differently. It would IMHO be good to move this to `Expr` and to merge `boxedType` and `needsBoxing` into one method or to move the core logic to a common helper method. 

Or, since `Expr.boxedType()` seems to be used only in `Expr.checkAmbiguousAssignmentToChoice`, maybe we can avoid `Expr.boxedType()` altogether and use `needsBoxing` instead for this check?

tests/reg_issue5268/reg_issue5268.fz.expected_err
```diff
@@ -1,21 +1,66 @@
...
+--CURDIR--/reg_issue5268.fz:55:3: error 5: Unused, non public field 'z'
+  z  (unit,f64,String)->i32 := !?!4711    # should work
+--^
+To solve this, do either of the following
+  - use the field
+  - set it to 'public'
+  - explicitly ignore the result by using '_' instead of 'z'
+
+
+--CURDIR--/reg_issue5268.fz:37:18: error 6: Could not find called feature
```
fridis, line 47: Pleaser add some calls like `ignore v` to this test such that we can see that the cases that should work actually do and those that should flag an error do so as well?

fridis (changes requested):
Nice cleanup and getting rid of one compilation phase! 

I added some minor comments and found a possible potential to avoid code duplication.

tests/reg_issue5268/reg_issue5268.fz.expected_err
michaellilltokiwa: I think I can/should not do that since in the test:
```
  # the test cases are all declared twice with only the second version being
  # actually called below.
  #
  # The reason is that the fact that there was an actual call had an effect on the
  # errors that were produced before #5268 was fixed.
```

fridis (approved):
looks nice.

PR #5402, opened 2025-06-20
docs: limit annotation `[Contains abstract features]` to current or base module

simonvonhackewitz:
Currently, every feature in the docs has the annotation because 0e9b57cdd9fc61488a7df03abc2e7cfd6d539a92 in module `json_encode` introduced an abstract feature in `Any`.

This will only show the annotation "[Contains abstract features]" if the abstract feature is in the same module (for which the docs are displayed) or in the base module. This way, [`String` in module `terminal`](https://fuzion-lang.dev/docs/terminal#String_0) will still have the annotation. But it will not work if a module depends on a module other than base.

I think there might be a bug in `Module.declaredOrInheritedFeatures()`, so that it wrongly includes features from other modules.

### DIFF
--- a/src/dev/flang/tools/docs/Docs.java
+++ b/src/dev/flang/tools/docs/Docs.java
@@ -375,7 +375,7 @@ public class Docs extends ANY
     // generate documentation per module
     for (var module : all_modules)
     {
-        var htmlTool = new Html(config, mapOfDeclaredFeatures, universe, module, all_modules);
+        var htmlTool = new Html(config, mapOfDeclaredFeatures, universe, module, all_modules, fe);
 
         mapOfDeclaredFeatures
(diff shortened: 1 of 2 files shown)


fridis:
> > I think there might be a bug in `Module.declaredOrInheritedFeatures()`, so that it wrongly includes features from other modules.
> 
> I think this is the way it should work, comment says "... as seen by this module...", so I would expect features from modules that the module depends on.

Sounds to me as if the `Module` instance that is used in a call to `Module.declaredOrInheritedFeatures()` by the documentation generator is wrong (i.e., the main module instead of the module we are creating the docs for). Could that be the case?

simonvonhackewitz:
superseded by #5465

PR #5403, opened 2025-06-20
docs: fix links of base module feature displayed in other modules

simonvonhackewitz:
The link of [`String` in module `terminal`](https://fuzion-lang.dev/docs/terminal#String_0) incorrectly points to the [`base` module page of `String`](https://fuzion-lang.dev/docs/base/String+%28no+arguments%29/) but it should point to the [version in `terminal`](https://fuzion-lang.dev/docs/terminal/String+(no+arguments)/) that only lists the features that `terminal` adds to `String`.

### DIFF
--- a/src/dev/flang/tools/docs/Html.java
+++ b/src/dev/flang/tools/docs/Html.java
@@ -362,7 +362,7 @@ public class Html extends ANY
 
   private String anchor(AbstractFeature af) {
     return "<div class='font-weight-600 ml-2'>"
-            + (noFeatureLink(af) ? "" : "<a class='fd-feature' href='" + featureAbsoluteURL(af) + "'>")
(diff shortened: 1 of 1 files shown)


src/dev/flang/tools/docs/Html.java
```diff
@@ -362,7 +362,7 @@ private boolean isVisible(AbstractFeature af)
 
   private String anchor(AbstractFeature af) {
     return "<div class='font-weight-600 ml-2'>"
-            + (noFeatureLink(af) ? "" : "<a class='fd-feature' href='" + featureAbsoluteURL(af) + "'>")
```
michaellilltokiwa, line 365: Can `featureAbsoluteURL` be removed then?

PR #5412, opened 2025-06-23
ast: change AbstractType.toString for this-types

michaellilltokiwa:
now `a.this.b.this` is just `a.b.this`

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -2067,8 +2067,6 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
       }
     else
       {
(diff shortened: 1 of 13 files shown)


michaellilltokiwa:
@maxteufel I talked to @fridis over lunch about this. For a this-type the outers do not bring further any info to the table. So this is just about how we want to _a_ print them and _b_ want them resolved. The resolving of fully qualified this-types does not yet work btw., see QualThisType where all but last parsed name is just thrown away:
```
  public QualThisType(List<ParsedName> qual)
  {
    super(SourcePosition.range(qual),
          qual.getLast()._name,
          Call.NO_GENERICS, null, Optional.of(TypeMode.ThisType));
  }
```

PR #5454, opened 2025-07-02
ast: cleanup, use Types.resolved.unitCall instead of new Call

michaellilltokiwa:


michaellilltokiwa:
Turns out we need the source position of the unit call sometimes...

PR #5471, opened 2025-07-09
fuir: apply deMorgan to make logic easier to read

michaellilltokiwa:
Also removed useless comment. A comment should explain the why not the what.


fridis:
I admit that the comment is useless. Instead, it would be good to have a comment that explains why a constructor should not be specialized, this is not obvious to me (even though I most likely wrote that code). Would be good to figure that out...

PR #5491, opened 2025-07-10
lib: update comment

### LINKED ISSUE #5489: `complex.type.lteq` comment fix
This was introduced in #5336: The comment inside `lteq` should replace the one above this feature, since the latter one is wrong now.

michaellilltokiwa:
fixes #5489

### DIFF
--- a/modules/base/src/num/complex.fz
+++ b/modules/base/src/num/complex.fz
@@ -58,7 +58,7 @@ is
   # see ORDERING COMPLEX NUMBERS. . . NOT* by David Angell:
   # https://web.maths.unsw.edu.au/~angell/articles/complexorder.pdf
   #
-  # NOTE: order is not compatible with addition/multiplication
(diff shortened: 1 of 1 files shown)


modules/base/src/num/complex.fz
```diff
@@ -53,16 +53,14 @@ is
     a.real = b.real && a.imag = b.imag
 
 
-  # total order ignoring imag
+  # lexicographic order
+  #
+  # see ORDERING COMPLEX NUMBERS. . . NOT* by David Angell:
+  # https://web.maths.unsw.edu.au/~angell/articles/complexorder.pdf
+  #
+  # NOTE: order is not compatible with addition/multiplication
```
simonvonhackewitz, line 61: In the pdf it says
> In fact, the lexicographic order is compatible with addition, but not with
multiplication.

simonvonhackewitz (changes requested):

PR #5492, opened 2025-07-10
docs: use relative paths in API docs when full HTML is generated

simonvonhackewitz:
so the API doc HTML files can be directly opened and used in the browser without a server


michaellilltokiwa:
Good, but it seems there is an NPE still in there:
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.toString()" because "replacement" is null

PR #5513, opened 2025-07-14
lib: fix `substring_codepoint`

michaellilltokiwa:


michaellilltokiwa:
Good example why we need coverage for base-lib :-)

PR #5515, opened 2025-07-14
ast: remove hack in applyTypePars_

michaellilltokiwa:


michaellilltokiwa:
@maxteufel No, I think the reasons are other changes related to this-types. I'm not sure which exactly though.

PR #5528, opened 2025-07-16
ast: add comment and rename `_unresolvedGenerics`

michaellilltokiwa:


michaellilltokiwa:
@maxteufel I think it is better to rename in another PR.

PR #5536, opened 2025-07-16
parser: unify handling of `()` and `{}` with respect to indentation, fix #4425

### LINKED ISSUE #4425: AoC: Different handling of indentation of code in `()` vs `{}` is confusing
I ran into this during [6 dec 24 Advent of Code](https://github.com/fridis/fuzion_aoc/tree/main/2024/06):

This code
```
ignore i32->bool p->{if p = 3 then
                       say "hi"
                     false}
```
works as expected
```
 > ~/fuzion/work/build/bin/fz indentation_error3.fz 
```
while the same code using `()`
```
ignore i32->bool p->(if p = 3 then
                       say "hi"
                     false)
```
fails parsing and produced indentation errors and type errors caused by the parser producing a wrong AST:
```
 > ~/fuzion/work/build/bin/fz indentation_error2.fz 

/home/fridi/fuzion/advent_of_code/fuzion_aoc/2024/06/indentation_error2.fz:3:22: error 1: Inconsistent indentation
                     false)

Indentation reference point is /home/fridi/fuzion/advent_of_code/fuzion_aoc/2024/06/indentation_error2.fz:1:22:
ignore i32->bool p->(if p = 3 then

While parsing: klammer, parse stack: klammer, bracketTerm, term, opTail, opExpr, operatorExpr (twice), expr, exprs, block (twice), lambda, plainLambda, callOrFeatOrThis, term, opTail, opExpr, operatorExpr (twice), actualSpace, actualSpaces, actualArgs, call (twice), callOrFeatOrThis, term, opTail, opExpr, operatorExpr (twice), expr, exprs, block (twice), unit


/home/fridi/fuzion/advent_of_code/fuzion_aoc/2024/06/indentation_error2.fz:2:24: error 2: Incompatible types in assignment
                       say "hi"

assignment to field : 'λ.call.result'
expected formal type: 'bool'
actual type found   : 'unit'
assignable to       : 'unit'
for value assigned  : 'say "hi"'
To solve this, you could change the type of the target 'λ.call.result' to 'unit' or convert the type of the assigned value to 'bool'.


/home/fridi/fuzion/advent_of_code/fuzion_aoc/2024/06/indentation_error2.fz:1:22: error 3: Incompatible types in assignment
ignore i32->bool p->(if p = 3 then

assignment to field : 'λ.call.result'
expected formal type: 'bool'
actual type found   : 'unit'
assignable to       : 'unit'
for value assigned  : ''
To solve this, you could change the type of the target 'λ.call.result' to 'unit' or convert the type of the assigned value to 'bool'.

3 errors.
```

fridis:
Now, `{`  and `(` can be used largely interchangably around code block, e.g., we can create a block as follows

    say (if 3 = 4 then
           say "strange"
         else
           say "this should run 1"
         "result from code block 1")

which is equivalent to the same code using `{`/`}`:

    say {if 3 = 4 then
           say "strange"
         else
           say "this should run 2"
         "result from code block 2"}

Simplified the grammar rules for `klammer`, `tuple` and `lambda`.

There are still some minor syntactical differences if one a single `operatorExpr` is put in `(`/`)` vs. `{`/`}`, in particular

    _ := l.zip m (a,b -> unit)  # with {}, would be parsed as declaration of `a` and `b` and not lambda
    _ := ("bla"
           + "blub")            # with {}, causes indentation error
    _ := (a).this               # with {}, causes qualifier expected for '.this' expression.

I will leave these for later to be fixed.

Also, there are still differences when parsing actual arguments in parentheses vs. a tuple with respect to the handling of commas: `(x,y->unit,z)` would be parsed as three actual arguments (`x`, `y->unit` and `z`), but as a two-tuple (cosisting of `x, y->unit` and `z`).

Add a regression test for #4425.

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -59,6 +59,20 @@ public class Parser extends Lexer
   }
 
 
+  /**
+   * Exception used for control flow to exit a call.
+   */
+  static class GiveUp extends RuntimeException
+  {
+
+    /**
+     * Pre-allocated instance of `GiveUp`
+     */
+    static GiveUp _INSTANCE_ = new GiveUp();
+
+  };
+
+
   /*----------------------------  constants  ----------------------------*/
 
 
@@ -1343,7 +1357,7 @@ indexTail   : ":=" exprInLine
     do
       {
         SourcePosition pos = tokenSourcePos();
-        var l = bracketTermWithNLs(BRACKETS, "indexCall", () -> actualCommas());
+        var l = bracketTermWithNLs(BRACKETS, "indexCall", () -> actualCommas(true));
         String n = FuzionConstants.FEATURE_NAME_INDEX;
         if (skip(":="))
           {
@@ -1556,7 +1570,7 @@ actualArgs  : actualSpaces
   {
     return (ignoredTokenBefore() || current() != Token.t_lparen)
       ? actualSpaces()
-      : bracketTermWithNLs(PARENS, "actualArgs", () -> actualCommas());
+      : bracketTermWithNLs(PARENS, "actualArgs", () -> actualCommas(true));
   }
 
 
@@ -1652,6 +1666,12 @@ actualArgs  : actualSpaces
   /**
    * Parse actualCommas
    *
+   * @param endAtComma true to treat `x, v->2*v` as two actuals `x` and `v->2*v`
+   * instead of one `(x,v)->2*v`.
+   *
+   * NYI: CLEANUP: It might be better to omit the case endAtComma==true and to always
+   * parse the case `x, v->2*v` as one actual.
+   *
 actualCommas: actualSome
             |
             ;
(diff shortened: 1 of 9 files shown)


src/dev/flang/parser/Parser.java
```diff
@@ -1893,54 +1913,88 @@ Expr opsTail(OpExpr oe)
...
-      {
-        return lambda(tupleElements);
-      }
-    else if (tupleElements.size() == 1)    // an expr wrapped in parentheses, not a tuple
-      {
-        var e = tupleElements.get(0);
-        if (e instanceof ParsedOperatorCall oc)
+    Expr res;
+    var f = fork();
+    var t = f.tuple();
```
michaellilltokiwa, line 1925: I feel like the code might be cleaner if `GiveUp` was caught here and tuple never returns null?

michaellilltokiwa: @fridis True, but exception is thrown in lambda, so my thinking is we are kind of throwing it another method anyway.

PR #5553, opened 2025-07-18
cleanup: move BoolConst handling to Match.createIf

michaellilltokiwa:


src/dev/flang/ast/Match.java
```diff
@@ -422,48 +422,59 @@ else if (c instanceof Block b && b.resultExpression() instanceof Match m)
...
-     */
-    if (elseB == null)
+    Expr result;
+    if (c instanceof BoolConst bc)
       {
-        var unit = new Call(pos, FuzionConstants.UNIT_NAME);
-        elseB = new Block(new List<>(unit));
+        result = bc.getCompileTimeConstBool()
+          ? b
+          : elseB;
```
fridis, line 436: I think `elseB` might be `null` here, which then might end up being returned as a result. Or am I missing something?

src/dev/flang/ast/Call.java
```diff
@@ -2793,17 +2793,11 @@ void checkTypes(Resolution res, Context context)
...
-  private Expr newIf(Expr cc, Expr block, Expr elseBlock)
+  private Expr createIf(Expr cc, Expr block, Expr elseBlock)
   {
-    return !(cc instanceof BoolConst bc)
-      ? Match.createIf(pos(), cc, block, elseBlock, false)
-      : bc.getCompileTimeConstBool()
-      ? block
-      : elseBlock;
+    return Match.createIf(pos(), cc, block, elseBlock, false);
   }
```
fridis, line 2801: Maybe this helper method is no longer needed.

fridis (changes requested):
Nor sure if the new code could result in `null` being returned by `createIf`, see comments in the code.

src/dev/flang/ast/Match.java
michaellilltokiwa: Ugh, I missed that...

PR #5554, opened 2025-07-18
ast: make sure freetypes are split off when creating cotype

michaellilltokiwa:


fridis:
Looks reasonable, but do you have an example where the old code does not work? Or is this a change in anticipation of future changes?

michaellilltokiwa:
@fridis I could not find an example that does not work. But my reasoning was that cotypes should only be created once we are sure that free types are split off, hence the change.

PR #5562, opened 2025-07-18
ast: In `AstErrors.declarationsInLazy` do not use `Expr.toString()`

fridis:
For error output, there is `AstErrors.s(Expr)` which is much nicer.


michaellilltokiwa (approved):
Yes, much nicer. I did not realize only such a small change was necessary.

PR #5568, opened 2025-07-21
base: Add `time.Clock` and `time.nano.run_once`/`time.nano.run_periodic`.

fridis:
This depends on PRs #5606 and #5607. 

This adds an abstract `ref` effect `Clock` that is then implemented by `time.nano` and could be implemented by more different clocks. 

Clock permits reading time and running code at a given times and periodically as follows
```
test_clock(C type : time.Clock) =>

  _ := C.clock.run_once (C.clock.read + time.duration.ms 500) (say "once!")

  start := (C.clock.read + time.duration.ms 500)
  period := time.duration.ms 100
  end := start + period * 2
  _ := C.clock.run_periodic start period end ()->
    say "thrice!"

time.nano ! ()->
  test_clock time.nano
```
The idea is to add more clock related features like interrupting an ongoing `sleep`, which would then make it possible to use different Clocks to control a thread or a whole thread pool to accept incoming jobs with different release times.


modules/base/src/time/Clock.fz
```diff
@@ -0,0 +1,101 @@
...
+  # run the given code once at the provided instant within the current thread
+  #
+  # In case the instant is in the past, do not run the code. In case the instant is
+  # in the future, wait until this clock reached the time specified by instant.
+  #
+  # return true if the code was actually run, false if the given time `t` is already
+  # in the past.
+  #
+  public run_once(t instant,
+                  code ()->unit) bool
```
michaellilltokiwa, line 67: Not sure if a good idea, but these could be Lazy instead.

michaellilltokiwa (approved):

modules/base/src/time/Clock.fz
fridis: `Lazy unit` and `()->unit` is essentially the same, but `Lazy` focuses more on the value than on the code, while here the code is IMHO the important thing.

PR #5582, opened 2025-07-22
ast: call remove seemingly unecessary resolve of generics

michaellilltokiwa:


michaellilltokiwa:
> I would prefer if you added a slight bit more of text/reasoning explaining why this might be unnecessary. Let's just see however if this also works with the extended tests.

I'm not 100% sure but I think the context should always be sufficient to resolve its generics.
On the other hand I don't understand why one would need the called features outer as context to resolve the call.
So I removed this code where I did not understand what the reason for it might be.

PR #5588, opened 2025-07-23
fe: remove override NormalType.replaceGenericsAndOuter

michaellilltokiwa:


fridis (commented):
please close, see @5591.

PR #5591, opened 2025-07-23
ast/fe/fuir: cleanup: resolved types are not a concept for the middle end

fridis:
Resolving of features and types happens during he front end phase only, so anything related to resolution should no longer be touched in the middle end. 

Replace dummy implementation of `AbstractType.replaceGenericsAndOuter` since this is redefined in `ast.ResolvedNormalType` and `fe.NormalType`

Moved `Clazz.asValue` to `AbstractType.asValue` to avoid the imports for `ResolvedType` in `Clazz`, the resolution is a front-end thing that should be unknown to the middle-end.


michaellilltokiwa (approved):
Better than my approach.

PR #5601, opened 2025-07-25
revert cfe45d7, since this does not work on windows

michaellilltokiwa:
Comment was, does it still apply?:
```
When doing a 'run_tests', the paths in the error output contained
'--CURDIR__/../modules', where '/../modules' was redundant and caused an error.
```


michaellilltokiwa:
The workaround does not seem to be necessary anymore.

PR #5627, opened 2025-07-30
Base time add histogram

fridis:
Add infrastructure to record and output timing histograms to analyze jitter in release or execution times. Added `examples/jitter` to measure the jitter in periodic code run via `time.nano.run_periodic` while decreasing the period until we the release time latency exceeds the period or we get an overrun into the next release. 

This measures the actual release times relative to the expected release times of one earlier iteration, i.e., ideally, this time is equal to the period, but we see jitter. 

The output are graphs like this:

                   ---  period  976µs, log count / linear time  ---               
    count
      1K                                 *                                 
    512                                  *                                 
    256                                  *                                 
    128                                  **                                
     64                                  ***                               
     32                                 ****                               
     16                                 ****                               
      8                                 ****                               
      4                                *****                               
      2                                *****                               
      1                                ******                              
         |       |       |       |       |       |       |       |       |       |
        30µs   270µs   510µs   751µs   991µs   1231µs  1472µs  1712µs  1953µs    >
    
      979µs  |  928µs  | 1055µs  |   14µs  |  2149   |

For this histogram, the actual times measured are divided into buckets for time ranges and the total number per bucket is counted and then shown. Also, the _average_, _min_-imum, _max_-imum times are shown as well as the standard deviation _sigma_ and the total _count_ of releases. 

This graphs shows that the desired period is met relatively often, but there were frequent outliers of about 100µs in both directions. 

This PR requires PRs #5637 and #5638 to be merged first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1373,11 +1373,11 @@ spellcheck:
 	bin/spell_check_java.sh
 
 # target to do a syntax check of fz files.
-# currently only examples/ are checked.
+# currently only code in bin/ and examples/ are checked.
 .PHONY: syntaxcheck
 syntaxcheck: min-java
-	find ./examples/ -name '*.fz' -print0 | xargs -0L1 $(FZ) -modules=clang,java.base,java.datatransfer,java.xml,java.desktop -noBackend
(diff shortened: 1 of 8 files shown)


michaellilltokiwa (changes requested):
Looking forward to working with this!

fridis (commented):
I will still update the code.

michaellilltokiwa (approved):

PR #5629, opened 2025-07-31
http: change visibility of `canonical_header_map.map_of`

maxteufel:
This has shown to be very useful when working with the module's API.


michaellilltokiwa (changes requested):
`container.Map.add` feature should be moved to `container.Map`

michaellilltokiwa (approved):

PR #5631, opened 2025-07-31
ast: replace `DotType` by simple `type_as_value` calls

michaellilltokiwa:


fridis (approved):
nice. Just some minor suggestions to improve comments.

PR #5641, opened 2025-08-01
lib: fix `state` effect does not work when instated using `infix !`

### LINKED ISSUE #5634: `state` effect does not work when instated using `infix !`
Take this example 
```
state i32 unit (i32 23) ()->
  _ := state_modify i32 *2
  say <| state_get i32
```
which works as expected
```
 > ../../build/bin/fz x0.fz
46
```
while the version using `infix !`
```
state i32 (i32 23) ! ()->
  _ := state_modify i32 *2
  say <| state_get i32
```
does not modify the state on a call to `state_modify`
```
 > ../../build/bin/fz x1.fz
23
```

michaellilltokiwa:
fixes #5634


modules/base/src/state.fz
```diff
@@ -72,13 +78,17 @@ is
...
+    res
 
 
   # set state to new, leaving the contents unchanged
   #
   public put (new S) state T S =>
-    state val new mode.next
+    res := state val new mode.next
+    res.replace
+    res
```
fridis, line 91: This is correct, but what puzzles me is that `mode.next` should return `repl` such that the constructor of `oneway_monad`, which is called by the constructor of `state` should already have done a call to `replace`. 

So there seems to be some problem with that mechanism, i.e., the constructor of `oneway_monad` either does not call `replace`, or calls it with the wrong effect type or the old state value.

fridis (commented):
See my comment below, the calls to `replace` should originally be implicitly done by `oneway_monad`.

PR #5651, opened 2025-08-01
ast: hasOuterRef add !isTypeParameter

michaellilltokiwa:
Found that outer refs are created for type parameters which seems weird and unnecessary.
Especially for `THIS#TYPE`  (this type in cotype)


fridis (approved):
good. I think I saw these strange outer refs as well, but did not fix that.

PR #5661, opened 2025-08-05
linux: fix failing jvm tests

michaellilltokiwa:
#
    # AI claims the following:
    #
    # On Linux, the JVM often does not automatically infer the classpath,
    # especially when launched via JNI_CreateJavaVM, so you must explicitly
    # provide it via -Djava.class.path= — otherwise your Java code may not
    # be found or load properly.
    #
    # Best Practice:
    #
    # Always set -Djava.class.path= explicitly when using JNI_CreateJavaVM,
    # even if it seems to work on your system.
    #

### DIFF
--- a/modules/base/src/fuzion/jvm.fz
+++ b/modules/base/src/fuzion/jvm.fz
@@ -34,7 +34,7 @@ private:public jvm : effect is
   # short-hand to for jvm with no additional class path
   #
   public type.use(T type, code ()->T) T =>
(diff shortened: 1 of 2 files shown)


michaellilltokiwa:
You can see the failures in the linux github runs.

michaellilltokiwa:
@maxteufel Good catch. Weird because I can still reproduce the problem on my machine.

PR #5677, opened 2025-08-07
fuir: cleanup remove `outer()`

michaellilltokiwa:
It seems unnecessary since `_outer` always equal to `outerRef.resultClazz` and _outer being a final field.


fridis:
Always nice to see some code go!

PR #5682, opened 2025-08-07
add missing dependencies for `make all` to README

### LINKED ISSUE #4985: Update dependencies in README
There are some dependencies missing, at least `antlr4` and maybe `asciidoc` and/or `asciidoctor`.

simonvonhackewitz:
fix #4985

[skip ci]


michaellilltokiwa:
I think we should probably rather change `ebnf.fz` to fail more gracefully when missing pcregrep and or antlr, by e.g. testing for exit code 127.

PR #5699, opened 2025-08-12
lib: simplify `buffered.read_fully`

michaellilltokiwa:


michaellilltokiwa:
> I don't think this is correct in case you want to read more than `i32.max` bytes.

I don't think this is currently possible. Since array can not be longer than `i32.max` anyway

michaellilltokiwa:
> True. We should document this. This basic API is not suitable for reading larger amounts of data anyway.

We should absolutely document this, but I'm unsure how. I think it would probably make sense to add an array.type.max_length and use this instead of i32.max since this is actually semantically what I want here.

PR #5747, opened 2025-08-15
lib: add percent encoding

simonvonhackewitz:
Percent encoding as specified in [RFC 3986 section 2.1](https://datatracker.ietf.org/doc/html/rfc3986#autoid-12)

[Wikipedia: Percent-encoding](https://en.wikipedia.org/wiki/Percent-encoding)


simonvonhackewitz:
Keeping terminology "encode" / "decode" instead of "escape" as discussed in person. Seems to be the term that most modern languages/libraries use.
Also continue using the name "percent encoding" over "URL encoding" for now, because RFC 3986 names it that way.

modules/base/src/encodings/percent.fz
```diff
@@ -0,0 +1,138 @@
...
+    decode enc_str false
+
+
+  # does the actual decoding
+  #
+  decode(enc_str String, tolerant bool) outcome (Sequence u8) =>
+    enc := enc_str.as_codepoints
+    len := enc.count
+
+    dec(res Sequence u8, i i32) outcome (Sequence u8) =>
```
michaellilltokiwa, line 119: My feeling is that instead of explicit recursion the code might be cleaner using a fuzion loop. Not sure though...

simonvonhackewitz: I thought about this as well, because I thought it might be easier to iterate over the codepoints instead of accessing them by index, but sometime three codepoints get consumed instead of one.
I think each branch of the `if` would have to return an intermediate result as well as the next index, and it would need an extra check for errors in the intermediate result. I didn't try it, but I think the code would not get any shorter.

PR #5757, opened 2025-08-19
docs: fix display of type constraint in api docs

### LINKED ISSUE #5751: API docs: constraint of type parameters seems to be broken
E.g. universe feature `infix <` is shown as
`infix < (T type : Type, a T, b T)`
but should be
`public infix <(T type : property.orderable, a, b T) bool =>`

simonvonhackewitz:
fix #5751


michaellilltokiwa:
@simonvonhackewitz I don't think we should skip ci in this case, since ci also builds the docs.

PR #5785, opened 2025-08-21
tests: sort effects

michaellilltokiwa:
so we do not have to update these recordings all the time...


michaellilltokiwa:
@maxteufel I thought about this but decided against it because we may intentionally emit effects in the order they are found. (This is not the case currently.) So I thought this solution is more "future proof".

PR #5791, opened 2025-08-21
run_tests: re-add slowest tests

michaellilltokiwa:
e.g:
```
testing JVM backend: 4 tests, running 1 tests in parallel.
....
4/4 tests passed, 0 skipped, 0 failed in 15s.

Slowest 5 tests using jvm backend:

5505ms abstractfeatures_negative
4311ms inherit_postcondition
3593ms inference_choice_generics
2253ms infer_via_contraints_type_parameter
```


michaellilltokiwa:
Does not work like this 9999ms is slower than e.g. 1s

PR #5819, opened 2025-08-28
docs: fix missing universe comment

simonvonhackewitz:


michaellilltokiwa:
We should probably not silently ignore the exception of `Files.readAllLines`

PR #5860, opened 2025-09-03
lib: change visbilities of handles(2)/bitset

michaellilltokiwa:


fridis (approved):
for a separate patch: replace the comment `unit bitset` by something else, I do not understand what this should mean (even though I probably wrote it originally...).

Issue #5903, opened 2025-09-08
java.lang.Error: check-condition failed: GeneratingFUIR.java:2743


simonvonhackewitz:
At 31f0ed626ca2ab3b4d832cf74f54822fdeff82a9 the following diff causes a failed check condition. Moving `DEL` into universe avoids it.

```diff
diff --git a/modules/base/src/container/Mutable_Hash_Map.fz b/modules/base/src/container/Mutable_Hash_Map.fz
index 559d978cd..a3399f16c 100644
--- a/modules/base/src/container/Mutable_Hash_Map.fz
+++ b/modules/base/src/container/Mutable_Hash_Map.fz
@@ -84,9 +84,15 @@ is
   #
   allocated_size := mut (min_size.as_f32 / upper_load_factor).ceil.as_i32
 
+  # element at this index was deleted, this index should be treated as occupied when looking up elements
+  # to not cause unexpected gaps that could lead to lost elements
+  # for insertions it is treated the same as an empty index
+  DEL is
+
   # the contents
   #
-  contents := mut (LM.env.new_array (option (tuple HK V)) allocated_size.get.as_i64 nil)
+  contents := mut (LM.env.new_array (choice nil DEL (tuple HK V)) allocated_size.get.as_i64 nil)
+
 
 
   # calculate the index of k within contents array in case of no conflict
@@ -127,16 +133,21 @@ is
   =>
     allocated_size <- new_alloc_size
     old_contents := contents.get.as_array
-    contents <- LM.env.new_array (option (tuple HK V)) allocated_size.get.as_i64 nil
-    for tup in (old_contents.filter o->o??) do
-      ok, ov := tup.get
-      store (idx ok) ok ov false
+    contents <- LM.env.new_array (choice nil DEL (tuple HK V)) allocated_size.get.as_i64 nil
+    for elem in old_contents do
+      match elem
+        t tuple HK V =>
+          ok, ov := t
+          store (idx ok) ok ov false
+        * => # noting to add
 
 
   # store key value pair at given index
   # updates value for existing key, does conflict resolution
   #
   store (at i64, k HK, v V, update_size bool) =>
+    is_update bool := has k
+
     match contents.get[at]
       nil     =>
         contents.get[at] := (k, v) # insert new key value pair
@@ -147,6 +158,12 @@ is
           contents.get[at] := (k, v)
         else                      # conflict
           store (collision at) k v update_size
+      DEL =>
+        if is_update
[...]

PR #5954, opened 2025-09-15
lib: improve performance of `Mutable_Hash_Map` when many succeeding hashes have conflicts

simonvonhackewitz:
requires #5903 to be fixed

### DIFF
--- a/modules/base/src/container/Mutable_Hash_Map.fz
+++ b/modules/base/src/container/Mutable_Hash_Map.fz
@@ -68,6 +68,11 @@ pre
 
 is
 
+  # element at this index was deleted, this index should be treated as occupied when looking up elements
+  # to not cause unexpected gaps that could lead to lost elements
+  # for insertions it is treated the same as an empty index
+  del is
+
   # number of entries currently in this map
   #
   load := mut 0
@@ -86,7 +91,7 @@ is
 
   # the contents
   #
-  contents := mut (LM.env.new_array (option (tuple HK V)) allocated_size.get.as_i64 nil)
+  contents := mut (LM.env.new_array (choice nil del (tuple HK V)) allocated_size.get.as_i64 nil)
 
 
   # calculate the index of k within contents array in case of no conflict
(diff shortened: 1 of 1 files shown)


modules/base/src/container/Mutable_Hash_Map.fz
```diff
@@ -101,7 +106,7 @@ is
   # return the next alternative position to check
   #
   collision (at i64) =>
-    (at + 1) % allocated_size.get.as_i64  # NYI: dumb collision function, check literature and improve!
+    (at + 904644503766293287) % allocated_size.get.as_i64
```
michaellilltokiwa, line 110: This magic number needs a comment and/or name.

modules/base/src/container/Mutable_Hash_Map.fz
```diff
@@ -127,26 +132,40 @@ is
...
       t tuple =>
         ek, _ := t
         if ek = k                 # no conflict, but remapping of k
           contents.get[at] := (k, v)
         else                      # conflict
-          store (collision at) k v update_size
+          store (collision at) k v update_size is_update
+      del =>
+        if is_update
+          store (collision at) k v update_size is_update
```
michaellilltokiwa, line 168: Why does a `put` of an element collide with a `del`?

michaellilltokiwa (commented):
It would be good to add a benchmark in /benchmarks covering the most important cases. But does not have to be done in this PR, I would be fine with an enhancement issue.

modules/base/src/container/Mutable_Hash_Map.fz
simonvonhackewitz: When elements get deleted, that index is marked with `del`, instead of the gap getting filled by moving the following elements. Therefore only `nil` marks the end of the linear probing, so when reaching a `del` it is not known if the same key might be in an later element and therefore an update instead of an insert.

michaellilltokiwa: Makes sense. Would make sense to add this as a comment, I think.

michaellilltokiwa (approved):

PR #5960, opened 2025-09-15
lib: add more (infix) set operators

### LINKED ISSUE #5927: implement more Set operators
∪ → union: all elements in either set.

∩ → intersection: elements common to both sets.

∆ / ⊖ → symmetric difference: elements in one set or the other, but not both.

∖ → set difference: elements in the first set but not in the second.

∈ → element of: shows membership

∉ → not element of: shows non-membership

⊆ → subset: every element of the first set is in the second (may be equal).

⊂ → proper subset: every element of the first set is in the second, but sets are not equal.

⊇ → superset: first set contains all elements of the second (may be equal).

⊃ → proper superset: first set strictly contains all elements of the second.

simonvonhackewitz:
fix #5927

### DIFF
--- a/modules/base/src/container/Set.fz
+++ b/modules/base/src/container/Set.fz
@@ -101,21 +101,72 @@ is
     Set.this.difference other
 
 
+  # all elements that are in either one of the sets, but not in both
+  #
+  public symmetric_difference(other container.Set E) container.Set E =>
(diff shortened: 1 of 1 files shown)


michaellilltokiwa (approved):
Nice!

Some tests would also be nice :-)

Issue #5963, opened 2025-09-16
DFA takes ~30s and >100 iterations for test mod_http_message


michaellilltokiwa:
```
101 features 111 types and 1 source files included in fum file.
DFA pre iteration #1: --------------------------------------------------calls:0,values:2,envs:0
DFA pre iteration #2: --------------------------------------------------calls:8397,values:7183,envs:0
DFA pre iteration #3: --------------------------------------------------calls:8575,values:7327,envs:0
DFA pre iteration #4: --------------------------------------------------calls:9065,values:7838,envs:0
DFA pre iteration #5: --------------------------------------------------calls:9225,values:7989,envs:0
DFA pre iteration #6: --------------------------------------------------calls:9241,values:7993,envs:0
DFA pre iteration #7: --------------------------------------------------calls:9317,values:8071,envs:0
DFA pre iteration #8: --------------------------------------------------calls:9340,values:8093,envs:0
DFA pre iteration #9: --------------------------------------------------calls:9347,values:8100,envs:0
DFA pre iteration #10: --------------------------------------------------calls:9388,values:8148,envs:0
DFA pre iteration #11: --------------------------------------------------calls:9402,values:8166,envs:0
DFA pre iteration #12: --------------------------------------------------calls:10266,values:9153,envs:0
DFA pre iteration #13: --------------------------------------------------calls:10817,values:9808,envs:0
DFA pre iteration #14: --------------------------------------------------calls:10925,values:9917,envs:0
DFA pre iteration #15: --------------------------------------------------calls:10992,values:9970,envs:0
DFA pre iteration #16: --------------------------------------------------calls:11013,values:9977,envs:0
DFA pre iteration #17: --------------------------------------------------calls:11037,values:9984,envs:0
DFA pre iteration #18: --------------------------------------------------calls:11045,values:9993,envs:0
DFA pre iteration #19: --------------------------------------------------calls:11046,values:9993,envs:0
DFA pre iteration #20: --------------------------------------------------calls:11051,values:9994,envs:0
DFA pre iteration #21: --------------------------------------------------calls:11051,values:9994,envs:0
DFA pre iteration #22: --------------------------------------------------calls:11243,values:10195,envs:0
DFA pre iteration #23: --------------------------------------------------calls:11274,values:10212,envs:0
DFA pre iteration #24: --------------------------------------------------calls:11287,values:10215,envs:0
DFA pre iteration #25: --------------------------------------------------calls:11295,values:10229,envs:0
[...]

PR #5964, opened 2025-09-16
fuir: cache lookup

michaellilltokiwa:
While investigating #5963 I found that fuir.lookup takes a lot of time.

Introducing a simple cache reduced the time for the test by about ~20%


fridis (changes requested):
A suggestion for a better `key` and avoiding multiple returns.

PR #5991, opened 2025-09-18
lib: rename `as_codepoints` -> `codepoints`

michaellilltokiwa:


simonvonhackewitz:
I often typed `.codepoints` first, so I'm in favour of this change.

But isn't it inconsistent with most other features that change or convert the type?

simonvonhackewitz (approved):

Issue #5892, opened 2025-09-08
lambdas for `Function` with type parameters


fridis:
Processing variadic arguments currently require quite some boilerplate code:
```
f(A type..., values A...) =>
  concat : container.typed_applicator String is
    public redef apply(T type, e String, v T) String
    =>
      e + $v
  values.typed_foldf "" concat
```
If we added variants of `Function` that have one or several type parameters, we could change this to
```
f(A type..., values A...) =>
  values.typed_foldf "" _,e,v->
      e + $v
```
or even
```
f(A type..., values A...) =>
  values.typed_foldf "" e,v->
      e + $v
```
The question is: Does this introduce a signficant compliation of the handling of `lambda`s? I think this might be quite simple to add.

PR #5994, opened 2025-09-18
ast: add support to infer partial and lambda from type parameter constraint

### LINKED ISSUE #5987: No type propagation from type constraint to lambda
Take this example
```
test(F type : i32->i32, f F) =>
  say "$F {f.call 42}"

test i32->i32 x->x+1
test i32->i32 (+1)

test x->x+1   # these should infer internal type of lambda for `F`
test (+1)
```
The latter two calls to `test` fail
```
 > ./build/bin/fz test_fun.fz

/home/fridi/fuzion/clean/fuzion.5/test_fun.fz:7:1: error 1: Failed to infer actual type parameters
test x->x+1   # these should infer internal type of lambda for `F`

In call to 'test', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'F'
Type inference failed for one type parameter 'F'


/home/fridi/fuzion/clean/fuzion.5/test_fun.fz:8:7: error 2: Incompatible types when passing argument in a call
test (+1)

Actual type for argument #1 'f' does not match expected type.
In call to          : 'test'
expected formal type: 'Unary i32 i32'
actual type found   : 'i32'
assignable to       : 'i32'
for value assigned  : '(+1)'
To solve this, you could change the type of the target 'f' to 'i32' or convert the type of the assigned value to 'Unary i32 i32'.

2 errors.
```
However, the constraint on `F` could solve this.
This would have the useful effect that feature `test` could get specialized for the actual (internal) type created for the lambda such that the calls require no boxing, the function value `f` would instead be a unit type and disappear.

fridis:
This fixes #5987 and is a step towards #5892, labdas for `Function` with type parameters.

Added a small test case.

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -1685,7 +1685,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
   AbstractType constraint(Context context)
   {
     if (PRECONDITIONS) require
-      (state().atLeast(State.RESOLVED_TYPES),
+      (state().atLeast(State.RESOLVED_DECLARATIONS),
        isTypeParameter());
 
     var result = context.constraintFor(this);
@@ -1710,7 +1710,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
   public AbstractType constraint()
   {
     if (PRECONDITIONS) require
-      (state().atLeast(State.RESOLVED_TYPES),
+      (state().atLeast(State.RESOLVED_DECLARATIONS),
        isTypeParameter());
 
     throw new Error("constraint() not redefined for "+getClass());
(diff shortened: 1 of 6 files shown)


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -1685,7 +1685,7 @@ public TypeKind defaultTypeKind()
   AbstractType constraint(Context context)
   {
     if (PRECONDITIONS) require
-      (state().atLeast(State.RESOLVED_TYPES),
```
michaellilltokiwa, line 1688: RESOLVING_TYPES is probably also okay? But does not matter much.

michaellilltokiwa (approved):

src/dev/flang/ast/AbstractFeature.java
fridis: I think this did not work. I am not completely happy since I do not fully understand why, but just went with this for now.

PR #6037, opened 2025-10-02
base: make `Sequence`, `list` and `array` `equatable`, `orderable` and `hashable`

fridis:
First, moved code for xxh-style hashes to `property.hashable` such that it can be used at several locations. Then added `equality`, `lteq` and `hashable` to Sequence and to `list`.

Remove `as_equatable` and `as_orderable` since these are no longer needed.

Updated tests.

This depends on #6036, so I make this a draft until that PR is merged.

### DIFF
--- a/bin/ebnf.fz
+++ b/bin/ebnf.fz
@@ -181,7 +181,7 @@ main =>
     tmp := !!"mktemp --directory"
     _ := !!"mkdir $tmp/fuzion_grammar"
     _ := io.file.write "$tmp/fuzion_grammar/Fuzion.g4" ebnf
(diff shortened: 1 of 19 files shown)


michaellilltokiwa:
Nice. Not sure if there are other programming languages that can do this?

Should we ensure that `finite.is_yes_or_unkown`?

michaellilltokiwa:
> Dynamic languages can do this, but checking it all statically is maybe something Fuzion is unique in :-).

Yes that is what is was thinking of. Failing at compile time.

michaellilltokiwa (approved):

PR #6081, opened 2025-10-14
Readme.md: add section on tests

michaellilltokiwa:
Use "Display the rich diff" to review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,8 @@ Have fun!
 
 ### Running all tests
 
+> Since there are a lot of tests this will likely take 1h or more. See below on how to run a single test.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -232,6 +236,67 @@ To compile the same example (requires clang C compiler):
...
+
+    # for jvm backend
+    make run_tests_jvm
+
+    # for c backend
+    make run_tests_c
+
+    # for interpreter backend
+    make run_tests_int
+
```
maxteufel: It should be documented that running all tests, or running all tests for a specific backend, will take quite a while nowadays and might be not what is desired (in many cases running specific tests is sufficient).

PR #6107, opened 2025-10-22
mod, web: extract port from url

michaellilltokiwa:


modules/web/src/web/shared.fz
```diff
@@ -53,27 +53,51 @@ module write_request(LM type : mutate, rm http.request_message) ! LM =>
...
+#
+# will result in
+#
+# (false, "tokiwa.software", 8080, "index")
+#
+# NYI: BUG: parse [host]:port format correctly bracketed ipv6
+# NYI: BUG: no port bounds check <= 65535 before casting
+# NYI: BUG: ignores userinfo: user:pass@host
+#
+module parse_url(url String) outcome (tuple bool String (option u16) (option String))
```
maxteufel: please reduce the tuple-itis and replace the tuple with a record type `parsed_url(is_http bool, host String, port option u16, target option String)`

PR #6110, opened 2025-10-23
mod crypto: wrapping libsodium

michaellilltokiwa:
currently two public features `crypto.password_hash` and `crypto.password_hash_verify`

### DIFF
--- a/tests/mod_crypto/Makefile
+++ b/tests/mod_crypto/Makefile
@@ -21,6 +21,8 @@
 #
 # -----------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


maxteufel (changes requested):
I propose putting the pure libsodium wrapper in a separate module `sodium` from the `crypto` module that handles user-facing code. In the future we might also want to call other crypto libraries from the `crypto` module. This will make this cleaner in advance.

maxteufel (approved):

PR #6115, opened 2025-10-23
fuir: remove `lookup_array_...` methods

michaellilltokiwa:
There is no need for those, we can use `clazzArg`


maxteufel:
The tests are still failing.

Issue #5335, opened 2025-06-05
We should make sure that one can not inherit from any parents of intrinsic features


michaellilltokiwa:
This would only lead to compiler errors, indicating missing intrinsics.

PR #6116, opened 2025-10-23
lib: change visiblity of atomic to `private:public`

michaellilltokiwa:
inheriting from atomic would be bad since it has intrinsics as child features.

related #5335


maxteufel (approved):
Looks good. Please submit a matching pull request to `fzweb` if necessary.

PR #6118, opened 2025-10-24
Makefile: remove min-java

michaellilltokiwa:
It is kind of unclear what min means here anyway...


michaellilltokiwa:
> I think the idea was to have something that could run all the examples and tests but not the rest. I used the `min-java` target heavily at some point because it is less time consuming than building `all` when I just needed the Java tests. So I don't know if I'm happy with this change.

Okay, makes sense. Closing this PR

PR #6121, opened 2025-10-24
lib: add String.as_string(max_codepoints)

michaellilltokiwa:
e.g.
```
fz -e 'say <| "hello world".as_string 3'
hel…
```

### DIFF
--- a/modules/base/src/String.fz
+++ b/modules/base/src/String.fz
@@ -42,6 +42,7 @@ public String ref : property.equatable, property.hashable, property.orderable is
       String.this
     else
       substring_codepoint 0 max_codepoints-1 + "…"
+
   # any concrete string must implement utf8
   public utf8 Sequence u8 => abstract
 


fridis (commented):
I see that this can often be useful, but I would rename it as something like `truncate` and implement it using `substrung_codepoint`.

modules/base/src/String.fz
```diff
@@ -30,6 +30,18 @@ public String ref : property.equatable, property.hashable, property.orderable is
...
+  #
+  public truncate(max_codepoints i32) String 
+    pre
+      debug: max_codepoints > 0
+  =>
+    if codepoint_length <= max_codepoints then
+      String.this
+    else
+      substring_codepoint 0 max_codepoints-1 + "…"
   # any concrete string must implement utf8
```
michaellilltokiwa:
```suggestion
      substring_codepoint 0 max_codepoints-1 + "…"

  # any concrete string must implement utf8
```

PR #6150, opened 2025-10-29
lib: cleanup argtype of baseX.encode

### LINKED ISSUE #6061: lib: base64, type for argument should be Sequence not array
```public encode(data array u8) Sequence u8 =>```
should probably be the other way round
```public encode(data Sequence u8) array u8 =>```

michaellilltokiwa:
fixes #6061

### DIFF
--- a/tests/base16/base16_test.fz
+++ b/tests/base16/base16_test.fz
@@ -66,7 +66,7 @@ base16_test is
     do
       plain, code_expected := tup
       code_actual := match plain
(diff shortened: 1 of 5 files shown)


maxteufel:
I'm a bit confused by this. I think at some point we intentionally switched `Sequence` to `array` to make sure we only have O(1) accesses inside features, now we're changing this back. Or am I misremembering this?

PR #6153, opened 2025-10-30
be/c: remove mem_zero for fzCur

michaellilltokiwa:
mem_zero was visibile in my profilings i did for #6152


maxteufel:
I find this a bit worrying to just remove this and suppress the warning it causes. @fridis mentioned in personal discussion that such a removal might be causing problems on non-Linux platforms.

Does this just appear often in profiling, or is there a lot of time spent for executing this.

PR #6166, opened 2025-10-30
remove open file mode

### LINKED ISSUE #6160: `io.file.write` appends data instead of failing or overwriting file

michaellilltokiwa:
fixes #6160


include/posix.c
```diff
@@ -726,7 +729,10 @@ void * fzE_file_open(char * file_name, int64_t * open_results, file_open_mode mo
   // make sure no fork is done while we open file
   fzE_lock();
 
-  FILE * fp = fopen(file_name, mode==FZ_FILE_MODE_READ ? "rb" : "a+b");
+  FILE *fp = fopen(file_name, "r+b");
+  if (!fp) {
+    fp = fopen(file_name, "w+b");
+  }
```
maxteufel: so now, writing is no longer possible at all unless they file doesn't exist in the first place?

michaellilltokiwa: Problem is, if we just use `w+b` then file is truncated. If we use `a+b` then any writes will always be at end of file.
This way we first try `r+b` which opens read/write. Then if file does not exist we try again with `w+b`, this then creates the file.

michaellilltokiwa: With this change truncation should never happen. See test.

PR #6167, opened 2025-10-30
lib: Sequence.statistics, return some stats as a string

michaellilltokiwa:


modules/base/src/Sequence/statistic.fz
```diff
@@ -110,3 +110,30 @@ public product T
   pre
     T : numeric
 => fold T.product
+
+
+# return some statistics for the Sequence
```
fridis: This comment should be much more specific!

PR #6168, opened 2025-10-30
remove option unsafeIntrinsics

### LINKED ISSUE #6165: `unsafeIntrinsics` is probably orphaned
File I/O is happening via native features now. This probably makes the option useless.

michaellilltokiwa:
does not work as intended anymore.
fixes #6165


maxteufel (approved):
Will need removal from `flang_dev` and other places too.

PR #6188, opened 2025-11-03
lib: add `String.from`, creating String from variable data

michaellilltokiwa:
current definition (but subject to further extension):
```
public type.from(data choice (Sequence u8) (Sequence codepoint)) String =>
```


fridis (approved):
nice!

PR #6191, opened 2025-11-03
Upgrade to Java 25

### LINKED ISSUE #5975: Upgrade to Java 25 (LTS version)
`grep -i java.*21` should find most place that need to be adjusted

simonvonhackewitz:
fix #5975

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -58,12 +58,12 @@ jobs:
       - name: install dependencies
         run: |
           brew install openjdk@25 bdw-gc gnu-sed make coreutils sqlite llvm wolfssl libsodium
-
-          # Homebrew does not auto-link openjdk@25
-          sudo ln -sfn "$(brew --prefix openjdk@25)/libexec/openjdk.jdk" /Library/Java/JavaVirtualMachines/openjdk-25.jdk
-          echo "JAVA_HOME=$(/usr/libexec/java_home -v 25)" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


src/dev/flang/be/jvm/classfile/ClassFileConstants.java
```diff
@@ -73,9 +73,13 @@ public interface ClassFileConstants
...
   public static byte[] VERSION_JDK_20  = new byte[] { 0, 0, 0, 64 };
   public static byte[] VERSION_JDK_21  = new byte[] { 0, 0, 0, 65 };   // LTS
+  public static byte[] VERSION_JDK_22  = new byte[] { 0, 0, 0, 66 };
+  public static byte[] VERSION_JDK_23  = new byte[] { 0, 0, 0, 67 };
+  public static byte[] VERSION_JDK_24  = new byte[] { 0, 0, 0, 68 };
+  public static byte[] VERSION_JDK_25  = new byte[] { 0, 0, 0, 69 };   // LTS
 
 
-  public static byte[] DEFAULT_VERSION = VERSION_JDK_21;
+  public static byte[] DEFAULT_VERSION = VERSION_JDK_25;
```
fridis: To change this, you have to check what has changed in the JVM spec for the class file and make sure that the class files we generate conform to that version. I suggest to either

- check that changes in the JVM spec and make sure we do nothing that could break, or
- leave this at 21 for now and add a NYI-comment.

fridis (changes requested):

PR #6203, opened 2025-11-04
Makefile: _simple_example, FUZION_BASE, order only dependency

michaellilltokiwa:
this should prevent frequent rebuilds of check_simple_example/record_simple_example.

see also: https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -516,13 +516,13 @@ $(FZJAVA): $(FZ_SRC)/bin/fzjava $(CLASS_FILES_TOOLS_FZJAVA)
 
 # this target may fail when executed standalone, see comment on $(BUILD_DIR)/tests
 #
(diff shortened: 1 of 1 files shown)


simonvonhackewitz:
Could the same be done with `MOD_TERMINAL`, allowing the target to be executed standalone and fix the following comment from target tests?

```
# tricky, we need MOD_TERMINAL to build (check|record)_simple_example
# but we are not adding this to those targets because we do not want
# (check|record)_simple_example to rebuild any time sth is changed in base lib
$(BUILD_DIR)/tests: $(FUZION_FILES_TESTS) $(MOD_TERMINAL) $(BUILD_DIR)/bin/check_simple_example $(BUILD_DIR)/bin/record_simple_example
```

maxteufel:
@simonvonhackewitz yes, i think that should work.

Issue #6176, opened 2025-10-31
Add feature `must_assign_to_field`, emit error if this is not done


michaellilltokiwa:
Might be helpful for features like `once` which are likely used wrongly when they are not immediately assigned to a field.

maxteufel:
how do you imagine this being used? as a precondition?

PR #6213, opened 2025-11-06
Add feature `must_assign_to_field`

maxteufel:
This causes an error to be emitted if a feature inheriting from it is not assigned to a field. We don't check for equality of instances and count assignments to check this, we rely on the fact that unused results are forbidden and constructors cannot have any non-unit result.


maxteufel:
The code is bad. Discussion can still happen in #6176.

PR #6214, opened 2025-11-06
nom: add xml parser

michaellilltokiwa:
Altough it is a bit of a stretch to call it an xml parser... Can only parse a subset of xml and also is not really error tolerant.

Goal is to use this in fzweb to fetch rss-feeds.

### DIFF
--- a/bin/run_tests.fz
+++ b/bin/run_tests.fz
@@ -231,7 +231,7 @@ main =>
         check concur.thread_pool.env.submit (run_single_test t)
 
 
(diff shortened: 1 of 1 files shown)


tests/mod_nom_parser_xml/mod_nom_parser_xml.fz
maxteufel: can you read the rss feeds from another file for readability of the tests? this would also work around the overlong strings.

maxteufel (approved):

PR #6241, opened 2025-11-12
post test results to influxdb

michaellilltokiwa:


Makefile
```diff
@@ -658,8 +658,8 @@ release: clean all
 # currently only code in bin/ and examples/ are checked.
 .PHONY: syntaxcheck
 syntaxcheck: min-java
-	find ./examples/ -name '*.fz' -print0 | xargs -0L1 $(FZ) -modules=terminal,clang,lock_free,java.base,java.datatransfer,java.xml,java.desktop -noBackend
-	find ./bin/ -name '*.fz' -print0 | xargs -0L1 $(FZ) -modules=terminal,clang,lock_free,java.base,java.datatransfer,java.xml,java.desktop -noBackend
+	find ./examples/ -name '*.fz' -print0 | xargs -0L1 $(FZ) -modules=terminal,clang,lock_free,java.base,java.datatransfer,java.xml,java.desktop,web,http,wolfssl -noBackend
```
maxteufel: `run_tests.fz` is only in `bin`, so `examples` shouldn't need those modules, should it?

maxteufel (changes requested):

Makefile
maxteufel: then they should probably be listed in some variable `TEST_MODULES` or similar.

PR #6244, opened 2025-11-13
base: add feature idx for index accesses

michaellilltokiwa:


maxteufel:
This looks good, but the reason I'm not approving is the bytecode size workaround. We should discuss how to handle that in general.

michaellilltokiwa:
I'm fine with not merging this for now but the bytecode size problem exits regardless of this PR. This just triggers it.

PR #6264, opened 2025-11-17
fuir: cleanup precondition

michaellilltokiwa:


maxteufel (commented):
I suppose DFA marks these special clazzes by now?

maxteufel (approved):

PR #6274, opened 2025-11-19
add jenkins fail mails

michaellilltokiwa:


maxteufel (commented):
`recipientProviders` should already include the committer with `developers()`. See <https://plugins.jenkins.io/email-ext/#plugin-content-pipeline-step>.

PR #6285, opened 2025-11-20
tools: add `-gdb`, easily start gdb for c code

michaellilltokiwa:


maxteufel (approved):
Nice! This will lower barriers to debugging. The only thing that slightly bothers me is the ever growing amount of options.

PR #6296, opened 2025-11-21
native features: switch from `array` to `mutate.array`

michaellilltokiwa:


maxteufel (approved):
Nice! This should avoid surprises with arrays being suddenly mutated.

PR #6297, opened 2025-11-24
fix typo

michaellilltokiwa:
[ci skip]

### DIFF
--- a/modules/base/src/net/ip_address.fz
+++ b/modules/base/src/net/ip_address.fz
@@ -21,7 +21,7 @@
 #
 # -----------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


modules/base/src/net/ip_address.fz
```diff
@@ -21,7 +21,7 @@
 #
 # -----------------------------------------------------------------------
 
-# ip_address - stores an IP address an an unsigned 128-bit integer,
+# ip_address - stores an IP address an unsigned 128-bit integer,
```
maxteufel:
```suggestion
# ip_address - stores an IP address in an unsigned 128-bit integer,
```

I think that's what this was supposed to be?

PR #6301, opened 2025-11-24
fix typos

michaellilltokiwa:
according to AI:

> The terms **"short-hand"** and **"shorthand"** are often confused, but they have distinct meanings:
> 
> ### **Shorthand**
> - **Definition:** A system of rapid writing using symbols or abbreviations for words and phrases, often used for taking dictation or notes quickly.
> - **Example:** Gregg shorthand, Pitman shorthand.
> - **Usage:** "She took the lecture notes in shorthand."
> 
> ### **Short-hand** (or **shorthand** as an adjective)
> - **Definition:** When used as an adjective (sometimes hyphenated as "short-hand"), it means something that is abbreviated, simplified, or serves as a quick reference.
> - **Example:** "The manual provided a short-hand guide to the software."
> - **Usage:** "He used a short-hand method to solve the problem."
> 
> ### **Key Difference**
> - **Shorthand** refers specifically to the writing system.
> - **Short-hand** (adjective) describes something that is concise or abbreviated.
> 
> In most cases, when referring to the writing system, **"shorthand"** (one word) is correct. When used as an adjective, **"short-hand"** (hyphenated) is acceptable but less common—**"shorthand"** is often used for both noun and adjective forms in modern usage.
>

### DIFF
--- a/modules/base/src/concur/atomic_access.fz
+++ b/modules/base/src/concur/atomic_access.fz
@@ -34,7 +34,7 @@ private:module atomic_access : effect is
     concur.atomic_access.default  /* NYI: `concur.` could be added automatically by front end */
 
 
-# short-hand for accessing mut effect in current environment
+# shorthand for accessing mut effect in current environment
 #
 module atomic_access_env =>
   atomic_access.install_default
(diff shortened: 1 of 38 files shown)


maxteufel:
According to https://en.wiktionary.org/wiki/shorthand, "short-hand" is archaic (even "less common" according to your text), so I don't see why we should change this.

michaellilltokiwa:
@maxteufel I now replaced all _short-hand_ by _shorthand_

Issue #5026, opened 2025-04-08
test `native_value` fails in github action with exit code 134


michaellilltokiwa:
In my ubuntu test VM, it works.

The exit code seems to indicate that something awful happens...

michaellilltokiwa:
seems fixed? #6312

PR #6312, opened 2025-11-26
try enable test native_value again

michaellilltokiwa:
since I can not reproduce the error on my machine, tried this:

    seq 100 | parallel -j 12 ../../build/bin/fz -jvm -JLibraries=clang native_value.fz


michaellilltokiwa:
fixes #5026

PR #6314, opened 2025-11-26
net: implement basic IP address parsing

maxteufel:


michaellilltokiwa (approved):
Good, I was a little confused by the matching comments while the from_string_v* variants only parse ips that do not require padding, right?

maxteufel:
Added some more comments for clarification. The `from_string_v4` variant does not require any padding, since we just do `parse_i32` on each block separated by `.`. In IPv6, two types of padding is necessary, once to remove ellipsis `::`, and another time to left-pad the `:` separated blocks (they are later split into blocks of length 2 hex digits).

PR #6323, opened 2025-11-27
Metrics  add some dfa metrics

michaellilltokiwa:


src/dev/flang/util/Influx.java
```diff
@@ -0,0 +1,75 @@
...
+    try
+      {
+        HttpResponse<String> response = client.send(
+          request, HttpResponse.BodyHandlers.ofString());
+        return response.body();
+      }
+    catch (Exception e)
+      {
+        return "Error sending to InfluxDB: " + e.getMessage();
+      }
```
maxteufel: it might be a sensible idea to (gracefully) log this somewhere (but not the other messages generated from the `postToInflux` method)

PR #6344, opened 2025-12-01
implement `from_u32` according to @fridis review suggestion

michaellilltokiwa:


modules/base/src/numeric.fz
```diff
@@ -205,7 +205,12 @@ public numeric : property.hashable, property.orderable is
   # maximum in case of overflow
   #
   public type.from_u32(v u32) numeric.this =>
-    if v = 0 zero else (from_u32 v-1) +^ one
+    if      v = 0 then zero
+    else if v = 1 then one
+    else
+      h := from_u32 v/2
+      l := from_u32 v%2
+      h+h+l
```
maxteufel: we previously used saturating addition, is there any reason we don't anymore?

michaellilltokiwa: I didn't notice this but I think we probably want a precond failure in cases like `u8.from_u32 9876`

michaellilltokiwa: But the comments needs to be changed then...

PR #6361, opened 2025-12-02
disable lsp test again

michaellilltokiwa:
[ci skip]


tests/lsp/skip
```diff
@@ -0,0 +1 @@
+NYI: BUG: LSP process does seem to be deterministic
```
maxteufel:
```suggestion
NYI: BUG: LSP process does not seem to be deterministic
```
?

this does not make sense to me otherwise

PR #6404, opened 2025-12-15
Makefile: rebuild only necessary java files

michaellilltokiwa:


michaellilltokiwa:
> I don't think we should change the build process in such a way that errors that could be detected at compile time will only be detected at runtime.
> 

Just to reiterate, this would only happen on rebuilds, not when doing a full clean build.
But anyway, I will close this PR then.

PR #6412, opened 2025-12-16
Resource arenas

michaellilltokiwa:
@tokiwa-software/developers looking for feedback on this.

Motivation: Currently there is no way to return a String backed by mapped_buffer without copying the buffer, (unless we are still inside the callback of the instated mmap-effect).

This PR adds a resource _arena_ of some sorts that handles the closing of the open resources. This means file.open/use now returns a file, no callback anymore. This file can then be used whenever as long as inside of the resources arena.

The user still does not have to, and currently can not, manually close the file. It is manged by the resources effect. Also the user still does not need to handle `File_Descriptor` in any way.


modules/base/src/resources.fz
```diff
@@ -0,0 +1,67 @@
...
+# when deinstated it closes the resources in reverse order
+# that they were opened.
+#
+public resources : effect is
+
+  # NYI:
+  _ := mut
+
+  # do not remove this type!
+  # it enforces boxing. Does not work with it.
```
maxteufel:
```suggestion
  # it enforces boxing. Does not work without it.
```

maxteufel (commented):
looks nice! how does this fare with multi-threaded environments?

michaellilltokiwa:
superseded by: #7486

PR #6433, opened 2025-12-18
lib: add `data_size` and `byte_unit`

simonvonhackewitz:
Example
```
❯ fz -e 'x := units.data_size.tb 1; say "A {x.as_string_si} HDD is only $x."' 
A 1.00 TB HDD is only 931 GiB.
```

Needs some testing, especially edge cases. ~~Everything above ≈15 EiB (or ≈18 EB) currently causes errors due to the maximum value of u64.~~
Maximum representable value with `u128` is 340,282,366 QB or 268,435,455 QiB.

@michaellilltokiwa I'd like some feedback on the overall approach and how it is structured.

### DIFF
--- a/modules/base/src/units/data_size.fz
+++ b/modules/base/src/units/data_size.fz
@@ -39,164 +39,164 @@ is
 
   # this units.data_size in kibibytes (1024**1 bytes), omitting fractional part
   #
-  public kib u128 => bytes / byte_unit_iec.kib.in_bytes
+  public kib u128 => bytes / byte_iec.kib.in_bytes
 
   # this units.data_size in kibibytes (1024**1 bytes), including fractional part
   #
-  public kib_float f64 => bytes.as_f64 / byte_unit_iec.kib.in_bytes.as_f64
+  public kib_float f64 => bytes.as_f64 / byte_iec.kib.in_bytes.as_f64
 
   # this units.data_size in mebibytes (1024**2 bytes), omitting fractional part
   #
-  public mib u128 => bytes / byte_unit_iec.mib.in_bytes
+  public mib u128 => bytes / byte_iec.mib.in_bytes
 
   # this units.data_size in mebibytes (1024**2 bytes), including fractional part
   #
-  public mib_float f64 => bytes.as_f64 / byte_unit_iec.mib.in_bytes.as_f64
+  public mib_float f64 => bytes.as_f64 / byte_iec.mib.in_bytes.as_f64
 
   # this units.data_size in gibibytes (1024**3 bytes), omitting fractional part
   #
-  public gib u128 => bytes / byte_unit_iec.gib.in_bytes
+  public gib u128 => bytes / byte_iec.gib.in_bytes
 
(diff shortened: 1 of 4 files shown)


michaellilltokiwa:
I think it might be easier to have just one type for data_size, private constructor and two inititing features, if necessary:
`data_size.type.from_si arg`,  `data_size.type.from_iec arg`

Does that make sense?

simonvonhackewitz:
Is there a convention for how many decimal places should be displayed? And when should the next larger unit be used?
Also, some control on when to add leading spaces for alignment would be good. They are useful when printing a table, but it looks strange in a single sentence.

michaellilltokiwa:
@simonvonhackewitz I would do it similar to:
```
  # create a string representation of this duration using the given unit.
  #
  public as_string(u time.time_unit) String =>
    n := $(nanos / u.in_ns)
    " "*(max 0 4-n.byte_length) + n + u.short_name
```

modules/base/src/units.fz
```diff
@@ -0,0 +1,26 @@
...
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature data_size
```
michaellilltokiwa:
```suggestion
#  Source code of Fuzion standard library feature units
```

modules/base/src/units.fz
```diff
@@ -0,0 +1,26 @@
...
+#
+# -----------------------------------------------------------------------
+
+# NYI comment
+#
```
michaellilltokiwa:
```suggestion
# units -- unit feature to group unit related features like data_size
#
```

michaellilltokiwa (changes requested):
Good but please add a comment what iec/si means. Also either add a test or an issue that tests for this are still missing.

michaellilltokiwa:
@simonvonhackewitz Hmm yes.
A good test is almost like a piece of art. In this particular case I see two goals of a test:
- test edge cases (e.g. u128.zero, u128.max) to gain confidence in the algorithm of e.g. as_string
- making sure code stays working as intended. The intention is either encoded in the test or the recording.

The test you added is good but I would probably focus more on edge cases than generate some more or less _random_ numbers.

modules/base/src/units/data_size.fz
```diff
@@ -0,0 +1,637 @@
...
+  public fixed infix - (other data_size) data_size
+  pre
+    safety: data_size.this >= other
+  =>
+    data_size (bytes - other.bytes)
+
+
+  # this units.data_size multiplied by factor n
+  #
+  public infix * (n u128) data_size
```
michaellilltokiwa: I missed this in my last review.
What is the rationale of `infix *` and `infix /`? 
I don't see that these make much sense?

michaellilltokiwa (approved):
Great!

PR #6437, opened 2025-12-19
ast: block/loop, remove `hasImplicitResult`

michaellilltokiwa:
I think this is a sharp edge without much benefit, so proposing to remove this.


maxteufel (changes requested):
as per discussion, seems to be incomplete (`lastIndexVarAsImplicitResult`, `booleanAsImplicitResult`)

PR #6458, opened 2026-01-08
fix off-by-one error in `String.split`

### LINKED ISSUE #6457: `String.split` drops last partial string
Try this
```
 > $FUZION/bin/fz -e 'say "a b c".split'
[a, b]

```
This is the root cause for [fuzion_aoc 2025/6](https://github.com/fridis/fuzion_aoc/issues/56) to fail.

simonvonhackewitz:
fix #6457


michaellilltokiwa:
That was fast :-)

michaellilltokiwa (approved):

PR #6460, opened 2026-01-09
only apply special handling for `else if` when both are in the same line

### LINKED ISSUE #6428: unjustified indentation error after nested `if`
Somehow, it does not reset the indentation correctly after the inner `if`. Does not matter whether this is in one line or not. 

```
ex is

  if false
  else
    if false then panic "should not happen"
    say "indentation is correct"
```

```
❯ FUZION_DISABLE_ANSI_ESCAPES=true fz /tmp/a.fz                                                                                                                               

/tmp/a.fz:6:5: error 1: Inconsistent indentation
    say "indentation is correct"
----^
Indentation reference point is /tmp/a.fz:3:3:
  if false
--^
While parsing: exprs, parse stack: exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), unit

one error.
```

This error happens in [AoC 2023/18](https://jenkins.tokiwa.software/job/flang.dev/job/main/814/testReport/aoc.2023/18/compile_skipped_examples___part2_fz__no_backend_/).

simonvonhackewitz:
fix #6428

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -2962,7 +2962,7 @@ elseBlock   : "else" block
     if (skip(true, Token.t_else))
       {
         // only use special handling for `else if` when they are in the same line
-        // otherwise it is an normal else block that might contain an `if`
+        // otherwise it is a normal else block that might contain an `if`
         if (current() == Token.t_if && elseLine == tokenSourcePos().line())
           {
             result = new Block(false, new List<Expr>(ifexpr(true)));


fridis (commented):
Good. 

There is one more case that I would like to see checked, though: What about this code
```
  if false
  else
  if true then say "ok"
```
and 
```
  if true
  else
  if true then say "ok"
```
I think, this should print `ok` in both cases, i..e, treat the second `if` as a completely new statements outside of the first `if`
Can you add these two cases to the regression test and check if they behave like that?

PR #6470, opened 2026-01-12
lib: change `time.duration.as_string` to not add padding by default

simonvonhackewitz:


modules/base/src/time/duration.fz
```diff
@@ -139,14 +139,44 @@ is
...
   # create a string representation of this duration using the given unit.
   #
   public as_string(u time.time_unit) String =>
+    as_string u false
+
+  # create a string representation of this duration using the given unit.
+  #
+  as_string(u time.time_unit, pad bool) String =>
+    n := $(nanos / u.in_ns)
+    "{pad ? n.as_string.pad_left 4 : n}{u.short_name}"
```
michaellilltokiwa:
```suggestion
    n := $(nanos / u.in_ns)
    "{pad ? n.pad_left 4 : n}{u.short_name}"
```

PR #6474, opened 2026-01-13
ast: for outerref set sourceposition to not available

michaellilltokiwa:


maxteufel (approved):
does this affect (error) output anywhere?

PR #6481, opened 2026-01-15
base: add `i128.as_f64`

simonvonhackewitz:


modules/base/src/i128.fz
```diff
@@ -241,6 +241,11 @@ public i128(public hi i64, public lo u64) : num.wrap_around, has_interval is
 
   public cast_to_u128 u128 => u128 hi.cast_to_u64 lo
 
+  # conversion to float
+  public fixed as_f64 f64 =>
+    abs_val := abs_to_u128 i128.this
+    sign.as_f64 * (abs_val.hi.as_f64 * (2.as_f64**64.as_f64) + abs_val.lo.as_f64)
```
michaellilltokiwa:
```suggestion
    sign.as_f64 * (abs_val.hi.as_f64 * (2.0**64.0) + abs_val.lo.as_f64)
```

michaellilltokiwa (commented):
Good, but I'm only 80% confident that there is no off by one error in there. Can you add an example that would catch it. I.e. an i128 that is larger than u64.max and do an as_string for this and for the converted number.

maxteufel (approved):

PR #6491, opened 2026-01-16
fix ArrayIndexOutOfBoundsException

michaellilltokiwa:


michaellilltokiwa:
Take a look at where this is called
```
        SourcePosition[] matched = new SourcePosition[cgs.size()];
        boolean ok = true;
        for (var c: cases())
          {
            ok &= ((Case) c).resolveType(res, cgs, context, matched);
          }
        var missingMatches = new List<AbstractType>();
```
This is an if/else so there are no _cases_  we have no real SourcePos, so we set SourcePosition.notAvailable.

PR #6497, opened 2026-01-20
base: `buffered_writer`, simplify, remove use of local mutate

michaellilltokiwa:


maxteufel (approved):
nice! much more readable

PR #6544, opened 2026-01-28
`String.join` use `fold` instead of `foldr`

michaellilltokiwa:
fixes the OOM when visiting fzweb/docs/base/index.html I don't fully understand why foldr leads to an explosion of `tail` evaluations. I set the heap in fzweb to `-Xmx100m` when testing this.


maxteufel (approved):
nice catch!

PR #6569, opened 2026-02-03
ast/Loop: set a source position for empty success blocks

maxteufel:
This improves error output in case of a "Block must end with a result expression" error.


maxteufel:
A reg test to test what exactly? I don't see a regression that happened here?

PR #6572, opened 2026-02-03
tests: add test, testing dfa calc lteq eval feature

michaellilltokiwa:


maxteufel (approved):
nice! one admittedly very pedantic nitpick about the commit message: it says "test" three times, maybe one time would be enough to know what it's about.

PR #6599, opened 2026-02-09
base: add more path operations

simonvonhackewitz:


michaellilltokiwa (changes requested):
Nice. I don't know where drop_first would be useful though...

PR #6606, opened 2026-02-10
parser: allow type tails for this-types

michaellilltokiwa:
It was not previously possible to specify a this-type.inner such as:
```
public type.empty mutate.this.array T =>
```
The work around that was used was to pass the type via type param again.
```
public type.empty (LM type : mutate) LM.array T =>
```

The Parser was extended to allow these this-types now. Which changes the code e.g. to create an empty mutable array 
with `lm:mutate is`
from:
```
lm.array u8 .empty lm
```
to
```
lm.array u8 .empty
```


fridis (changes requested):
Very nice. 

In the Fuzion code, there are still some `.type.new` that I think could be replaced by just  `.new`. 

Please add an example to the PR's description that explains this nicely such that we can include it in the news.

PR #6607, opened 2026-02-10
LibraryOut: let jdk pick secure random algo

michaellilltokiwa:


michaellilltokiwa:
> I very much prefer to chose a good algorithm here explicitly and not pick whatever the underlying JDK has chosen. I probably put the NYI comment here because I did not make an effort to understand what a good algorithm is. This is critical since a collision in the version numbers could break the security!

Hmm, I understand. This is what docs for SecureRandom say:

> 
> This class provides a cryptographically strong random number generator (RNG).
> 
> A cryptographically strong random number minimally complies with the statistical random number generator tests specified in[ FIPS 140-2, Security Requirements for Cryptographic Modules](http://csrc.nist.gov/cryptval/140-2.htm), section 4.9.1. Additionally, SecureRandom must produce non-deterministic output. Therefore any seed material passed to a SecureRandom object must be unpredictable, and all SecureRandom output sequences must be cryptographically strong, as described in RFC 1750: Randomness Recommendations for Security.

Personally I would have more faith in security in letting the JVM decide which algorithm is currently the _best_ than in our research. Also keep in mind we would need to update this if the algorithm is broken one day.

PR #6630, opened 2026-02-12
http: change version parsing code

michaellilltokiwa:
The main reason for changing this code is that
it used index access on a list.

### DIFF
--- a/modules/http/src/http.fz
+++ b/modules/http/src/http.fz
@@ -51,9 +51,11 @@ module crlf => "\r\n"
 # to that protocol within elements that require a minor version identifier.
 # https://datatracker.ietf.org/doc/html/rfc9110#section-2.5-6
 #
-module parse_http_version(ver String, err error->error) =>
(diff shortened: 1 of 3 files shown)


modules/http/src/http.fz
```diff
@@ -44,3 +44,16 @@ module as_header_map(kvs Sequence (String,String)) =>
...
 #
 module crlf => "\r\n"
+
+
+# When a major version of HTTP does not define any minor versions,
+# the minor version "0" is implied. The "0" is used when referring
+# to that protocol within elements that require a minor version identifier.
+# https://datatracker.ietf.org/doc/html/rfc9110#section-2.5-6
+#
+module parse_http_version(ver String, err error->error) =>
```
simonvonhackewitz: Why not pass the full version string (including the protocol)? Then it could check the protocol as well, and the error would not need to be passed as an argument.

simonvonhackewitz (approved):

PR #6638, opened 2026-02-16
base: `Finger_Tree`, sligth performance improvements

michaellilltokiwa:


modules/base/src/container/Finger_Tree.fz
```diff
@@ -125,11 +125,24 @@ single(T type, data Sequence T) : container.Finger_Tree T is
       r.prepend data
   redef modus choice (container.empty T) (container.single T) (container.deep T) => container.single data
   redef fill_internal_array(ia fuzion.sys.internal_array T, idx i32) unit =>
-    for i in 0..data.count-1 do
+    for i:=0, i+1
+    while i<data.count
```
maxteufel: is this much slower than `i in 0..data.count-1`? if so, this is probably a performance bug elsewhere, and this only a workaround?

maxteufel (approved):

PR #6652, opened 2026-02-18
base: make `idx` a template feature

### LINKED ISSUE #6423: Make `idx` a template feature
I really like the idea behind #6244 making it possible to use any small integer type to index an array.

An alternative and maybe more powerful and more efficient way to do this might be to make `idx` a template feature with, e.g., `as_u64 u64 => abstract` and have the indexed types inherit from `idx`. Then, the array feature could be declared like this:
```
  public redef index [ ] (i I: idx) T
  =>
    internal_array[i.as_i32]
``` 
The pros are that even user defined types could be used as index and the overhead of the `match` would disappear since all features using a type parameter with constraint `idx` will get specialized for the actual type. 

The cons are that this might cause confusion and there might be cases where it does not work nicely, we need to try.

michaellilltokiwa:
fixes #6423

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ MOD_MAIL              = $(BUILD_DIR)/modules/mail.fum
 MOD_WEB               = $(BUILD_DIR)/modules/web.fum
 MOD_SODIUM            = $(BUILD_DIR)/modules/sodium.fum
 MOD_CRYPTO            = $(BUILD_DIR)/modules/crypto.fum
+MOD_TOKIWA            = $(BUILD_DIR)/modules/tokiwa.fum
 MOD_WEBSERVER         = $(BUILD_DIR)/modules/webserver.fum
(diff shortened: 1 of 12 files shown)


modules/base/src/time/histogram.fz
```diff
@@ -292,8 +292,8 @@ is
     total_number <- total_number + 1
     if total_number.get > ignore_first
       if ignore_last > 0
-        ld := ignore_last_buffer[ignore_last_buffer_index]
-        ignore_last_buffer[ignore_last_buffer_index] := d
+        ld := ignore_last_buffer[ignore_last_buffer_index.get]
+        ignore_last_buffer[ignore_last_buffer_index.get] := d
```
maxteufel: what is the reason for this change?

maxteufel (approved):

PR #6663, opened 2026-02-18
base: add `io.copy`

michaellilltokiwa:


michaellilltokiwa:
@maxteufel This of course needs to be specified in the specific case.
Easy for e.g. copy file/array/mmaped buffer to network stream/pipe .
Into the other direction I imagine sth like reading the network stream / pipe to exhaustion.
Anyway these are just examples. Not yet sure how much sense they make.

PR #6665, opened 2026-02-18
base: add `Function.of`, lambda helper

michaellilltokiwa:
to declare a lambda that converts i32 to u64:
```
Unary u64 i32 .of (.as_u64)
```

### DIFF
--- a/modules/base/src/Function.fz
+++ b/modules/base/src/Function.fz
@@ -38,7 +38,7 @@ public Function(public R type,
     abstract
 
 
-  # convenience feature to declare a lambda
+  # convenience feature for declaring a lambda
   #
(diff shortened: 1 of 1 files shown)


modules/base/src/Function.fz
```diff
@@ -38,6 +38,15 @@ public Function(public R type,
     abstract
 
 
+  # convenience feature to declare a lambda
```
michaellilltokiwa:
```suggestion
  # convenience feature for declaring a lambda
```

maxteufel:
Can you give an example where this would be used over just using `(.as_u64)`? I fail to see how it would be helpful?

PR #6687, opened 2026-02-20
base: path, revert separator changes

michaellilltokiwa:
they weren't thought through carefully enough and don't work as intended

### DIFF
--- a/tests/lib_fileio_mmap/mmap_test.fz.effect
+++ b/tests/lib_fileio_mmap/mmap_test.fz.effect
@@ -1,4 +1,3 @@
-envir.vars
 fuzion.runtime.check_fault
 fuzion.runtime.contract_fault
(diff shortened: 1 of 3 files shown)


maxteufel (approved):

modules/base/src/path.fz
```diff
@@ -184,7 +184,7 @@ is
   # e.g. "/folder/file" or "relative_folder/file"
   #
   public redef as_string String =>
-    "{if is_absolute then path.separator else ""}{if names.is_empty && !is_absolute then "." else String.join names path.separator}"
+    "{if is_absolute then "/" else ""}{if names.is_empty && !is_absolute then "./" else String.join names "/"}"
```
simonvonhackewitz: Why did you add a special case for `./`?

PR #6692, opened 2026-02-23
Revert "base/error: do not allow line breaks at the end of the message"

maxteufel:
This reverts commit 5062ae11d848ee68d738e2ea48765403a62f2aba.


michaellilltokiwa:
@fridis, @maxteufel  What about this approach instead?:
```
pedantic : !msg.ends_with "\n"
```
or
```
pedantic : !msg.contains "\n"
```

Issue #6709, opened 2026-02-24
Forbid operator before call with arguments?


fridis:
This is IMHO a little confusing:
```
  a,b,c := (5,8,17)
  a(i i32) => i*10
  x := -a +b   # -(a +b)  = 80
  y := (-a) +c # (-a) + c = 12
  say x
  say y
```
I might make sense to forbid a prefix or infix operator in front of a call with space separated arguments (`$f a`) and require parentheses instead (`$(f a)` or `$f(a)`). 

Also, an infix operator that is separated by white space at only one side as in `(a) +b` or `a+ b` might`` as well be forbidden since `a +b` is interpreted completely differently. 

I came across this during review of #6708. This is related to #6238, #5252 and #4556.

PR #6714, opened 2026-02-25
parser: Disallow unbalanced infix operators such as `3 +x` or `x- 1`

fridis:
An unbalanced infix operator is one that has white space either before or after the operator and no white space on the other side. This can be confusing, so this patch forbids this case and produces a corresponding syntax error. In particular, the following is forbidden now:
```
  c := (a) +b
  d := a+ b  
  e := (a)+ b
  f := (a) +b
  g := a+ (b)
```
instead, we need white space either on both side as in 
```
  c := (a) + b
  d := a + b  
  e := (a) + b
  f := (a) + b
  g := a + (b)
```
or no white space on either side
```
  c := (a)+b
  d := a+b  
  e := (a)+b
  f := (a)+b
  g := a+(b)
```
One exception is the `infix :` operator that may omit the white space on the left side to permit code like
```
  pre
    debug: i >= 3
```
Still allowed are prefix operators in call argument such as
```
  h := f +b
```
or applied to a call
```
  i := +f b
```

This solves the second part related to infix operators described in #6709.

### DIFF
--- a/tests/visibility/visibility.fz
+++ b/tests/visibility/visibility.fz
@@ -257,7 +257,7 @@ visibility is
     x0 := 2;         chck $x0="2"                                      "x := 2"
     x1 := x0.as_f64; chck ($x1="2.0"  || $x1="2")                      "x := x.as_f64"
     x2 := x1 * 3.14; chck ($x2="6.28" || $x1="6.28000000000000024869") "x := x * 3.14"
-    x3 := x2.as_i32; chck $x3="6"                                     "x := x.as_i32"
+    x3 := x2.as_i32; chck $x3="6"                                      "x := x.as_i32"
   visi6
 
   visi7 =>


maxteufel (approved):

src/dev/flang/parser/OpExpr.java
```diff
@@ -203,6 +204,19 @@ else if (// 'a + * b' => 'a + (* b)' and
           { // infix op:
             Expr e1 = expr(max-1);
             Expr e2 = expr(max+1);
+            if (!op._whiteSpaceBefore &&  op._whiteSpaceAfter && !op._text.equals(":") ||  // we still allow code like `pre debug: i >= 3`.
```
michaellilltokiwa: As discussed, the special case `:` should probably go. We should create an issue for this.

michaellilltokiwa (approved):

PR #6763, opened 2026-03-09
parser: fix `isFeaturePrefix`

### LINKED ISSUE #6762: Syntax error when parsing this-constraint in contract
```
~/openvscode-server-fuzion/vscode-fuzion/fuzion (issue/5592)130$ cat ~/playground/test.fz
f is
  h(U type)
  pre U : f.this
  =>
    say U
~/openvscode-server-fuzion/vscode-fuzion/fuzion (issue/5592)$ m6 no-java && fz ~/playground/test.fz
make: Nothing to be done for 'no-java'.

/home/sam/playground/test.fz:3:17: error 1: No line break may occur at this position
  pre U : f.this⏎

This code is part of an expression that must reside within a single line.
While parsing: term, parse stack: term, opExpr, operatorExpr, expr, skipInheritanceCall, skipInheritanceCallList, skipInherits, skipFeaturePrefix, isFeaturePrefix, expr, exprs, block (twice), contract, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), unit
To solve this, enclose the expression in parentheses '(' and ')'.

one error.
```

michaellilltokiwa:
fixes https://github.com/tokiwa-software/fuzion/issues/6762, when using this-types in type-constraint


fridis (commented):
not sure about the skipInheritanceCallList change. Can you explain?

michaellilltokiwa:
Problem was that e.g. `U : f.this` was checked for `isFeaturePrefix`, this then lead to an (unrecoverable) syntax error even though done in a fork of the parser. Now we parse `opExpr()` instead of `call0` in `skipInheritanceCall` which does not emit a syntax error. But this needed a change that `=>` must not be considered an operator in several places.

fridis (approved):
nice. 

Total number of errors reported in the negative tests stays the same, good that `choice_negative` reports one fewer errors, but  then there is one more in `reg_issue1945_ambiguity_for_lambda_result_neg`.

Issue #5712, opened 2025-08-13
implicit `outer.this` results in crash during monomorphization


fridis:
I ran into this while working on #5681, but decided not to fix it there but create this issue. Take the following example
```
o is
  a is
    ignore (ref : i is)
  i is
b : o.a is
ignore b
```
This example creates an instance `ref : i` that inherits from `o.this.i`. This results in a crash:
```
 > PRECONDITIONS=true POSTCONDITIONS=true ./build/bin/fz test_outer_this.fz

error 1: java.lang.UnsupportedOperationException: Unimplemented method 'generics'
	at dev.flang.fe.ThisType.generics(ThisType.java:55)
	at dev.flang.ast.AbstractType.dependsOnGenericsNoOuter(AbstractType.java:843)
	at dev.flang.fuir.Clazz.<init>(Clazz.java:308)
	at dev.flang.fuir.GeneratingFUIR.newClazz(GeneratingFUIR.java:346)
	at dev.flang.fuir.GeneratingFUIR.newClazz(GeneratingFUIR.java:280)
	at dev.flang.fuir.GeneratingFUIR.newClazz(GeneratingFUIR.java:280)
	at dev.flang.fuir.Clazz.directParents(Clazz.java:546)
	at dev.flang.fuir.Clazz.parents(Clazz.java:567)
[...]
```
Here is a similar example 
```
o is
  a is
    ignore j
  i is
  j : i is

b : o.a is
ignore b
```
that results in a `check`-condition failure
```
 > PRECONDITIONS=true POSTCONDITIONS=true ./build/bin/fz test_outer_this_b.fz

error 1: java.lang.Error: check-condition failed: Clazz.java:1865 "(Errors.any() || i == of);"
	at dev.flang.util.ANY.check(ANY.java:440)
	at dev.flang.fuir.Clazz.findOuter(Clazz.java:1865)
	at dev.flang.fuir.Clazz.replaceThisType(Clazz.java:626)
	at dev.flang.fuir.Clazz.lambda$replaceThisType$0(Clazz.java:628)
	at dev.flang.ast.AbstractType.applyToGenericsAndOuter(AbstractType.java:2242)
	at dev.flang.fuir.Clazz.replaceThisType(Clazz.java:628)
	at dev.flang.fuir.Clazz.lambda$actualGenerics$1(Clazz.java:695)
	at dev.flang.util.List.map(List.java:558)
[...]

PR #6786, opened 2026-03-17
fuir: fix crash in `directParents` due to this-type not replaced

michaellilltokiwa:
This does not fix all examples in #5712, but it is a start.


src/dev/flang/fuir/Clazz.java
```diff
@@ -551,7 +551,8 @@ private Set<Clazz> directParents()
         var pt = p.type();
         var t1 = isRef() && !pt.isVoid() ? pt.asRef() : pt.asValue();
         var t2 = _type.actualType(t1);
-        var pc = _fuir.newClazz(t2);
+        var t3 = replaceThisType(t2, new List<>() /* NYI: correct? */);
```
maxteufel: …but i also worry a bit about the empty inheritance list here.

michaellilltokiwa: Yes, but there are other places where exactly this is also not correct yet.

PR #6822, opened 2026-03-26
ast: Simplification (and maybe fix) in `ctypeInheritanceCall`

fridis:
Replaced `!isCallToOUterRef() && !isField()` by `isConstructur()`, which is IMHO the main point here, a constructor sets a specific target type, while for a field or function, the result type is important.

This is an attempt to simplify the code introduced in #6818.

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -800,12 +800,13 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
   Call cotypeInheritanceCall(SourcePosition p, List<AbstractType> typeParameters, Resolution res, AbstractFeature that, Expr target)
   {
     var o = outer();
-    var oc = o == null || o.isUniverse()                            ? Universe.instance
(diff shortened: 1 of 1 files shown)


src/dev/flang/ast/AbstractFeature.java
```diff
@@ -800,12 +800,13 @@ && outer().isCotype()
...
     var o = outer();
-    var oc = o == null || o.isUniverse()                            ? Universe.instance
-      : target instanceof AbstractCall ac && !ac.isCallToOuterRef() && ! ac.calledFeature().isField()
-      ? ac.cotypeInheritanceCall(res, that)
-      : o.cotypeInheritanceCall(p, new List<>(o.selfType(),
-                                              o.genericsAsActuals().map(that::rebaseTypeForCotype)),
-                                res, that, null);
+    var oc = o == null || o.isUniverse()
+      ? Universe.instance
+      : target instanceof AbstractCall ac && ac.calledFeature().isConstructor()
```
michaellilltokiwa: What about choices? Maybe even:
```suggestion
      : target instanceof AbstractCall ac && ac.calledFeature().definesType()
```

PR #6864, opened 2026-04-07
ast: better human readable names for lambdas

fridis:
The human readable base name of the type of a lambda expression `x->x+1` used to be just `λ`, while the actual base name was something like `#FUN123`. With this PR, the human readable base name use the source code of the lambda and looks like `(λx->x+1)`.

The output of code like
```
f(code F : i32->R) =>
  say "$F $R {code 42}"
f x->x+1
f ($)
f (.as_u8)
f x->
  if x=3 then
    say "this is a really long lambda expression"
    nil
  else
    option 42*x
```
now is
```
Type of '(λx->x+1)' Type of 'i32' 43
Type of '(λ($))' Type of 'String' 42
Type of '(λ(.as_u8))' Type of 'u8' 42
Type of '(λx-> if x=3 then say "this is a really ...)' Type of 'option i32' 1764
```
instead of
```
Type of 'λ' Type of 'i32' 43
Type of 'λ' Type of 'String' 42
Type of 'λ' Type of 'u8' 42
Type of 'λ' Type of 'option i32' 1764
```
It turned out that the human readable names where used in names of identifiers in generated C and Java code, in particular outer ref fields. So I also had to change this and split up `AbstractFeature.qualifiedName` into a human and internal version.


michaellilltokiwa (approved):
Great stuff, should we do sth. similar for anonymous features?

PR #6873, opened 2026-04-07
Add priority queue interface with a binary heap–based implementation

simonvonhackewitz:


modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,365 @@
...
+
+  # does this queue contain elem?
+  #
+  public redef contains(elem T) bool =>
+    elements.contains elem
+
+
+  # create a string representation of the binary heap
+  #
+  public dbg_str String =>
```
maxteufel: please avoid using abbreviations in public API features

modules/base/src/container/Mutable_Priority_Queue.fz
```diff
@@ -0,0 +1,130 @@
...
+  # does this queue contain elem?
+  #
+  public contains(elem T) bool
+    pre T : property.equatable
+  => abstract
+
+
+  # all elements in this queue as an unordered Sequence
+  #
+  public as_unordered Sequence T => abstract
```
michaellilltokiwa: Not sure if/why this is needed?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,365 @@
...
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature mutable_priority_queue
```
michaellilltokiwa:
```suggestion
#  Source code of Fuzion standard library feature Binary_Heap_Queue
```

michaellilltokiwa (commented):
Lots of small(er) comments from me, sorry. Overall structure is pretty much what I would expect. We have a mutable queue definition with enqueue/dequeue functionality. Of to a good start!

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,365 @@
...
+#
+# -----------------------------------------------------------------------
+
+module:public Binary_Heap_Queue(
```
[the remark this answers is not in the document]
simonvonhackewitz: like this?

```
┌───┬───┐   ┌───────────────┬───┐
│   v   v   │               v   v
1   2   3   4   5   6   7   8   9 
    │   │   ^   ^   ^   ^
    └───────┴───┘   │   │
        │           │   │
        └───────────┴───┘
```

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,365 @@
...
+      sub_r := c.1.bind lvl
+
+      if sub_l.is_nil # leaf - no more subtrees
+        lvlres := [str.pad_center str_len]
+
+        lvlres
+      else
+        # width of current node, left and right subtree
+        cur_w, sub_l_w, sub_r_w :=
+          {
```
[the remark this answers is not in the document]
simonvonhackewitz: yes

```
./build/modules/base/src/container/Binary_Heap_Queue.fz:234:13: error 1: Inconsistent indentation
            sub_w := tmp_sub_l_w + tmp_sub_r_w + 1 # width of subtrees with space between them
------------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:231:9:
        cur_w, sub_l_w, sub_r_w :=
--------^
While parsing: exprs, parse stack: exprs, block (twice), elseBlockOpt, ifexpr (twice), term, opExpr, operatorExpr, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:235:13: error 2: Inconsistent indentation
            if sub_w >= str_len # subtrees are wider, need to pad current node
------------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:231:9:
        cur_w, sub_l_w, sub_r_w :=
--------^
While parsing: exprs, parse stack: exprs, block (twice), elseBlockOpt, ifexpr (twice), term, opExpr, operatorExpr, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:245:9: error 3: Inconsistent indentation
        lh := sub_l_w/2 # half width of left subtree, excluding center char
--------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:218:7:
      str := elements[_,i].as_string
------^
While parsing: exprs, parse stack: exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:246:9: error 4: Inconsistent indentation
        rh := sub_r_w/2 # half width of right subtree, excluding center char
--------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:218:7:
      str := elements[_,i].as_string
------^
While parsing: exprs, parse stack: exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:249:9: error 5: Inconsistent indentation
        sub_r_new := sub_r.or_else [$" "," "," "] .as_list.concat_list [$""].cycle.as_list # extension needed if the right side is only a leaf but the left side contains a subtree
--------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:218:7:
      str := elements[_,i].as_string
------^
While parsing: exprs, parse stack: exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:251:9: error 6: Inconsistent indentation
        lvlres:= [str.pad_center cur_w]
--------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:218:7:
      str := elements[_,i].as_string
------^
While parsing: exprs, parse stack: exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:255:9: error 7: Inconsistent indentation
        lvlres
--------^
Indentation reference point is ./build/modules/base/src/container/Binary_Heap_Queue.fz:218:7:
      str := elements[_,i].as_string
------^
While parsing: exprs, parse stack: exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, implFldOrRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit


./build/modules/base/src/container/Binary_Heap_Queue.fz:251:9: error 8: Duplicate feature declaration
        lvlres:= [str.pad_center cur_w]
--------^^^^^^
Feature that was declared repeatedly: 'container.Binary_Heap_Queue.dbg_str.lvl.lvlres'
originally declared at ./build/modules/base/src/container/Binary_Heap_Queue.fz:226:9:
        lvlres := [str.pad_center str_len]
--------^^^^^^
To solve this, consider renaming one of these two features, e.g., as 'lvlresʼ' (using a unicode modifier letter apostrophe 'ʼ' U+02BC) or adding an additional argument (e.g. '_ unit' for an ignored unit argument used only to disambiguate these two).


./build/modules/base/src/container/Binary_Heap_Queue.fz:232:13: error 9: Could not find called feature
            tmp_sub_l_w := sub_l.or_panic[0].codepoint_length
------------^^^^^^^^^^^
Feature not found: 'tmp_sub_l_w' (no arguments)
Target feature: 'container.Binary_Heap_Queue.dbg_str.lvl'
In call: 'tmp_sub_l_w'


./build/modules/base/src/container/Binary_Heap_Queue.fz:231:33: error 10: Destructuring mismatch between number of visible fields and number of target variables.
        cur_w, sub_l_w, sub_r_w :=
--------------------------------^
Found no visible argument fields
while there are 3 destructuring variables.
```

benchmarks/priority_queue_benchmark.fz
```diff
@@ -0,0 +1,126 @@
...
+    random_enqueue_dequeue 1E7
+    random_enqueue_dequeue 1E8
+
+
+
+  random_log_performance(queue_size i32) tuple time.histogram time.histogram =>
+
+    sample_size := 1E4 + 20
+
+    simple_random 13 ()->
```
michaellilltokiwa: What is this for? Seems this does not do anything when effect is instated

benchmarks/priority_queue_benchmark.fz
```diff
@@ -0,0 +1,126 @@
...
+    random_enqueue_dequeue 1E5
+    random_enqueue_dequeue 1E6
+    random_enqueue_dequeue 1E7
+    random_enqueue_dequeue 1E8
+
+
+
+  random_log_performance(queue_size i32) tuple time.histogram time.histogram =>
+
+    sample_size := 1E4 + 20
```
michaellilltokiwa: Strange number, needs a comment :-D

benchmarks/priority_queue_benchmark.fz
```diff
@@ -0,0 +1,126 @@
...
+          dur :=time.stopwatch (()->q.enqueue rand)
+          hist_enq.add dur
+        else
+          t
+
+      total_deq :=
+        for t := time.duration.zero, t + dur
+            i in 1..size
+        do
+          dur :=time.stopwatch (()->_ := q.dequeue)
```
michaellilltokiwa:
```suggestion
          dur := time.stopwatch (()->_ := q.dequeue)
```

benchmarks/priority_queue_benchmark.fz
```diff
@@ -0,0 +1,126 @@
...
+        else
+          t
+
+      say "{hist_enq}Total time to enqueue $size elements: $total_enq\n"
+      say "{hist_deq}Total time to dequeue $size elements: $total_deq\n"
+
+  show_random_enqueue_dequeue unit =>
+    say """
+        \n********************************************************************************
+        * Fill empty queue with random elements, then empty it                         *
```
michaellilltokiwa:
```suggestion
        {" Fill empty queue with random elements, then empty it ".pad_center 80 "*"}
```

benchmarks/priority_queue_benchmark.fz
```diff
@@ -0,0 +1,126 @@
...
+
+      hist_enq := time.histogram.new "Enqueue $size" nil
+      hist_deq := time.histogram.new "Dequeue $size" nil
+
+      total_enq :=
+        for t := time.duration.zero, t + dur
+            i in 1..size
+        do
+          rand := random.next_i32
+          dur :=time.stopwatch (()->q.enqueue rand)
```
michaellilltokiwa:
```suggestion
          dur :=vtime.stopwatch (()->q.enqueue rand)
```

michaellilltokiwa (commented):
I think the histogram is a good start.
What I'm missing is - the probably most important metric - is WCET (worst case execution time). An also some heuristic how the determine if this is within a certain range.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature Binary_Heap_Queue
+#
+# -----------------------------------------------------------------------
+
+# An mutable priority queue implemented using a binary heap
```
fridis:
```suggestion
# A mutable priority queue implemented using a binary heap
```

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+  # less means an element comes first, i.e. has higher priority
+  # that can be either the highest value (max-queue) or the lowest (min-queue)
+  #
+  less_or_equal Binary bool T T,
+
+  min_size option i64,
+
+  # unordered elements to construct the queue from
+  #
+  initial_elements option (Sequence T)) ref : container.Mutable_Priority_Queue T is
```
fridis: This is confusing, a LF helps: 

```suggestion
  initial_elements option (Sequence T)
  
  ) ref : container.Mutable_Priority_Queue T is
```

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+            down_heap rc
+
+      # always check left child, if swap was done with right child
+      # this will terminate immediately
+      if !less_or_equal elements[i] elements[lc]
+        swap i lc
+        down_heap lc
+
+
+  # add an element to the queue
```
fridis, line 201: Here, it would be important to document the worst-cased and amortized complexity.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+  public redef enqueue(elem T) unit =>
+    elements.add elem
+    up_heap (elements.length - 1)
+
+
+  # get and remove the element with highest priority
+  #
+  # this is the element with either minimum or maximum value
+  # depending on the chosen queue/comparator
+  #
```
fridis, line 216: what is the worst-case and amortized time complexity?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+  # is represented as following
+  #
+  #   ╭───┬───╮   ╭───────────────┬───╮
+  #   │   v   v   │               v   v
+  #  [1] [2] [3] [4] [5] [6] [7] [8] [9]
+  #       │   │   ^   ^   ^   ^
+  #       ╰───────┴───╯   │   │
+  #           │           │   │
+  #           ╰───────────┴───╯
+  #
```
fridis: Please explain what `ahnentafel ordering` means. My guess is that for any downward branch `a - b` in this tree, e.g. `[3] - [6]`, there must be `less_or_equal a b`. Is this right?

Below, you refer to the `heap property`. so you should define what is the heap property. Something like for all `i`, there must be `less_or_equal elements[i] elements[2*i+1] && less_or_equal elements[i] elements[2*i+2]`

Does the `heap property` include any requirement of the ordering of siblings like `[2]` and `[3]` or `[6]` and `[7]`?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+          nil => LM.array T .empty
+          i i64 => LM.array T .empty i
+      elem Sequence T => LM.array T .from_Sequence elem
+
+  # fix heap if created from initial elements
+  if initial_elements.ok
+    build_heap_fix 0
+
+
+  # restores the heap property on a randomly arranged array
```
fridis, line 113: what is the `heap property`, must be defined somewhere?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+      if !less_or_equal elements[i] elements[prio_child]
+        swap i prio_child
+        build_heap_fix prio_child
+
+
+  # for a given index return the index of the parent node in the tree
+  # if not already the root
+  #
+  parent(i I : integer) option I
+    pre debug 5 : i.as_i64 < elements.length
```
fridis: There are no public APIs that allow passing indices directly, so there is no benefit making the internal APIs support different index types since in the end we will need `i64` anyway. So just use `i64` throughout and avoid all the casts.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+        # swap with right child only if it is less_or_equal than left child
+        if less_or_equal elements[rc] elements[lc] && !less_or_equal elements[i] elements[rc]
+            swap i rc
+            down_heap rc
+
+      # always check left child, if swap was done with right child
+      # this will terminate immediately
+      if !less_or_equal elements[i] elements[lc]
+        swap i lc
+        down_heap lc
```
fridis: I think it would be clearer to separate these two steps, i.e., first find the smaller child, and then `swap` / `down_heap` with that child. Also, using `match` instead of `bind` avoids overhead of lambdas (which are not optimized well at the moment). 

So maybe try something like this:

```suggestion
  # first check if left or right is smaller
  smaller := match c.0   # left child
               lc => match c.1   # right child
                       rc  => if less_or_equal elements[rc] elements[lc] then rc else lc
                       nil => lc
               nil =>  option I nil
  match smaller
    s i64 =>
      if !less_or_equal elements[i] elements[s] then
        # swap i with the smaller child
        swap i s
        down_heap s
    nil => 
```
This also avoids the explanation `if swap was done with right child this will terminate immediately`, which I found hard to understand.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+    else
+      res := elements[0]
+      elements[0] := elements.remove_last.or_panic
+      down_heap 0
+      res
+
+
+  # enqueue elem and then dequeue (return and remove) the highest priority element
+  #
+  # this is more efficient than an enqueue followed by a dequeue
```
fridis: what does `more efficient` mean? What is the worst-case and amortized complexity?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+      elements[0] := elem
+      down_heap 0
+      res
+    else
+      elem
+
+
+  # get the first element without removing it from the queue
+  # nil iff queue is empty
+  #
```
fridis, line 252: What is the worst-case and amortized complexity?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+  public redef peek option T =>
+    elements.length = 0 ? nil : elements[0]
+
+
+  # does this queue contain elem?
+  #
```
fridis, line 260: What is the worst-case and amortized complexity?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+
+    elements.length = 0 ? "" : (lvl 0).as_string "\n"
+
+
+  # print the string representation of the binary heap
+  #
+  public dbg_say unit => # NYI: CLEANUP: remove debug code
+    say elements
+    say as_string_tree_pretty
+    say "peek=$peek"
```
fridis: This should IMHO not be part of this interface. Just `as_string*` should be enough.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+
+  # level in the binary tree, root is 0,
+  # direct children of root are 1, their children are 2, etc
+  #
+  level(i I : integer) i32 =>
+    (i.as_i64.as_f64 + 1) .log 2 .as_i32
+
+  # check if the heap property still holds
+  #
+  public check_heap_property bool => # NYI: CLEANUP: remove debug code
```
fridis: instead of removing this (as suggested by the NYI comment), you could add it after operation like `enqueue` or `dequeue` as follows
```
  check
    debug 5 : check_heap_property
```

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+
+    chk1(i, j I : integer) bool =>
+      if !less_or_equal elements[i] elements[j]
+        e := error "'{elements[i]}' at level {level i} should come after '{elements[j]}' at level {level j}"
+        say e
+        false
```
fridis, line 386: better `panic` or `fault` here, there should be no dependency on `io.Out`.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,488 @@
...
+  # type of the elements in the queue
+  #
+  public T type,
+
+  # comparator to determine priority of the elements
+  #
+  # less means an element comes first, i.e. has higher priority
+  # that can be either the highest value (max-queue) or the lowest (min-queue)
+  #
+  less_or_equal Binary bool T T,
```
fridis: I still think this should use `T type : property.orderable`, instead of adding its only mechanism. But this can be changed later. 

If you really wish to give APIs that permit providing a comparator, those APIs could wrap the elements into an `property.orderable` type based on that comparator. 

Just one thing: use `less_than_or_equal` or `lteq`, this is more commonly used with `than`.

modules/base/src/mutate/array.fz
```diff
@@ -33,13 +33,18 @@ module:public array (
        # length of the array to create
        public redef length i64,
 
+       # length below which the internal array wont shrink, must be multiple of 2
+       min_length i64,
+
```
fridis, line 38: I am still very unhappy about this. `mutate.array` is a fundamental and supposedly very simple data structure. Adding functionality here to shrink the size and to store the minimum makes this more complex and less efficient for _all_ uses of `mutate.array`, just to make the implementation of one use, `Binary_heap_Queue` a bit more convenient. 

This is not how this should be done. 

Instead, if you really need this, create a flexible array or similar that uses `mutate.array` and add your favorite functionality there!

For now, you can use this for your assignment, but I would like this to be cleaned up eventually.

modules/base/src/mutate/array.fz
```diff
@@ -71,6 +78,26 @@ is
     set length := length+1
 
 
+  # remove the last element of the this array
+  #
+  # amortized O(1) but worst case of O(n)
+  #
+  public remove_last option T =>
+    if length = 0
```
fridis, line 87: This requires a call to `check_and_replace` to make sure the corresponding mutate effect is instated. 

This is not good since it permits mutation without requiring a mutate effect. 

To avoid such bugs, I would like `mutate.array` to stay as simple as possible.

fridis (commented):
Good work! 

I have a number of comments, but mostly some more documentation and some code that I would like to see at other places, so nothing major to commplain. 

I have not checked the tests and benchmarks yet, but that should not be blocking. 

Some comments can be addressed later, so do not consider them as blocking your documentation etc.

modules/base/src/container/Binary_Heap_Queue.fz
fridis: You explain well how the tree is built, but you forgot what ordering properties exist between which elements, which is probably a fundamental part of the `heap property`.

modules/base/src/container/Binary_Heap_Queue.fz
simonvonhackewitz: thats why it says `NYI: CLEANUP: remove debug code`

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test priority_queue_test
+#
+# -----------------------------------------------------------------------
+
+priority_queue_test =>
```
fridis, line 26: please add a header comment like 

```suggestion
# tests for Fuzion base module feature `Binary_Heap_Queue` 
#
priority_queue_test =>
```
you can add more detail if you like.

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+    say (actual = expected ? "OK" : "FAILED: got '$actual' but expected '$expected'")
+
+
+  # expect(actual, expected T : property.equatable, err_msg option String) =>
+  #   if actual = expected
+  #     say "OK"
+  #   else
+  #     match err_msg
+  #       nil => say "FAILED: got '$actual' but expected '$expected'"
+  #       m String => say "FAILED: $m"
```
fridis: do not check in commented code unless there is a good reason for that and that reason is given in a comment.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,500 @@
...
+    elements.length
+
+
+
+
+
+  # Create a new priority queue that returns the smallest element first
+  # using natural ordering of the elements
+  #
+  public fixed redef type.empty_min_first container.Binary_Heap_Queue LM T
```
fridis, line 420: This and the following features should be called `empty` for simplicity, `<=` is the natural order.  See below comment for `max`-features.

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,500 @@
...
+  public fixed redef type.min_first_from(initial_elements Sequence T) container.Binary_Heap_Queue LM T
+    pre else T : property.orderable
+  =>
+    container.Binary_Heap_Queue LM T lteq nil initial_elements
+
+
+  # Create a new priority queue that returns the largest element first
+  # using natural ordering of the elements
+  #
+  public fixed redef type.empty_max_first container.Binary_Heap_Queue LM T
```
fridis, line 452: For later: All of these `max` features are not needed, and the corresponding `xyz_min_first` features should be renamed as `xyz`. 

Instead, if we want to make it easier to reverse the order, we should provide a wrapper `property.reverse_order(T type : property.orderable, v T)` and then if one wants a reverse order, one can wrap the elements using this.

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+    say q2.as_string_tree_pretty
+    say ""
+    say q2.as_string_tree_vertical
+    q2.enqueue 0.5
+    q2.enqueue 0.1
+    say ""
+    say q2.as_string_tree_pretty
+    say ""
+    say q2.as_string_tree_vertical
+    say ""
```
fridis, line 75: A test should IMHO not print the trees, but check the properties, as you do below. 

If you want to test `as_string*` or verify the internal representation, you can use `expect` here as well. But it is not so nice to mix `expect` and `say`

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+    expect q1.count 1
+    expect q1.peek "F"
+    expect q1.dequeue "F"
+    expect q1.is_empty true
+    expect q1.count 0
+
+
+    # Order of priorities
+    say "\nOrder of priorities:"
+    q2 := container.Binary_Heap_Queue lm f64 .max_first_from []
```
fridis, line 61: You first tested `empty_min_first`, now you test `max_first_from`. This is very arbitrary, a good test would test all the aspects of a queue created using `empty_min_first`, and then repeat all the tests with `min_first_from`, and then `empty_max_first`, and then `max_first_from`. 

Yes, the number of tests needed explodes here. This is why it is a better idea to have `property.reverse_order` and test that completely separately.

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+        head := q2.dequeue
+    while head.ok
+    do
+      yak "$prev: "
+      expect (prev.or_panic >= head.or_panic) true
+    else
+      say prev
+
+
+    # Edge cases
```
fridis, line 86: describe in a comment what the edge cases are that you are testing.

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+    expect q3.dequeue "aaa"
+    say ""
+    say q3.as_string_tree_pretty
+    say ""
+
+    # Heap property min-heap
+    say "\nHeap property min-heap:"
+    q4 := container.Binary_Heap_Queue lm i32 .empty_min_first
+
+    for i in (1..10).reverse ++ [i32.min, i32.max]
```
fridis, line 118: Describe what you test and use `expect` to verify. Why `(1..10).reverse ++ [i32.min, i32.max]`?

To be more systematic, test queues with different number of elements, including really many, e.g., 
```
  for n := 0, n+1+n/2
  while n < 1000000 do
    ...
```
Then test different inputs, e.g., 
```
    inputs := [ id (Sequence i32) 0..n,                    # ascending
                                  n .. 0 : -1              # descenging
                                  [42]*n                   # all the same
                                  [i32.max] ++ 1..n        # first is outlier 
                                  (0..(n-1)) ++ [i32.min]  # last is outlier  
                                  # add more if you like , but document what they are
                  ]
    for input in inputs do
```
and then perform the tests using something like (pseudo-code)
```  
      q := PrioQueue.empty
      for i in inputs do
        q.enqueue i
      for expected in input.sort 
          cnt := n, cnt-1
      do
        expect q.count cnt          # check correct count
        expect q.peek expected      # check correct peek
        expect q.dequeue expected   # check correct dequeue
      expect q.count 0              # check correct final count
      expect q.peek nil             # check correct final peek
      expect q.dequeue nil          # check correct final dequeue
```

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+
+    q6 := container.Binary_Heap_Queue lm i32 .empty_comp (id ((i32, i32)->bool) (a, b)->((digit_sum a) <= (digit_sum b)))
+    _ := q6
+    simple_random 42 ()->
+      for i in 0..9
+          r := random.next_i32 199 - 99
+      do
+        say "enqueue $r"
+        q6.enqueue r
+        say q6.as_string_tree_pretty
```
fridis, line 165: Ideally, you run all the test cases above additionally with a custom comparator, i.e, add one loop level for this. 

So, instead of `q := PriorityQueue.empty` in my comment above, you would have 
```
for q in [PriortyQuue.empty, 
          PriorityQueue.empty_comp, 
          etc. ...]
do
  ...
```

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+
+
+
+
+  sort_test_random(random_test_size i32) =>
```
fridis, line 176: what does this test do? Add comment!

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+
+
+
+  # mix enqueue and dequeue and compare with sorted list
+  #
+  # NYI: OPTIMIZATION: slow for higher enq rate due to sorting of list
+  #
+  # 🡩 : enqueue operation
+  # 🡫 : dequeue operation returned smallest element
+  # - : dequeue returned noting, queue expected to be empty
```
fridis:
```suggestion
  # - : dequeue returned nothing, queue expected to be empty
```

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+          new_sort :=
+            if enq
+              e := random.next_i32 test_size
+              yak "🡩"
+              q.enqueue e
+              (sort ++ [e]).sort.as_list
+            else
+              match q.dequeue
+                e i32 => yak (e = sort[0] ? "🡫" : "\nError in round $i expected '{sort[i]}' but cot '$e'\n")
+                nil => yak (sort.is_empty ? "-" : "\nError queue is empty but sort is $sort\n")
```
fridis, line 243: why don't you use `expect`?

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test priority_queue_test
+#
+# -----------------------------------------------------------------------
+
+priority_queue_test =>
+
+  exp_msg(actual, expected T : property.equatable, err_msg String) =>
+    say (actual = expected ? "OK" : "FAILED: $err_msg")
```
fridis: This seems unused. remove!

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+
+  expect(actual, expected T : property.equatable) =>
```
fridis, line 31: Add a comment that says what this does

tests/priority_queue/priority_queue_test.fz
```diff
@@ -0,0 +1,226 @@
...
+    say (actual = expected ? "OK" : "FAILED: got '$actual' but expected '$expected'")
```
fridis: below, I suggest using `expect` much more often. It might make sense to output nothing instead of '"OK"'.

fridis (commented):
The test is already good in the sense that it covers diifferent and important cases. 

I made a number of suggestions how this test could be more systematic. Please just pick and chose comments that you would like to include in your thesis and ignore what is too much for now. 

Some design decisions (like using a compare function instead of property.orderable, adding min/max variants, etc.) result in an explosion of configurations and make it harder to test this.  At a later point in time, I would like you to revisit these decisions.

modules/base/src/container/Mutable_Priority_Queue.fz
```diff
@@ -0,0 +1,142 @@
...
+# An abstract mutable priority queue, returning the element with the highest priority first
+#
+public Mutable_Priority_Queue(public T type) ref is
+
+  # add an element to the queue
+  #
+  public enqueue(elem T) unit => abstract
+
+
+  # iteratively add all element to the queue
```
maxteufel:
```suggestion
  # iteratively add all elements to the queue
```

modules/base/src/container/Mutable_Priority_Queue.fz
```diff
@@ -0,0 +1,142 @@
...
+
+
+  # number of elements in this queue
+  #
+  public count i64 => abstract
+
+
+  # NYI: UNDER DEVELOPMENT:
+  # Does remove make sense? If so should there be a remove_one and remove_all?
+  # public remove(elem T) => abstract
```
maxteufel: do you know an answer on this by now?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,500 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature Binary_Heap_Queue
+#
+# -----------------------------------------------------------------------
+
+# A mutable priority queue implemented using a binary heap
+#
+# Time complexity of internal operations is amortized, du to expanding/shrinking
```
maxteufel:
```suggestion
# Time complexity of internal operations is amortized, due to expanding/shrinking
```
is this comment still true?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,500 @@
...
+      (option left, id (option i64) (right.as_i64 < elements.length ? right : nil))
+    else
+      (option i64 nil, option i64 nil)
+
+
+  # swap elements for the given two indices
+  #
+  swap(a, b i64) unit
+    pre debug: (i64 0) <= a.as_i64 < elements.length
+        debug: (i64 0) <= b.as_i64 < elements.length
```
maxteufel: i don't think you should need the `.as_i64`, do you?

modules/base/src/container/Binary_Heap_Queue.fz
```diff
@@ -0,0 +1,500 @@
...
+  #
+  less_or_equal Binary bool T T,
+
+  min_size option i64,
+
+  # unordered elements to construct the queue from
+  #
+  initial_elements option (Sequence T)
+
+) ref : container.Mutable_Priority_Queue T is
```
fridis, line 56: Why is this a `ref` type? Only template features should be `ref`, i.e., features that have `abstact` inner features or that are likely to be used via inheritance and require redefition.

simonvonhackewitz: moved to #7024

tests/priority_queue/priority_queue_test.fz
simonvonhackewitz: To be able to use a custom error message with a leading newline.

tests/priority_queue/priority_queue_test.fz
simonvonhackewitz: I've tried to mix different combinations so that the features are all tested once.
I didn't want to test all possible combinations, because either the test would run forever or it could only test small samples.

michaellilltokiwa (approved):
Good work!

PR #6892, opened 2026-04-08
ast: cleanup misuse of `propagateExpectedType`

michaellilltokiwa:
propagateExpectedType was used to add fields in match. Now this is done separately in resolveSyntaxSugar2.

The main _idea_ is to have `Match` act like an expression that produces a result and only in resolveSyntaxSugar2, assign the result of the cases to a field and replace the Match by a call the field (that is unless the field is a result field).


src/dev/flang/ast/Impl.java
```diff
@@ -297,6 +297,7 @@ else if (rt instanceof FunctionReturnType)
    */
   public void visit(FeatureVisitor v, AbstractFeature outer)
   {
+    v.action(this);
```
michaellilltokiwa, line 300: This is moved up, since in syntaxSugar2 we first _try_ to assign match to a resultfield. Only if this is not possible - basically if the match is a subject of another match - it creates its own result field.

modules/base/src/list.fz
```diff
@@ -596,10 +596,11 @@ is
   # 'list.this' and the empty list 'nil'.
   #
   public redef tails list (list A) =>
+    lazy_value => (list.this ? nil => nil | c Cons => c.tail.tails)
```
michaellilltokiwa, line 599: Needed to be changed due to implementation restriction that a lazy may not contain a field (`c`).
Not sure why this worked before this change...

tests/reg_issue2559/reg_issue2559.fz
```diff
@@ -21,4 +21,8 @@
 #
 # -----------------------------------------------------------------------
 
-say (0 : (match true { a true_ => { say a; nil }; b false_ => { say b; nil } }))
+lazy_value => match true { a true_ => { say a; nil }; b false_ => { say b; nil } }
```
michaellilltokiwa, line 24: Needed to be changed due to implementation restriction that a lazy may not contain a field (a,b).
Not sure why this worked before this change...

src/dev/flang/ast/Expr.java
```diff
@@ -378,7 +378,10 @@ void loadCalledFeature(Resolution res, Context context)
    */
   Expr assignToField(Resolution res, Context context, Feature r)
   {
-    return new Assign(res, pos(), r, this, context);
+    var result = new Assign(res, pos(), r, this, context);
+    result.wrapValueInLazy(res, context);
+    result.unwrapValue  (res, context);
```
maxteufel, line 383: I don't understand why this happens/is necessary?

maxteufel (approved):
Looks good, but I have one question.

src/dev/flang/ast/Expr.java
michaellilltokiwa: This is usually done in TYPES_INFERENCING phase. Since we do the assign to field for matches in SYNTAX_SUGAR2 we need to do this here for these cases.

Issue #6895, opened 2026-04-09
C backend does not join threads by default


maxteufel:
The JVM backends joins all threads at the end of the program implicitly, the C backend does not. Test with this example (and remove the comment around the join as necessary):
```
ex_threads1 =>
  t1 := concur.threads.spawn ()->
    now := time.nano.read
    end_of_time := now + (time.duration.min 60)
    period := time.frequency 2 (time.duration.s 1)
    _ := time.nano.run_periodic now period end_of_time ()->
      time.nano.sleep (time.duration.ms 100)
      say "hello from one"

  _ := t1 /* .join */
```

michaellilltokiwa:
This is due to no `finally` being run for default effects. I believe this will not be an issue when we get rid of default effects.

```
  # join all running threads
  #
  public redef finally unit =>
    running.for_each t->
      t.join
```

Issue #6341, opened 2025-12-01
`expanding_array` is not thread safe (yet)


michaellilltokiwa:
and since it is used in String.concat, concatting Strings is also potentially not thread safe.

PR #6905, opened 2026-04-09
base: provide wormholes to teleport effects between threads

fridis:
This PR first introduces an API to obtain all the effect types used in the system. 

This API is then used to replace the default effect mechanism by an effect `big_bang_singularity` that iterates over all effects and instates default instances where available. 

Then, this API is used to create a snapshot (`wormhole`) of the current environment that may be `teleport`ed to another environment. This is used for new threads to inherit the effect environment from the thread they were spawned from, but this API is also available for users 

There are still some ToDos here:

- helper features like `io.out` that just perform `io.Out.env` are no longer needed since the code to instate the default effect has been removed. We should better use `effect_type.env` directly instead.
- `effect.default` is still used at some places and some cleanup like removing `default0` is needed.
- It is unclear how well DFA supports wormholes
- there are currently no checks that teleporting an effect would actually be sound. We need some way to distinguish teleportable effects from those that are not
- also, the lifespan of wormholes is not enforced. It might make sense to disable a wormhole once the original effect is no longer instated or similar
- it should be made impossible to instate an effect like `io.Out.default_value` to hide the need of this effect.


michaellilltokiwa:
Very cool PR. Not done reading this yet but a random thought.
Apart from potential soundness issues, there is also the issue that this potentially enables certain security issues,
think e.g. `random` being _wormholed_. We need to find a way to at least minimize these potential fatalities. 

Btw this probably _fixes_ some open issue: #6895, #6341

fridis:
@michaellilltokiwa , thanks for the comment. I do not expect that this fixes any threading related issues yet, but this gives a basis to do that. My next plan is to work on `mutate` effects that can be shared between threads. 

The biggest possible attack vector this open is that code that has certain permissions (i.e, instated effects) may now create a wormhole and pass these on to other code, including code running in a different thread if that wormhole gets passed on to that thread.

modules/base/src/concur/threads.fz
```diff
@@ -37,9 +37,30 @@ public threads (
...
+      ewh.teleport code
+
+
+  # spawn a new thread without inheriting the current thread's
+  # effect environment, i.e., using a new, empty effect environment.
+  #
+  # This is intended for use in, e.g., thread pools that are used to perform
+  # tasks that may require different effects.
+  #
+  public spawn_detached(# code ot run int he new thread without an new, empty effect environment
```
michaellilltokiwa, line 61: Is detached a good word here? There is e.g. `pthread_detach` which allows:
"When a detached thread terminates, its resources are automatically released back to the system without the need for another thread to join
       with the terminated thread." 

But the detaching here has nothing to do with this.

tests/effect_wormholes/effect_wormholes.fz
```diff
@@ -0,0 +1,90 @@
...
+#
+#  Source code of Fuzion test effect_wormholes
+#
+# -----------------------------------------------------------------------
+
+# This test uses two new `io.Out` effect handlers `out_A` and `out_B` and
+# performes nested `instate`s for these creating worm holes for the default
+# environment and the the environments using `out_A` and `out_B`.
+#
+# It spawns threads checking whilie `out_A` and `out_B` are instated and uses
```
michaellilltokiwa:
```suggestion
# It spawns threads checking while `out_A` and `out_B` are instated and uses
```

tests/effect_wormholes/effect_wormholes.fz
```diff
@@ -0,0 +1,90 @@
...
+# warmholes in the new threads to change the environments
```
michaellilltokiwa:
```suggestion
# wormholes in the new threads to change the environments
```

michaellilltokiwa (approved):

modules/base/src/concur/threads.fz
fridis: I see that new thread is not attached to the current thread's effect environment, which is why I chose `detached`. But I am open for suggestions for better names!

michaellilltokiwa: What about disconnected or isolated?

fridis: or `spawn_no_inherited_effects`, `spawn_virgin`, `spawn_no_effects`? I am still thinking...

fridis: `spawn_without_inherited_effects` is used now.

PR #6909, opened 2026-04-10
allow specifying scheduling policy and priority of a thread

maxteufel:
This is limited to the C backend due to the JVM not exposing any
interface for doing so. Likewise, no implementation for Windows is
provided yet.

### DIFF
--- a/include/posix.c
+++ b/include/posix.c
@@ -74,6 +74,11 @@ static_assert(SIGPIPE == 13, "signal definition different than expected");
 static_assert(SIGALRM == 14, "signal definition different than expected");
 static_assert(SIGTERM == 15, "signal definition different than expected");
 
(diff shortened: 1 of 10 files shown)


include/posix.c
```diff
@@ -552,6 +552,18 @@ void fzE_thread_join(void * thrd)
 }
```
michaellilltokiwa, line 559: It may make sense to assert the magic numbers:

```
static_assert(SCHED_OTHER == 0, "magic number diff that exp");
static_assert(SCHED_FIFO == 1, "magic number diff that exp");
static_assert(SCHED_RR == 2, "magic number diff that exp");
static_assert(ENOSYS == 38, "magic number diff that exp");
```

michaellilltokiwa (approved):
Do we also need to be able to set this at thread start?
I think at thread start is probably the more common case?

PR #6915, opened 2026-04-13
fe: add reg test #5998

### LINKED ISSUE #5998: An arbitray feature that inherits from `Function` cannot be used for lambda
Try this code that declares `myf` as its function type
```
myf : Function bool i32 is
q(f myf) => f.call 1 |> say

q a->a>0
```
this produces this error
```
 > ./build/bin/fz myfun.fz 

/home/fridi/fuzion/clean/fuzion.5/myfun.fz:4:3: error 1: Wrong number of arguments in lambda expression
q a->a>0

Lambda expression has one argument while the target type expects -1 arguments.
Arguments of lambda expression: 'a'
Expected function type: myf
To solve this, remove 2 arguments from the list 'a' before the '->' of the lambda expression.

one error.
```
while it should best just work.

michaellilltokiwa:
fixes #5998

This just works:
```
myf : Function bool i32 is
q(MYF type: myf, f MYF) => f.call 1 |> say

q a->a>0
```

The error message is now much better. So I just had to add the reg test.

### DIFF
--- a/tests/reg_issue5998/Makefile
+++ b/tests/reg_issue5998/Makefile
@@ -1,0 +1,25 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


maxteufel:
I'm not completely happy with the claim that this fixes the mentioned issue, when @fridis writes there that it should work (notably without your workaround using type parameters).

michaellilltokiwa:
@maxteufel Oh sorry, I forgot to mention this. I discussed this with @fridis. He does not think this should work anymore. The error message is fine. myf is a value. There is no (reasonable) way the lambda can act as this value.

maxteufel (approved):

PR #6927, opened 2026-04-14
Added support for addition/subtraction of time to dates

Nimux2:


modules/base/src/time/date_time.fz
```diff
@@ -85,11 +85,36 @@ is
                               "T$hour_as_string:$minute_as_string:$second_as_string.$nano_second_as_string")
 
 
-  # # NYI
-  # infix + (other time.duration) date_time is
+  # internal helper: total nanoseconds since unix epoch
+  #
+  public nanos u64
+  =>
+    unix_time_stamp.as_u64 * 1000000000 + nano_second.as_u64
```
michaellilltokiwa:
```suggestion
    unix_time_stamp.as_u64 * 1E9 + nano_second.as_u64
```

modules/base/src/time/date_time.fz
```diff
@@ -85,11 +85,36 @@ is
                               "T$hour_as_string:$minute_as_string:$second_as_string.$nano_second_as_string")
 
 
-  # # NYI
-  # infix + (other time.duration) date_time is
+  # internal helper: total nanoseconds since unix epoch
```
michaellilltokiwa, line 93: If this is supposed to be internal, consider removing the public visibility?

michaellilltokiwa (commented):
A good start and congrats to your first PR in the main repo!

I think this method will have problems with dates before 1970 and dates after 2038 due to using `unix_time_stamp`. Can you please add tests with concerning dates?

tests/lib_time_date_time/lib_time_date_time.fz
```diff
@@ -50,7 +51,18 @@ lib_time_date_time =>
   say "c is_before a = {c < a}"
   say "c is_before b = {c < b}"
   say "c is_before c = {c < c}"
-
+  say  "{a} + {duration} = {a + duration}"
+  say  "{a} - {duration} = {a - duration}"
+  check a + duration = time.date_time 2024 1 30 11 45 42 100000000
+  check a - duration = time.date_time 2024 1 30 9 14 42 100000000
+  #test := time.date_time 2280 12 31 23 59 59 0
```
michaellilltokiwa: You can catch a precondition fault like this, to test if the failure actually occurs:
```
(fuzion.runtime.pre_fault
    .try ()->
      code...
    .catch s->
      alternative_code...
``

File: tests/lib_time_date_time/lib_time_date_time.fz [hunk 4]
```diff
@@ -26,6 +26,7 @@ lib_time_date_time =>
   a := time.date_time 2024 1 30 10 30 12 1E8
   b := time.date_time 2024 1 30 10 30 12 1E7
   c := time.date_time 2023 1 30 10 30 12 1E8
+  duration := time.duration.h 1 + time.duration.min 15 + time.duration.s 30
```
michaellilltokiwa: I would suggest a different name here, maybe `one_hour_15min_30s`?

modules/base/src/time/date_time.fz
```diff
@@ -231,5 +352,5 @@ days_in_month(year, month i32) =>
 public is_valid_date(year, month, day i32) bool
   pre
     debug: 1 ≤ month ≤ 12
-=>
+  =>
```
michaellilltokiwa: I think we usually do not indent the `=>` as far as the pre

modules/base/src/time/date_time.fz
```diff
@@ -170,6 +224,67 @@ is
     time.date_time year month day.as_i32 hour.as_i32 minute.as_i32 second.as_i32 0
 
 
+  # create a date_time from given seconds since 1970-01-01T00:00:00 (i64 version for negative values, i.e. dates before 1970)
+  #
+  public type.from_seconds_i64(n i64) time.date_time
+    pre debug    : n < 3E11 # this restriction is arbitrary (year 11476)
+  =>
+    total_seconds := n
+    days_raw := total_seconds / 86400 # number of days since 1970-01-01 (can be negative for dates before 1970)
```
michaellilltokiwa: I would suggest to give this numbers e.g. `86400` a name.
e.g. `days_since_1970 := 86400`

modules/base/src/time/date_time.fz
```diff
@@ -170,6 +224,67 @@ is
     time.date_time year month day.as_i32 hour.as_i32 minute.as_i32 second.as_i32 0
 
 
+  # create a date_time from given seconds since 1970-01-01T00:00:00 (i64 version for negative values, i.e. dates before 1970)
+  #
+  public type.from_seconds_i64(n i64) time.date_time
+    pre debug    : n < 3E11 # this restriction is arbitrary (year 11476)
```
michaellilltokiwa, line 225: If the restriction is arbitrary then it should just be removed, I think. Or what is the rational behind it?

modules/base/src/time/date_time.fz
```diff
@@ -23,11 +23,18 @@
 #
 # -----------------------------------------------------------------------
 
+
+
 # Represents a date and a time in the Gregorian calendar, without any specification of
```
fridis, line 29: no more than 2 blank lines, please:-)

modules/base/src/time/date_time.fz
```diff
@@ -199,6 +315,12 @@ is
 
     ternary_compare.is_yes_or_unknown
 
+public supported_year(y i32) bool
+  pre
+    debug: y <= 2260
+    debug: 1650 <= y
```
fridis: This should work:
```suggestion
    debug: 1650 <= y <= 2260
```

modules/base/src/time/date_time.fz
```diff
@@ -199,6 +315,12 @@ is
 
     ternary_compare.is_yes_or_unknown
 
+public supported_year(y i32) bool
```
fridis: please add a comment.

modules/base/src/time/date_time.fz
```diff
@@ -23,11 +23,18 @@
...
 # Represents a date and a time in the Gregorian calendar, without any specification of
 # a time zone or reference point.
 #
+
+
+
 public date_time(public year, month, day, hour, minute, second, nano_second i32) : property.orderable
 pre
+  debug: 1650 ≤ year
+  debug: year ≤ 2260
```
fridis: better `debug: supported_year year`

modules/base/src/time/date_time.fz
```diff
@@ -170,6 +224,68 @@ is
...
+        m, r1 :=
+          for rest := rest_days, rest - days_in_month.as_i64
+              mm := 12, mm-1
+              days_in_month := time.days_in_month y mm
+          while rest > days_in_month.as_i64
+          else
+            (mm, rest)
+
+        days_in_month_local := time.days_in_month y m
+        (y, m, days_in_month_local.as_i32 - r1.as_i32 + 1) # for dates before 1970, the day is calculated as the number of days in the month minus the leftover days, because we calculated backwards
```
fridis: I think this should work just as well
```suggestion
        for rest := rest_days, rest - days_in_month.as_i64
            mm := 12, mm-1
            days_in_month := time.days_in_month y mm
        while rest > days_in_month.as_i64
        else
          days_in_month_local := time.days_in_month y mm
          (y, mm, days_in_month_local.as_i32 - rest.as_i32 + 1) # for dates before 1970, the day is calculated as the number of days in the month minus the leftover days, because we calculated backwards
```
Similarly, this code could move into the `else` clause of the previous loop.

modules/base/src/time/date_time.fz
```diff
@@ -199,6 +306,11 @@ is
 
     ternary_compare.is_yes_or_unknown
 
+public supported_year(y i32) bool # Causes a debug error if the given year would cause an overflow, returns true otherwise
+  pre
+    debug: 1650 <= y <= 2260 # Dates too far away from the unix epoch (1970-01-01) would cause an overflow in the unix_time_stamp calculation, which is currently implemented using a 64-bit integer counting nanoseconds since the epoch.
+  =>
+  true
```
maxteufel: this feature should just return `1650 ≤ y ≤ 2260` as its result:
```suggestion
public supported_year(y i32) bool =>
  1650 <= y <= 2260 # Dates too far away from the unix epoch (1970-01-01) would cause an overflow in the unix_time_stamp calculation, which is currently implemented using a 64-bit integer counting nanoseconds since the epoch.
```

If you want to cause a contract failure in the middle of a feature, then use a check-condition on `debug: supported_year y`. Note that it is probably not a good idea to cause a contract failure in the middle of a feature in this case; it should more likely be handled with graceful error handling or a precondition failure already.

modules/base/src/time/date_time.fz
```diff
@@ -136,6 +166,27 @@ is
     days.as_u64*86400 + time_seconds.as_u64
 
 
+  # convert to unix time stamp (i64 version for dates before 1970)
+  #
+  public unix_time_stamp_i64 i64
+  =>
+    _ := supported_year year
```
maxteufel: You can't just apply my hints from here and hope that it will all work out. The suggestions are just clarifications on what I mean.

This will now do nothing: `supported_year year` will return a `bool`, and that will just get ignored (`_ :=`). You need to adjust this, and all the other places where this is done, like I suggested.

Nimux2: supported_year is already tested in date_time's preconditions, which ends up with a similar result to testing it in each method. I could remove the "_ := supported_year year" parts, or do the reverse and only test inside methods.

public date_time(public year, month, day, hour, minute, second, nano_second i32) : property.orderable
pre
  debug: supported_year year
  debug: 1 ≤ month ≤ 12
  debug: is_valid_date year month day
  debug: 0 ≤ hour ≤ 23
  debug: 0 ≤ minute ≤ 59
  debug: 0 ≤ second ≤ 60 # possible leap seconds
  debug: 0 ≤ nano_second ≤ 1E9
is

Nimux2:
I think it's ready now. Could someone review and merge it ?

michaellilltokiwa (commented):
It looks good now! One more thing should be changed. Some of the comments need line breaks, since they are too long.

Nimux2:
@michaellilltokiwa It's done.

modules/base/src/time/date_time.fz
```diff
@@ -199,6 +308,9 @@ is
 
     ternary_compare.is_yes_or_unknown
 
+public fixed supported_year(y i32) bool =>
+  1650 <= y <= 2260 # Dates too far away from the unix epoch (1970-01-01) would cause an overflow in the unix_time_stamp calculation,
+  # which is currently implemented using a 64-bit integer counting nanoseconds since the epoch.
```
michaellilltokiwa, line 318:
```suggestion
public fixed supported_year(y i32) bool =>
  # Dates too far away from the unix epoch (1970-01-01) would cause an overflow in the unix_time_stamp calculation,
  # which is currently implemented using a 64-bit integer counting nanoseconds since the epoch.
  1650 <= y <= 2260
  

```
I think we currently mostly put the comments before the expression they are concerning.
Can you change this, then the PR is ready to merge.

michaellilltokiwa (approved):

PR #6929, opened 2026-04-14
parser: fix precendence table, revert #6793

### LINKED ISSUE #6925: Multiple successive operators should be allowed
I think the fix #6793 might be wrong:

The following code using `%` works
```
 > ./build/bin/fz -e 'unit.prefix % => {}; (% % unit) |> say'
unit
```
While it does not work using `:`, as we see here
```
 > ./build/bin/fz -e 'unit.prefix : => {}; (: : unit) |> say'

command line:1:27: error 1: Multiple successive operators are not allowed.
unit.prefix : => {unit}; (: : unit) |> say


one error.
```
This is strange, there is something wrong in the handling of `:`, maybe related to left vs. right associativity?
In any case, I think the error `Multiple successive operators are not allowed` should be removed, the example should be parsed as `((unit.prefix :).prefix :).infix |> say`

michaellilltokiwa:
fixes #6925

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2605,12 +2605,6 @@ public class AstErrors extends ANY
      skw("fixed")+ " fixed is only allowed on function features not in universe.");
   }
 
-  public static void multipleOperatorsFound(SourcePosition p)
-  {
-    error(p,
-      "Multiple successive operators are not allowed.", "");
-  }
-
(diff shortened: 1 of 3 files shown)


maxteufel (approved):

michaellilltokiwa:
The actual fix is to not use `-1` for the precendence of `:` but `0`. This is due to the logic in OpExpr.toExpr()

maxteufel:
maybe add a comment above the table then stating that a non-negative integer is required.

Issue #6914, opened 2026-04-13
static assertion of SCHED_ magic numbers fails on MacOS


michaellilltokiwa:
```
./build/include/posix.c:77:15: error: static assertion failed due to requirement '1 == 0': magic number different than expected
   77 | static_assert(SCHED_OTHER == 0, "magic number different than expected");
      |               ^~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/pthread/pthread_impl.h:54:36: note: expanded from macro 'SCHED_OTHER'
   54 | #define SCHED_OTHER                1
      |                                    ^
./build/include/posix.c:78:15: error: static assertion failed due to requirement '4 == 1': magic number different than expected
   78 | static_assert(SCHED_FIFO == 1, "magic number different than expected");
      |               ^~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/pthread/pthread_impl.h:55:36: note: expanded from macro 'SCHED_FIFO'
   55 | #define SCHED_FIFO                 4
      |                                    ^
./build/include/posix.c:80:15: error: static assertion failed due to requirement '78 == 38': error number different than expected
   80 | static_assert(ENOSYS == 38, "error number different than expected");
      |               ^~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/errno.h:209:25: note: expanded from macro 'ENOSYS'
  209 | #define ENOSYS          78              /* Function not implemented */
      |                         ^
``

michaellilltokiwa:
This is fixed now, #6939, #6940

PR #6939, opened 2026-04-15
include/posix: handle different SCHED_ magic constants on macOS

maxteufel:
Partly addresses #6914.

### DIFF
--- a/include/posix.c
+++ b/include/posix.c
@@ -74,9 +74,6 @@ static_assert(SIGPIPE == 13, "signal definition different than expected");
 static_assert(SIGALRM == 14, "signal definition different than expected");
 static_assert(SIGTERM == 15, "signal definition different than expected");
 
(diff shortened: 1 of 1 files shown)


include/fz.h
```diff
@@ -423,6 +423,11 @@ void * fzE_thread_create(void *(*code)(void *),
 // NYI: UNDER DEVELOPMENT:  add return value
 void fzE_thread_join(void * thrd);
 
+/*
+ * Convert internal policy number to system policy number.
+ */
+int fzE_thread_setschedparam_convert_policy(int policy);
```
michaellilltokiwa: I don't think this needs to be added to `fz.h`. We don't want to use this from fuzion?

michaellilltokiwa (approved):

michaellilltokiwa (approved):

PR #6943, opened 2026-04-15
rename: `byte_length`/`codepoint_length` as `byte_count`/`codepoint_count`

### LINKED ISSUE #6309: String, rename byte_length as byte_count?
The name is misleading, I think:
```
# length of this string in bytes
  public byte_length i32 => utf8.count

  # length of this string in codepoints
  public codepoint_length i32 => codepoints.count
```

michaellilltokiwa:
fixes #6309


include/shared.c
maxteufel, line 1: not sure if this should be changed here, the length is statically known here

maxteufel (approved):

include/shared.c
michaellilltokiwa: Yes, I reverted it now.

PR #6948, opened 2026-04-15
fuir: improve error message, "Call has an ambiguous result type..."

### LINKED ISSUE #6933: confusing error, with pos in base.fum for small example
First error seems okay, second one is weird:
```
Y ref is

  c option Y.this := Y.this
  say (c.and_then x->Y.this)

_ := Y
```
```
home/sam/playground/test.fz:12:7: error 1: Call has an ambiguous result type since target of the call is a 'ref' type.
  say (c.and_then x->Y.this)

The result type of this call depends on the target type.  Since the target type is a 'ref' type that may represent a number of different actual dynamic types, the result type is not clearly defined.
Called feature: 'switch.and_then'
Original result type: 'option Y.this'
Type depending on target: 'Y.this'
Target type: 'Y'
To solve this, you could try to use a value type as the target type of the call, e,g., 'switch A B', or change the result type of 'switch.and_then' to no longer depend on 'Y.this'.


{base.fum}/switch.fz:77:14: error 2: Call has an ambiguous result type since target of the call is a 'ref' type.
      a A => f a

The result type of this call depends on the target type.  Since the target type is a 'ref' type that may represent a number of different actual dynamic types, the result type is not clearly defined.
Called feature: 'Function.call'
Original result type: 'option Y.this'
Type depending on target: 'Y.this'
Target type: 'Y'
To solve this, you could try to use a value type as the target type of the call or change the result type of 'Function.call' to no longer depend on 'Y.this'.

```

michaellilltokiwa:
now uses the clazzes type as the target type
fixes #6933

### DIFF
--- a/tests/reg_issue4248/reg_issue4248.fz.expected_err
+++ b/tests/reg_issue4248/reg_issue4248.fz.expected_err
@@ -6,7 +6,7 @@ The result type of this call depends on the target type.  Since the target type 
 Called feature: 'e.me'
 Original result type: 'e.this'
 Type depending on target: 'e.this'
-Target type: 'p.this'
+Target type: 'p.me'
 To solve this, you could try to use a value type as the target type of the call, e,g., 'e', or change the result type of 'e.me' to no longer depend on 'e.this'.
 
 one error.
--- a/src/dev/flang/fuir/Clazz.java
+++ b/src/dev/flang/fuir/Clazz.java
@@ -1857,7 +1857,7 @@ class Clazz extends ANY implements Comparable<Clazz>
     var err = new List<Consumer<AbstractCall>>();
     var ft = t; // final variant of t to be used in lambda
     BiConsumer<AbstractType, AbstractType> foundRef = (from,to) ->
-      { err.add((c)->AstErrors.illegalOuterRefTypeInCall(c, false, feature(), ft, from, to)); };
+      { err.add((c)->AstErrors.illegalOuterRefTypeInCall(c, false, feature(), ft, from, _type)); };
 
     t = handDown(t, select, foundRef, inh);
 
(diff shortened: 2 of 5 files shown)


maxteufel:
What is `.me`? I have never seen this before.

michaellilltokiwa:
This is a name of a feature in the respective test.

maxteufel (approved):

tests/reg_issue4248/reg_issue4248.fz.expected_err
```diff
@@ -6,7 +6,7 @@ The result type of this call depends on the target type.  Since the target type
 Called feature: 'e.me'
 Original result type: 'e.this'
 Type depending on target: 'e.this'
-Target type: 'p.this'
+Target type: 'p.me'
```
fridis, line 9: `p.me` is the (internal) type of the Function `me`, not a type that the user could actually use. And also, IMHO not the type that really is the root cause of this problem. 
The problem here is that for `p.type.x` the result type of `a.me` can be either `p` or `f`. This could be solved here by boxing, but it could not be solved if `me` would, e.g., return an `option` like `me => option e.this` where `option f` cannot be boxed to get `option p`.

fridis (commented):
I am not happy with this error message and I would like to see it improved. But I think that your change might make it more difficult to understand the problem. 

Also, it would be much better if we could find a rule that could be checked in the front end to flag this error earlier. 

I have not thought this through, but maybe something like the following would work: 

If the result type of a feature `a.b.c.d` is or uses `x.this` for any of the outer features `a`, `b`, or `c`, then the next inner feature  `b`, `c`, or `d`, respectively, must be marked as `fixed`.

PR #6952, opened 2026-04-16
concur: implement pinning a thread to a specific CPU core

maxteufel:
Like the implementation for specifying a scheduling policy and priority, this is limited to the C backend and only available on Linux for now.

### DIFF
--- a/include/posix.c
+++ b/include/posix.c
@@ -25,6 +25,9 @@ Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
  *---------------------------------------------------------------------*/
 
 #define _POSIX_C_SOURCE 200809L
+#ifdef __linux__
+#define _GNU_SOURCE
+#endif
 
 #ifdef GC_THREADS
 #define GC_DONT_INCLUDE_WINDOWS_H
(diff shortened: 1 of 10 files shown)


include/posix.c
```diff
@@ -582,6 +589,23 @@ int fzE_thread_setschedparam(void * thrd, int policy, int priority)
...
+ * Set the scheduling CPU affinity of a running thread.
+ */
+int fzE_thread_setaffinity(void * thrd, int core)
+{
+#ifdef __linux__
+  cpu_set_t cpuset;
+  CPU_ZERO(&cpuset);
+  CPU_SET(core, &cpuset);
+
+  return pthread_setaffinity_np(*(pthread_t *)thrd, sizeof(cpu_set_t), &cpuset);
```
michaellilltokiwa, line 602: What is the difference of this to `sched_setaffinity`?

maxteufel: the `sched_*` functions work on a `pid_t` thread id, the `pthread_*` functions work on `pthread_t`. We have the latter due to our use of pthread. (do note: `sched_setaffinity` is likewise a Linux thing even though it does not have the `_np` suffix.)

michaellilltokiwa (approved):

PR #6958, opened 2026-04-16
base: improve comments, rename type parameters, make type fully qualified

fridis:
Make the code clearer and avoids problem with implicit `container.this` outer type.

### DIFF
--- a/modules/base/src/container/Typed_Sequence.fz
+++ b/modules/base/src/container/Typed_Sequence.fz
@@ -226,7 +226,7 @@ public typed_zipper(# the result type of two values that were zipped
 public Typed_Value ref is
 
 
-  # Feed this value's type, the comulative value `e` and this value itself into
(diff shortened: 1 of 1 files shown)


modules/base/src/container/Typed_Sequence.fz
```diff
@@ -226,10 +226,10 @@ public typed_zipper(# the result type of two values that were zipped
 public Typed_Value ref is
 
 
-  # Feed this value and its type into `a.apply`.  Add the
-  # cumulative value `e` and return the resulting cumulative value.
+  # Feed this value's type, the comulative value `e` and this value itself into
```
michaellilltokiwa:
```suggestion
  # Feed this value's type, the cumulative value `e` and this value itself into
```

michaellilltokiwa (approved):

maxteufel (approved):

PR #6961, opened 2026-04-17
ast: cleanup remove unneeded `f_array_internal_array`

michaellilltokiwa:


fridis (approved):
nice. 

Surprising side-effect on the expected error output, but ok.

michaellilltokiwa:
@fridis Indeed.

PR #6962, opened 2026-04-17
concur/thread: set_affinity: specify CPU(s) by mask instead of ID

maxteufel:


modules/base/src/concur/thread.fz
```diff
@@ -74,9 +74,11 @@ module:public thread(thrd Thread) : property.equatable is
       error "fzE_thread_setschedparam error {e}"
 
 
-  public set_affinity(core i32) outcome unit
+  # pin a thread to the CPUs given by core_mask
+  #
+  public set_affinity(core_mask i32) outcome unit
```
michaellilltokiwa: So this is limited to 32 cpus? Also why i32 and not u32?

michaellilltokiwa (approved):
Perfect!

PR #6967, opened 2026-04-20
Makefile: use shells realpath instead of the make one

### LINKED ISSUE #6021: `make` called to create reference manual fails if `build/generated/doc` is missing
try this:
```
 > rm -rf build/generated/doc
 > make build/doc/reference_manual/fuzion_reference_manual.pdf 
mkdir -p build/generated/doc
cd . && git log modules/base/src/encodings/unicode/data.fz  | grep --extended-regexp "^Date:" | head | sed "s-Date:   -:UNICODE_VERSION: -g" | head -n1 > /unicode_version.adoc
/bin/sh: 1: cannot create /unicode_version.adoc: Permission denied
make: *** [Makefile:1230: build/generated/doc/unicode_version.adoc] Fehler 2
```
It looks as if the call to `$(realpath $(@D))` returns the empty string if `$(@D)` does not exist. 
A second call to `make` works: 
```
 > make build/doc/reference_manual/fuzion_reference_manual.pdf 
mkdir -p build/generated/doc
cd . && git log modules/base/src/encodings/unicode/data.fz  | grep --extended-regexp "^Date:" | head | sed "s-Date:   -:UNICODE_VERSION: -g" | head -n1 > /home/fridi/fuzion/fuzion/build/generated/doc/unicode_version.adoc
mkdir -p build/generated/doc
[...]
```

michaellilltokiwa:
fixes #6021

Here is what the AI writes:

> 1. $(realpath …) is a Make function, not a shell command
> It is evaluated by make itself, before the shell runs.
> It expects a Make variable or literal path, not shell expressions.
> 2. $(@D) is an automatic variable
> It only has a value inside a recipe during rule execution.
> Outside of a rule (or in the wrong context), it expands to empty.

### DIFF
--- a/docs.mk
+++ b/docs.mk
@@ -86,7 +86,7 @@ $(DOC_JAVA): $(JAVA_FILE_UTIL_VERSION) $(JAVA_FILE_FUIR_ANALYSIS_ABSTRACT_INTERP
 
 $(BUILD_DIR)/generated/doc/unicode_version.adoc:
 	mkdir -p $(@D)
-	cd $(FZ_SRC) && git log modules/base/src/encodings/unicode/data.fz  | grep --extended-regexp "^Date:" | head | sed "s-Date:   -:UNICODE_VERSION: -g" | head -n1 > $(realpath $(@D))/unicode_version.adoc
+	cd $(FZ_SRC) && git log modules/base/src/encodings/unicode/data.fz  | grep --extended-regexp "^Date:" | head | sed "s-Date:   -:UNICODE_VERSION: -g" | head -n1 > $$(realpath $(@D))/unicode_version.adoc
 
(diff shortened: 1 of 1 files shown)


maxteufel (dismissed):

michaellilltokiwa:
@maxteufel I'm open to _better_ or different suggestions.

maxteufel:
The idiomatic `Makefile` solution is to create a separate target for the `mkdir` command and then depend on it. The expansion of `realpath` by `make` does not work here because it is [expanded before `mkdir` is run](https://stackoverflow.com/a/65837502).

maxteufel (approved):
Nice! The directory target could also be added as a dependency for all other targets that have a `mkdir -p $(@D)`.

PR #6970, opened 2026-04-20
base: cleanup effects initialization

fridis:
Use explicit `effect_type.env.operation` instead of helper features that call `effect_name => effect_type.env`.  These helpers were only needed for default effect installation, which is no done automatically.

Remove calls of `effect.default`.

Restructure code for `memoize` to allow default effect not to depend on instance of user-defined outer feature.

Added new instrinsic `effect.type.instate_at_singularity`, renamed `replace0` as `set0` and allowed this to set an effect even if there is none instated yet. Removed `default0`.


tests/mod_lock_free_Map_threads/ctrie_threads.fz.effect
```diff
@@ -1,9 +1,5 @@
-((list i32).group_by#2 i32).lm
```
michaellilltokiwa, line 1: It is unclear to me why some effects are not listed anymore?

fridis: Some have a different name now (`io.file.*` instead of some heir of `io.file`), and some —like the example `lm` here— should never have been shown since these are local effects use for temporary mutability but without any state change visible to the caller.

michaellilltokiwa (approved):

PR #6980, opened 2026-04-21
tests: add reg test #1169

### LINKED ISSUE #1169: syntax for operator names in dot-notation is a little awkward
See this example:
```
a is
  nose is
    sneeze => say "achoo"
  infix +(o a) : nose => infix + .this.sneeze
a + a
```
The space between `+` and `.` is required. It is not possible to write
```
  infix +(o a) : nose => infix +.this.sneeze
```
or
```
  infix +(o a) : nose => (infix +).this.sneeze
```
or
```
  infix +(o a) : nose => infix (+).this.sneeze
```
We might want to allow one of the alternatives using `(`/`)`.

michaellilltokiwa:
One of the mentioned alternatives in the issue now works. Adding a regression test for this.

fixes #1169


maxteufel (approved):
Nice. Although I would note that now since for features that take arguments, something like `foo x y z .or_panic` works, the variant in the original issue with a space feels not that awkward anymore.

PR #6982, opened 2026-04-22
cleanup: rename containsUndefined->containsArtificialType

michaellilltokiwa:


fridis (changes requested):
I am not very happy with this name since it is not at all descriptive of what this means (same holds for `isArtificialType()` that is used in `isUndefined()`). 

Both `t_UNDEFINE` and `t_FORWARD_CYCLIC` are undefined types during type inference, the first for type parameters, the second for the result type in a recursive call (at least if I remember this correctly).  In a recent patch, I extended this to return `true` for `t_ERROR` as well since that suppresses some subsequent errors. 

If you find a better name, please go ahead, something like `containsTypeThatHasNotYetBeenInferred()` but shorter :-).

PR #6996, opened 2026-04-24
fuir: change `clazzInstantiatedHeirs` to return empty array for value clazzes

michaellilltokiwa:
Rationale is that we need heirs info only if clazz is Ref, otherwise the info is useless anyway.


maxteufel (approved):
nice cleanup.

PR #7000, opened 2026-04-24
http: implement chunked Transfer-Encoding

### LINKED ISSUE #6978: http: add support for reading chunked HTTP messages
<https://github.com/tokiwa-software/fuzion/blob/a336e040b28d6635344dc088e3bde53181035917/modules/http/src/http/Message.fz#L256-L258>
This not being implemented yet sometimes [causes errors in fzweb](https://github.com/tokiwa-software/fzweb/issues/407).

maxteufel:
Fixes #6978. This does not implement comments for the chunks, whitespaces between tokens, or trailers (headers after the end of content). All of those seem like rather questionable HTTP features.

Test with
```sh
fz -jvm -JLibraries="wolfssl" -modules=web -e 'say <| web.get4 "https://anglesharp.azurewebsites.net/Chunked"'
```

### DIFF
--- a/modules/http/src/http/Message.fz
+++ b/modules/http/src/http/Message.fz
@@ -221,6 +221,198 @@ body_reader(LM type : mutate, reader_size i32, _ unit) : io.Read_Handler is
             e error => e
 
 
+# version of body reader that handles connections with Transfer-Encoding:
+# chunked
+#
+# content with chunked transfer encoding arrives at the client in chunks of
+# data, where each chunk specifies a length (encoded in hex digits) and its
+# content, which must not exceed the given length. chunks are separated by CR
+# LF. upon receiving a chunk with length zero, the connection can be deemed
+# terminated.
+#
+# relevant RFC sections:
+# https://datatracker.ietf.org/doc/html/rfc2616#section-3.6.1
+# https://datatracker.ietf.org/doc/html/rfc7230#section-4.1
+# https://datatracker.ietf.org/doc/html/rfc9112#section-7.1
+#
+body_reader_chunked(LM type : mutate) : io.Read_Handler is
+
+  chunk_start is
+  chunk_size is
+  chunk_post_size is
+  chunk_end_start is
+  chunk_end_finalize is
+  chunk_body is
+  chunk_post_body_start is
+  chunk_post_body_finalize is
+  chunk_end_of_file is
+
+  chunk_enum : choice
+    chunk_start
+    chunk_size
+    chunk_post_size
+    chunk_end_start
+    chunk_end_finalize
+    chunk_body
+    chunk_post_body_start
+    chunk_post_body_finalize
+    chunk_end_of_file is
+
+  position := mut chunk_enum chunk_start
+  current_chunk_size := mut u64 0
+
+  update_current_chunk_size (new option u64) choice (Sequence u8) io.end_of_file error =>
+    match new
+      nil => error "overflow while parsing chunk size"
+      i u64 =>
+        current_chunk_size <- i
+        position <- chunk_size
+        Sequence u8 .empty
+
+  handle_start (s Sequence u8) choice (Sequence u8) io.end_of_file error =>
+    c := s.first.or_panic .as_u64
+
+    (io.buffered LM).reader.env.discard 1
+
+    current_chunk_size <- 0
+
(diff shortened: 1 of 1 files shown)


modules/http/src/http/Message.fz
```diff
@@ -255,7 +435,7 @@ module body_helper(LM type : mutate, is_req bool, header_fields container.Map St
 
       # If a Transfer-Encoding header field is present and the chunked transfer coding (Section 7.1) is the final encoding, the message body length is determined by reading and decoding the chunked data until the transfer coding indicates the data is complete.
       if transfer_encodings.last.or_panic = "chunked"
```
michaellilltokiwa, line 449: Is the content of the header field also normalized to lower_case?

modules/http/src/http/Message.fz
```diff
@@ -221,6 +221,186 @@ body_reader(LM type : mutate, reader_size i32, _ unit) : io.Read_Handler is
...
+  handle_end_start (s Sequence u8) choice (Sequence u8) io.end_of_file error =>
+    c := s.first.or_panic
+
+    (io.buffered LM).reader.env.discard 1
+
+    if c = 0x0d
+      position <- chunk_end_finalize
+      Sequence u8 .empty
+    else
+      error "NYI: trailers"
```
michaellilltokiwa: What exactly is missing here? Maybe a pointer to the part in the spec would be helpful. Also qualifier `UNDER DEVELOPMENT`

modules/http/src/http/Message.fz
```diff
@@ -221,6 +221,186 @@ body_reader(LM type : mutate, reader_size i32, _ unit) : io.Read_Handler is
...
+    chunk_end_finalize
+    chunk_body
+    chunk_post_body_start
+    chunk_post_body_finalize
+    chunk_end_of_file is
+
+  position := mut chunk_enum chunk_start
+  current_chunk_size := mut u64 0
+
+  update_ccs_or_err (new option u64) choice (Sequence u8) io.end_of_file error =>
```
michaellilltokiwa: unclear to me what ccs stands for?

modules/http/src/http/Message.fz
```diff
@@ -221,6 +221,186 @@ body_reader(LM type : mutate, reader_size i32, _ unit) : io.Read_Handler is
...
+          position <- chunk_body
+        else
+          position <- chunk_post_body_start
+      nil => panic "should not happen, underflow when calculating new chunk size"
+
+    chunk
+
+
+  handle_post_body_start (s Sequence u8) choice (Sequence u8) io.end_of_file error =>
+    c := s.first.or_panic
```
michaellilltokiwa, line 368: These two lines could be deduplicated, they are used a lot

michaellilltokiwa (approved):
Good stuff! Code is easy to read and to the point. Maybe a few (high-level) comments would help. I don't really know how chunked transfer works for example.

modules/http/src/http/Message.fz
maxteufel: We need to normalize this here, but we cannot assume that in general for all headers (cookies are case sensitive): <https://www.rfc-editor.org/rfc/rfc7230#section-4>.

modules/http/src/http/Message.fz
maxteufel: <https://www.rfc-editor.org/rfc/rfc2616#:~:text=The%20trailer%20allows,%20see%20section%2014.40).>

modules/http/src/http/Message.fz
maxteufel: Keeping this after discussion, as the savings from moving this to a separate feature are not that high (we still need `c := take_first_byte_and_discard` in every feature).

modules/http/src/http/Message.fz
maxteufel: done separately in #7012.

modules/http/src/http/Message.fz
maxteufel: `current_chunk_size`, clarified by renaming the feature as `update_current_chunk_size`.

maxteufel:
@michaellilltokiwa I've addressed your comments and also added a small explanation of chunked transfer encoding around `body_reader_chunked`.

PR #7002, opened 2026-04-27
ast: cleanup in `handDownListThroughInheritsCalls`

michaellilltokiwa:


fridis (approved):
Nice. Every `if` and every condition we remove helps to improve this complex code!

PR #7007, opened 2026-04-27
fuir: attempty to simplify `handDown`

michaellilltokiwa:
- merge the two `replace_inherited_this_type`
- replaced while-loop by recursion which seems much simpler in this case

### DIFF
--- a/src/dev/flang/fuir/Clazz.java
+++ b/src/dev/flang/fuir/Clazz.java
@@ -1891,19 +1891,15 @@ class Clazz extends ANY implements Comparable<Clazz>
    * that is to be chosen. NO_SELECT otherwise.
    *
    */
-  private AbstractType handDown(AbstractType t, int select, BiConsumer<AbstractType, AbstractType> foundRef, List<AbstractCall> inh0)
+  private AbstractType handDown(AbstractType t, int select, BiConsumer<AbstractType, AbstractType> foundRef, List<AbstractCall> inh)
   {
     if (PRECONDITIONS) require
       (t != null,
        Errors.any() || t != Types.t_ERROR,
        Errors.any() || (t.isOpenGeneric() == (select >= 0)));
 
-    for (AbstractCall c : inh0)
-      {
-        t = t.replace_inherited_this_type(c.calledFeature(), feature(), foundRef);
-      }
 
-    return handDown(t, select, foundRef, feature(), feature().isField());
+    return handDown(t, select, foundRef, inh, feature().isField());
   }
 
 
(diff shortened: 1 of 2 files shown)


src/dev/flang/fuir/Clazz.java
```diff
@@ -1891,59 +1891,52 @@ private Clazz handDown(AbstractType t, int select, List<AbstractCall> inh)
...
-    var child = this;
-    AbstractFeature parent = feature();
-    while (child != null)
+    return handDown(t, select, foundRef, inh, feature().isField());
+  }
+
+
+  // iterate using {@code child} and {@code parent} over outer clazzes starting at
+  // {@code this} where {@code child} is the current outer clazz and {@code parent} is the
+  // parent feature the previous inner clazz' feature was inherited from.
```
maxteufel: please adjust this comment, `child` and `parent` no longer exist in the code.

src/dev/flang/fuir/Clazz.java
```diff
@@ -1891,59 +1891,52 @@ private Clazz handDown(AbstractType t, int select, List<AbstractCall> inh)
...
+    // find outer that inherits this clazz, e.g.
+    //
+    //   Any.me =>
+    //     res := Any.this
+    //     res
+    //   x : Any is
+    //
+    // here, for {@code x.me.res} inherited from {@code Any.me.res}, the
+    // inheritance is two features out when {@code x} ({@code childf}) inherits
+    // form {@code Any} ({@code parent}).
```
maxteufel: same here

maxteufel (changes requested):

src/dev/flang/fuir/Clazz.java
```diff
@@ -1897,59 +1891,55 @@ private Clazz handDown(AbstractType t, int select, List<AbstractCall> inh)
...
+
+    // NYI: UNDER DEVELOPMENT: this code needs documentation / motivation
+    // particularly it is unclear to me how the order in which
+    // we do these things affects t
+
+    for (AbstractCall c : inh)
       {
         t = t.replace_inherited_this_type(c.calledFeature(), feature(), foundRef);
       }
+    var t1 = AbstractFeature.handDownThroughInheritsCalls(t, select, inh);
```
maxteufel, line 1927: i'm a bit vary about dropping the `inh != null` check, why can we assume this here?

michaellilltokiwa: I think, we only need to do this in the frontend, I believe the code was copied from there. Anything else would IMHO be a bug in the compiler.

michaellilltokiwa: btw I not only removed the null check but also `tryFindInheritanceChain` by `findInheritanceChain`

maxteufel (approved):

PR #7018, opened 2026-04-29
http: handle extensions for chunked transfer encoding

maxteufel:
Per the spec, clients that do not understand the chunk extensions must ignore them.

### DIFF
--- a/modules/http/src/http/Message.fz
+++ b/modules/http/src/http/Message.fz
@@ -239,6 +239,7 @@ body_reader_chunked(LM type : mutate) : io.Read_Handler is
 
   chunk_start is
   chunk_size is
+  chunk_extension is
   chunk_post_size is
   chunk_end_start is
   chunk_end_finalize is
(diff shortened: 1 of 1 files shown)


modules/http/src/http/Message.fz
```diff
@@ -258,8 +260,11 @@ body_reader_chunked(LM type : mutate) : io.Read_Handler is
     chunk_post_body_finalize
     chunk_end_of_file is
 
+  extension_limit := 1024.as_u64 * 64 # arbitrary value, see below
```
michaellilltokiwa:
```suggestion
  extension_limit := (u64 1024) * 64 # arbitrary value, see below
```

michaellilltokiwa: This results in one less cast in the end though one could argue that the optimizer should optimize this away in the end anyway.

michaellilltokiwa (approved):

PR #7030, opened 2026-04-30
base: Detect multi-thread use of mutate effect, add `container.Ring_Buffer`

fridis:
`mutate` now detects if it is used by multiple threads and flags an error (precondition safety fault) if that happens.  Added `mutate.exclusive` as a mechanism that may later be used for something like a `synchronized` block or a transaction in a transactional-memory implementation of `mutate`.

Added `container.Ring_Buffer`, a simple generic ring buffer implementation that I use as a test case for the work on multi-threaded `mutate` impementations.

Added `tests/ring_buf` that tests `container.Ring_Buffer`, but that also should become a test of multi-threaded mutate effects once they are available.

I had to add `concur.threads.env.current` to access the current thread to be able to detect that a `mutate` effect is used by several different threads.

Added `racy_mutate` as a temporary workaround to be used where a `mutate` is required by several threads. Added its use in tests that required this. 

Remove use of `mutate` or `mut` in `http.fum` and `webserver.fum`, used the mutate provided as a type parameter instead.


modules/webserver/src/webserver.fz
```diff
@@ -46,7 +46,7 @@ public Webserver(/*NYI: UNDER DEVELOPMENT: custom thread pool TP type : concur.t
           while is_running.read do
             accept_res := net.server.env.accept.bind unit conn->
 
-              lm : mutate is
```
michaellilltokiwa, line 49: I don't think it is necessary to mark this as `racy_mutate`. The instance is created in the thread accepting the connection, i.e. main thread, but it should only ever be used in one thread, the once that processes the connection.

tests/reg_issue6882/reg_issue6882.fz
```diff
@@ -39,5 +39,7 @@ reg_issue6882 =>
                    _ := lm.env
                  .catch _->
 
+            _ := lm.env
```
michaellilltokiwa: Why is this necessary?

modules/base/src/container/Ring_Buffer.fz
```diff
@@ -0,0 +1,110 @@
...
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature container.Ring_Buffer
+#
+# -----------------------------------------------------------------------
+
+
+# Ring_Buffer -- A simple fixed-size ring buffer
```
michaellilltokiwa, line 25: We already have something called `Circular_Buffer`. I think we should merge these two in the future.

modules/webserver/src/webserver.fz
fridis: This is required, otherwise `tests/mod_webserver` fails with 
```
0a1,76
> 
> error 1: FATAL FAULT `*** panic ***`: *** invalid mutate for 'Webserver.(λ()->).call.(λ()-> h := middlewares.reduce hdlr ((r,...).call.(λ()-> log "listening on port: $port").call.loop.(λconn-> lm : mutate is _ := conn.in_thr...).call.lm'
> Call stack:
> fuzion.sys.fatal_fault#2 at {base.fum}/fuzion/sys.fz:n
> fuzion.type.runtime.type.fault.type.default_value.(λkind_and_msg-> kind, msg := kind_and_m...).call#1 at {base.fum}/fuzion/runtime/fault.fz:n
> fuzion.runtime.fault.cause#1 at {base.fum}/eff/fallible.fz:n
> panic.type.default_value.(λ(msg->fuzion.runtime.fault.env.cause (...).call#1 at {base.fum}/panic.fz:n
> panic.cause#1 at {base.fum}/eff/fallible.fz:n
```

modules/base/src/container/Ring_Buffer.fz
fridis: Oh, I missed that. Yes, these should be merged eventually. I chose `Ring_Buffer` since it is a very simple structure to start with, but it requires all the basic thread-interactions like waiting on an empty buffer etc. 

So instead of trying to force a new multi-threaded mutate mechanism into any existing datatype, I would like to use `Ring_Buffer` as a template and later apply multi-threaded mutates, once they exist, to other data structures.

michaellilltokiwa (approved):

PR #7036, opened 2026-05-04
ast: Type.toString omit `.type` on humanReadable=true

michaellilltokiwa:
This aligns better to what actually has to written in Fuzion source code.

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -2455,7 +2455,7 @@ there is no common super type of the two types (Types.t_ERROR)
           {
             result = result + ".this";
           }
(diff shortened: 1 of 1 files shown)


fridis:
I am undecided about this, it is nice to output what is written in the source code, but it can be confusing since `xyz` and `xyz.type` are different types. We need to look at examples, I think it is ok to output `xyz.abc` instead of `xyz.type.abc` when `abc` is a type feature of `xyz`, but for plain types, just omitting '.type` is probably usually not a goot idea. Similar for outer types, would be good to print `pqr.stu.type` instead of `pqr.type.stu.type`, but just `pqr.stu` is a different type.

PR #7037, opened 2026-05-05
base: remove some old NYI

michaellilltokiwa:


modules/base/src/container/Set.fz
```diff
@@ -180,10 +180,6 @@ is
   # equality
   #
   public fixed redef type.equality(a, b Set.this) bool
-  # NYI: CLEANUP: precondition useless because `pre true` is inherited from property.equatable.type.equality
```
michaellilltokiwa, line 183: This can not be a precondition anyway.

maxteufel (approved):

PR #7038, opened 2026-05-05
base: remove `effect.type.return`

michaellilltokiwa:


modules/base/src/effect.fz
```diff
@@ -358,23 +358,6 @@ public effect is
   public abortable bool => true
 
 
-  # abort the current execution and return from the surrounding call to
```
michaellilltokiwa, line 361: I think we should stick with `abort` for now.

maxteufel (approved):

PR #7040, opened 2026-05-05
Modified Jitter to run on another thread with high priority.

Nimux2:
Modified Jitter to run on another thread with high priority.


examples/jitter/jitter.fz
```diff
@@ -150,4 +155,19 @@ jitter =>
       e error => say $e
       void    => # will not happen
 
-  test_jitter time.nano time.nano
+  concur.threads ! ()->
+    time.nano ! ()->
```
maxteufel: why does it need `time.nano` here when before it worked without?

examples/jitter/jitter.fz
```diff
@@ -150,4 +155,19 @@ jitter =>
...
+      if run_on_dedicated_thread
+        thrd := concur.threads.env.spawn ()->
+          test_jitter time.nano time.nano
+        match thrd.set_policy concur.scheduling_policy.sched_fifo false 99
+          e error => say "could not raise jitter thread priority: $e. Hint: error 38 means backend does not implement thread policy (try native backend with -c), error 1 means insufficient privileges (CAP_SYS_NICE/root on Linux)."
```
maxteufel: message too verbose

maxteufel (commented):
This is a start but I think this task encompasses more that we can already do:

We should keep track of the jitter/timing data in the main thread. Thread running the important code communicates back to the main thread, which stores information in some data structure.

That will then make it easy to add more separate threads doing other things, and using `set_affinity` on some of them.

Also I would like a comparison of the current state and the new state of this example.

examples/jitter/jitter.fz
Nimux2: I checked, and turns out it isn't actually necessary. I think it's a remnant of something I did last week, that I just forgot to remove since it doesn't cause errors.

examples/jitter/jitter.fz
```diff
@@ -150,4 +155,18 @@ jitter =>
       e error => say $e
       void    => # will not happen
 
-  test_jitter time.nano time.nano
+  concur.threads.type.default_value.or_panic ! ()->
```
maxteufel: you should not instate an effect from an instance of the same effect that has already been instated.

michaellilltokiwa: Yes it is not necessary to instate the effect manually since effect analysis will find the usage and instate the effect at big bang (provided the effect implements `type.default_value`).

But also it is not really wrong.

> you should not instate an effect from an instance of the same effect that has already been instated.

Another comment. If instantiation is done like this I would expect that big bang does _not_ instate concur.threads, not sure though. Thus there would not be an instated instance yet.
Even if big bang already instates `concur.threads` it would not be the same instance since `concur.threads.default_value` creates a fresh/new instance every time.

examples/jitter/jitter.fz
```diff
@@ -150,4 +155,17 @@ jitter =>
       e error => say $e
       void    => # will not happen
 
-  test_jitter time.nano time.nano
+  if run_on_dedicated_thread
+    thrd := concur.threads.env.spawn ()->
+      test_jitter time.nano time.nano
+    match thrd.set_policy concur.scheduling_policy.sched_fifo false 99
+      e error => say "jitter thread priority: $e. Error 38 : backend does not implement thread policy \n Error 1 : insufficient privileges."
```
michaellilltokiwa: `panic` is probably appropriate in this case.
```suggestion
      e error => panic "jitter thread priority: $e. Error 38 : backend does not implement thread policy \n Error 1 : insufficient privileges."
```

examples/jitter/jitter.fz
```diff
@@ -150,4 +155,17 @@ jitter =>
...
+      e error => panic "jitter thread priority: $e. Error 38 : backend does not implement thread policy \n Error 1 : insufficient privileges."
+      unit    => unit
+    _ := thrd.join
+  else
+    test_jitter time.nano time.nano
+
+        # How to run with high priority :
+        # FUZION_HOME=build bash bin/fz -c examples/jitter/jitter.fz
+        # sudo setcap cap_sys_nice=ep ./jitter
+        # ./jitter
```
maxteufel: indentation level of this should also be adjusted

maxteufel (approved):
Looks good.

PR #7041, opened 2026-05-05
base: add defer effect

maxteufel:
This can be used to postpone running some piece of code until after the effect has been deinstated.

### DIFF
--- a/modules/base/src/defer.fz
+++ b/modules/base/src/defer.fz
@@ -1,0 +1,32 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
Nice idea! But do you have any use-case for this?

modules/base/src/defer.fz
```diff
@@ -0,0 +1,32 @@
...
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature defer
+#
+# -----------------------------------------------------------------------
+
+
+# defer - postpone execution of a feature until the end of a piece of code
+#
+public defer (code () -> unit) : effect is
```
michaellilltokiwa: I think this should be done differently, since how it is currently done `defer` does not need to be an effect.

Rather than the effect taking a lambda it should have a feature to register any deferred actions.
`defer.env.add ...` then the deferred actions are run when effect is deinstated.

modules/base/src/defer.fz
```diff
@@ -0,0 +1,48 @@
...
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature defer
+#
+# -----------------------------------------------------------------------
+
+
+# defer - postpone execution of code until the end of a piece of code
```
michaellilltokiwa:
```suggestion
# defer - postpone execution of code until effect is deinstated
```

modules/base/src/defer.fz
```diff
@@ -0,0 +1,48 @@
...
+#
+private:public defer (code_to_run list ()->unit) : effect is
+
+  # add code to effect
+  #
+  # no specific ordering of running these is guaranteed on effect finalization
```
michaellilltokiwa, line 31: I think it probably makes most sense to use LIFO ordering, using a stack.

maxteufel: this is what happens currently using `list`, i still think we should not give a guarantee to that to the caller

michaellilltokiwa: Okay, we change later if need be.

michaellilltokiwa (approved):

PR #7047, opened 2026-05-07
fe: change redef type par rule

### LINKED ISSUE #7046: No error raised for potentially incompatible type parameter contraint
This compiles without error:
```
f is
  a (T type : codepoint) unit =>
g : f is
  redef a(T type : String) unit =>
```

michaellilltokiwa:
fixes #7046

I don't think it should be legal to relax type par constraints as is. I strongly suspect this to be a soundness issue.
What would probably be okay is to change the check from `constraintAssignableFrom` to `constraintAssignableFromWithoutTaggingAndBoxing`

### DIFF
--- a/src/dev/flang/fe/SourceModule.java
+++ b/src/dev/flang/fe/SourceModule.java
@@ -1715,10 +1715,10 @@ An xref:fuzion_opentypeparameter[open type parameter] argument to a feature that
                         (argo.isOpenTypeParameter() != argr.isOpenTypeParameter()                ) ||
             /*
     // tag::fuzion_rule_REDEF_TYPE_CONSTRAINTS[]
-A xref:fuzion_type_constraint[type constraint] of a xref:fuzion_typeparameter[type parameter] must be redefined using a xref:fuzion_type_constraint[type constraint] that is xref:fuzion_constraint_assignable[constraint assignable] from the original  xref:fuzion_typeparameter[type parameter]'s  xref:fuzion_type_constraint[type constraint].
+A xref:fuzion_type_constraint[type constraint] of a xref:fuzion_typeparameter[type parameter] must be redefined using a xref:fuzion_type_constraint[type constraint] that is equal to the original xref:fuzion_typeparameter[type parameter]'s  xref:fuzion_type_constraint[type constraint].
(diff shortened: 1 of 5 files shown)


maxteufel:
Is there anything that makes you believe this is wrong? I see no problem with *using* this at least.

michaellilltokiwa:
My thoughts were along these lines:
```
A ref is
  a(T type : codepoint, cp T) =>
    say cp

B : A is
  redef a(T type : String, cp T) =>
    say "redefed $cp"

a := A
b := B
arr array A := [a, b]
arr[0].a "a"
arr[1].a "a"
a.a "a"
b.a "ab"
```
where `arr[1].a "a"` requires no boxing of "a" but `b.a "ab"` requires boxing.

maxteufel:
I don't see a problem with the given example. The only case I would wonder about is `arr[1].a "ab"`, but this is handled correctly:
```
$ fz ex7047.fz

ex7047.fz:13:8: error 1: Incompatible type parameter
arr[1].a "ab"

formal type parameter 'T : codepoint'
actual type parameter 'String'

one error.
```

fridis:
I agree with @maxteufel, see my comments to #7046. The problematic case is if the redefinition would be more restrictive, but this produces an error as follows: 
```

/home/fridi/fuzion/fuzion/test7046.fz:20:11: error 1: Wrong type parameter constraint in redefined feature
  redef a(T type : codepoint, cp T) =>

In 'B.a' that redefines 'A.a'
type parameter constraint is       : 'codepoint'
type parameter constraint should be: 'String'

Original argument declared at /home/fridi/fuzion/fuzion/test7046.fz:16:5:
  a(T type : String, cp T) =>

To solve this, change constraint of type parameter to 'String' at /home/fridi/fuzion/fuzion/test7046.fz:20:11:
  redef a(T type : codepoint, cp T) =>
```
Only the wording in this error message is stricter than needed: `constraint should be: 'String'` should better be  `constraint should be constraint assignable to 'String'` and the reference manual should have a definition of constraint assignable that matches `AbstractType.constraintAssignableFrom`.

fridis (commented):
See comment above.

PR #7058, opened 2026-05-08
merge `Ring_Buffer` / `Circular_Buffer`

michaellilltokiwa:
I decide to basically use `Ring_Buffer` and remove `Circular_Buffer` since `Ring_Buffer` uses the new exclusive mechanisms. They are/were pretty similar anyway.


modules/base/src/container/Ring_Buffer.fz
```diff
@@ -187,3 +187,9 @@ is
   =>
     M.env.exclusive ()->
       used
+
+
+  # amount of elements free for adding
+  #
+  public available i64 =>
+    size - count
```
maxteufel: re discussion: open question: should this be in M.env.exclusive?

modules/base/src/io/buffered/writer.fz
```diff
@@ -62,11 +62,11 @@ is
...
   public flush outcome unit =>
-    if buffer.buffered = 0
-      unit
+    for res := outcome unit, wh.write [b.or_panic]
+        b := buffer.poll
+    while b.ok
     else
-      n := buffer.buffered # we are expecting to write this number of bytes
-      wh.write (buffer.flush n)
+      res
```
maxteufel, line 70: this changes semantics, doesn't it? this might keep polling if there is something filling up the buffer quickly enough, that didn't happen before?

michaellilltokiwa: Hmm, yes. But isn't this effectively single-threaded anyway? So there should be no way that something is filling up the buffer.

modules/base/src/container/Ring_Buffer.fz
michaellilltokiwa: @fridis What is your opinion on this?

modules/base/src/io/buffered/writer.fz
maxteufel: That may be true but I don't like changing the semantics in a way like this in a pull request like this.

michaellilltokiwa: @maxteufel Can you explain why the semantics need to stay like they are? I don't see a convincing reason currently.

maxteufel: I don't per se oppose changing the semantics here. But I'd prefer doing that in a separate pull request or at least commit. That way, it is very clear that the semantics have changed when looking at the logs of finding a potential bug introduced due to that change.

michaellilltokiwa: Okay, makes sense.

modules/base/src/container/Ring_Buffer.fz
fridis: `M.env.exclusive` currently cannot be nested, so this is not possible (but I think I will change exclusive to permit nesting). 

Since `size` is immutable and `count` is already exclusive, there is no need for `available` to use `exclusive`.

modules/base/src/container/Ring_Buffer.fz
```diff
@@ -187,3 +187,9 @@ is
   =>
     M.env.exclusive ()->
       used
+
+
+  # amount of elements free for adding
+  #
+  public available i64 =>
```
fridis: My first intuition for `available` was that this must be the number of elements that can be `take`n without blocking. Maybe `unused`, `free`, `empty_slots`, or similar would be better?

fridis (changes requested):
This PR came faster than I expected, but I guess `Ring_Buffer` is stable enough to be used.

maxteufel (approved):

PR #7061, opened 2026-05-08
be/int: add workaround for current CI test failures

michaellilltokiwa:
[ci skip]

test run: https://github.com/michaellilltokiwa/fuzion/actions/runs/25558458158/job/75023433917

### DIFF
--- a/src/dev/flang/be/interpreter/Layout.java
+++ b/src/dev/flang/be/interpreter/Layout.java
@@ -172,7 +172,7 @@ class Layout extends FUIRContext
         case c_f64  -> 2;
         case c_void -> 0;
         // NYI: BUG: this does not work yet:
-        // FUIR should ne present us fields that
+        // FUIR should not present us with fields that
         // effectively store unit values
         // default     -> get(fc).size();
         default     -> get(fc).size() == 0 ? 1 : get(fc).size();
--- a/src/dev/flang/be/interpreter/Layout.java
+++ b/src/dev/flang/be/interpreter/Layout.java
@@ -171,7 +171,11 @@ class Layout extends FUIRContext
         case c_f32  -> 1;
         case c_f64  -> 2;
         case c_void -> 0;
-        default     -> get(fc).size();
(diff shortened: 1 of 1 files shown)


src/dev/flang/be/interpreter/Layout.java
```diff
@@ -171,7 +171,11 @@ private int fieldSize(int f)
         case c_f32  -> 1;
         case c_f64  -> 2;
         case c_void -> 0;
-        default     -> get(fc).size();
+        // NYI: BUG: this does not work yet:
+        // FUIR should ne present us fields that
```
fridis:
```suggestion
        // FUIR should not present us fields that
```

src/dev/flang/be/interpreter/Layout.java
```diff
@@ -171,7 +171,11 @@ private int fieldSize(int f)
...
+        // effectively store unit values
+        // default     -> get(fc).size();
+        default     -> get(fc).size() == 0 ? 1 : get(fc).size();
```
fridis, line 178: I think the backend should be able to cope with fields of size zero. The question if FUIR should remove accesses to unit type fields or not is IMHO independent of this, there are all sorts of optimizations or code transformations that may result in fields having no data, so I think the backend should best be able to handle this. 

BTW, further up, there is the case for `fuir().clazzIsUnitType(fc)`, which returns `0`. Why does `0` not work here?

src/dev/flang/be/interpreter/Layout.java
maxteufel: should ne? what does that mean?

michaellilltokiwa:
```suggestion
        // FUIR should not present us with fields that
```

michaellilltokiwa: @maxteufel I fixed the comment

maxteufel (approved):

fridis (dismissed):

PR #7063, opened 2026-05-08
base: fix state effect does not work when instated using `infix !`

### LINKED ISSUE #5634: `state` effect does not work when instated using `infix !`

michaellilltokiwa:
fixes #5634

### DIFF
--- a/examples/jitter/jitter.fz
+++ b/examples/jitter/jitter.fz
@@ -105,8 +105,7 @@ jitter =>
       # time for next release used by state effect
       next_release(t time.instant) is
 
-      # NYI: BUG: #5634: use `infix !` in this example once #5634 is fixed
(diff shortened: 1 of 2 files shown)


fridis (approved):
nice. 

At some point, we probably should remove all the code related to `oneway_monad` since all this is there fore is to show that effects and monads are interchangable.

PR #7064, opened 2026-05-08
be/int: add skip file for slow test, priority queue

michaellilltokiwa:
[ci skip]

test does not seem to complete within 10min


fridis (changes requested):
Can you add an issue to reduce the test complexity and add a reference to that issue to `skip_int`?

fridis (approved):

PR #7067, opened 2026-05-08
base: add `blocking_mutate`, a variant of `mutate` with blocking `exclusive`

fridis:
This uses `concur.sync.mutex` for synchnronization.  Extended `Ring_Buffer` to use exclusive also for single variable reads since this is required for proper synchronization.

Extended `tests/ring_buf` by actual concurrent accesses using busy wait loops.


modules/base/src/mutate/array2.fz
```diff
@@ -45,9 +45,20 @@ pre
   safety: length0 * length1 ≤ data.length.as_i64
 is
 
+
+  # A mutable array might be inaccessible if we are running in a thread that is
+  # not permitted be the underlying mutate instance to access its mutable elements
```
michaellilltokiwa:
```suggestion
  # not permitted by the underlying mutate instance to access its mutable elements
```

modules/base/src/mutate/array.fz
```diff
@@ -45,10 +45,20 @@ pre
 is
 
 
+  # A mutable array might be inaccessible if we are running in a thread that is
+  # not permitted be the underlying mutate instance to access its mutable elements
```
michaellilltokiwa:
```suggestion
  # not permitted by the underlying mutate instance to access its mutable elements
```

modules/base/src/mutate/array3.fz
```diff
@@ -43,9 +43,20 @@ pre
   safety: length0 * length1 * length2 ≤ data.length.as_i64
 is
 
+
+  # A mutable array might be inaccessible if we are running in a thread that is
+  # not permitted be the underlying mutate instance to access its mutable elements
```
michaellilltokiwa:
```suggestion
  # not permitted by the underlying mutate instance to access its mutable elements
```

modules/base/src/mutate.fz
```diff
@@ -90,11 +91,24 @@ public mutate : linear_effect is
                                                              else nil
 
 
-  # test if the current thread is the same as the thread this mutate was created in
+  # thread that currently has exclusive acces to this mutate. This is the creating thread
```
michaellilltokiwa:
```suggestion
  # thread that currently has exclusive access to this mutate. This is the creating thread
```

modules/base/src/mutate.fz
```diff
@@ -90,11 +91,24 @@ public mutate : linear_effect is
...
+  # for a mutate that is not sharable by threads. Other wise, this is `nil` if no thread
```
michaellilltokiwa:
```suggestion
  # for a mutate that is not sharable by threads. Otherwise, this is `nil` if no thread
```

modules/base/src/concur/blocking_mutate.fz
```diff
@@ -0,0 +1,84 @@
...
+  =>
+    match M.get_if_instated
+      x blocking_mutate.this => id = x.id
+      nil                    => false
+
+
+  # cleanup feature to destroy mutex and other resources associated with this
+  # effect.
+  #
+  # NYI: UNDER DEVELOPMENT: Ensure that this does not get called or redefined
```
michaellilltokiwa: I think we already ensure the first part:
```java
  public static void mustNotCallEffectFinally(Call call)
  {
    error(call.pos(), "Must not call " + ss("<effect>.finally") + ".",
      ss("<effect>.finally") + " is called automatically.");
  }
```

```suggestion
  # NYI: UNDER DEVELOPMENT: Ensure that this does not get redefined
```

fridis: I was not aware of this. Would be nice to have a different solution that works without special handling in the compiler, One would be to declare
```
  module:public base_only is
```
and then add an argument
```
  public finally(_ base_only) unit =>
```
We could even keep the original 
```
  public finally unit =>
```
to make it easier to have `finally` added to user defined effects.

fridis: Added #7068 for this.

michaellilltokiwa (approved):

PR #7069, opened 2026-05-11
Modules use `infix >>?` instead of `match ... nil => false`

fridis:
I think this makes the code easier to read.


michaellilltokiwa (approved):
Looks better to me too

PR #7070, opened 2026-05-11
modules: First version of `mutate.wait`

fridis:
The idea is as follows: `mutate.wait` receives a condition. If that condition does not hold, it will wait until that condition holds after an `exclusive` section.  Unlike Java's `wait`/`notify`, the condition is not checked by the waiting thread, but by the thread that performed the change. So there are fewer unnecessary wake-ups and thread switches.

`exclusive_when` was added as a convenience feature to enter an exclusive section only after waiting for a given condition.

What is currently missing is some sort of order, the threads are woken up in the order they have started waiting, thread priorities are not taken into account, but they should be.

Also missing is support for nested exclusive sections for the same `blocking_mutate` instance. This is currently does not produce a proper error, but it most likely does not work either. 

Using these new features, I extended `Ring_Buffer` with `take` and `put`, which are blocking variants of `poll` and `add`.


michaellilltokiwa:
I think main needs to be merged first, then test rerecorded. At least I fell into this trap countless of times.

modules/base/src/concur/blocking_mutate.fz
```diff
@@ -59,16 +63,43 @@ public blocking_mutate : mutate is
       res := code()
       set exclusive_thread := nil
       blocking_mutate.this.replace
+
+      match waiting.first
+        f concur.thread =>
+          for t := f, nxt
+              nxt := t.next
+          while !t.park_condition()
+          until nxt = f  # we eant once around the linked list and found norhing
```
```suggestion
          until nxt = f  # we went once around the linked list but found nothing
```

modules/base/src/concur/thread.fz
```diff
@@ -105,3 +107,127 @@ is
...
+  # is parked before it is released.
+  #
+  module park(F type : ()->unit, condition ()->bool, code F)
+  =>
+    cond.synchronized ()->
+      code()
+      set park_condition := condition
+      set parked := true
+      while parked   # perform this in a while to catch spurious wakeups
+        _ := cond.wait      # NYI: CLEANUP: Why is there a `bool` result for `wait`? Remove!
```
michaellilltokiwa, line 137: bool result is because waiting may fail, e.g. in java `Condition.await() throws InterruptedException`

modules/base/src/mutate.fz
```diff
@@ -99,12 +99,14 @@ public mutate : linear_effect is
...
-  # This is usually the case for the same thread this mutate was created in.  Furthermore,
-  # child features of mutate may provide support for multi-threaded access to mutable data,
-  # so they may permit other threads as well.
+  # check if the current thread is permitted to access mutable state managed by mutate.this
+  # exclusively.
   #
-  module is_correct_thread
+  # For a single-thread mutate, this is usually the case for the same thread this mutate was
+  # created in.  Furthermore, multi-threaded mutate implementations may ensure exclusive
+  # access to other threads via redefining the `exclusive` featurrs.
```
michaellilltokiwa:
```suggestion
  # access to other threads via redefining the `exclusive` features.
```

modules/base/src/mutate.fz
```diff
@@ -99,12 +99,14 @@ public mutate : linear_effect is
...
-  # test if the current thread is permitted to access mutable state managed by mutate.this.
-  # This is usually the case for the same thread this mutate was created in.  Furthermore,
-  # child features of mutate may provide support for multi-threaded access to mutable data,
-  # so they may permit other threads as well.
+  # check if the current thread is permitted to access mutable state managed by mutate.this
+  # exclusively.
   #
-  module is_correct_thread
+  # For a single-thread mutate, this is usually the case for the same thread this mutate was
+  # created in.  Furthermore, multi-threaded mutate implementations may ensure exclusive
```
michaellilltokiwa:
```suggestion
  # created in.  Furthermore, multi-thread mutate implementations may ensure exclusive
```

modules/base/src/mutate.fz
```diff
@@ -247,6 +249,39 @@ public mutate : linear_effect is
     code()
 
 
+  # wait for the given condition to becomes true by a mutation of the underlying data.
```
michaellilltokiwa:
```suggestion
  # wait for the given condition to become true by a mutation of the underlying data.
```

modules/base/src/mutate.fz
```diff
@@ -247,6 +249,39 @@ public mutate : linear_effect is
...
+  #
+  # In case this mutate is not multi-threaded and `condition()` is `false`, this will
```
michaellilltokiwa: wording: _is not multi-threaded_ vs- _is not multi-thread_?

modules/base/src/concur/thread.fz
```diff
@@ -105,3 +107,127 @@ is
...
+  #
+  module park_condition ()->bool := ()->true
+
+
+# A list of threads that are parked, linked for `next`/`prev`.
+#
+# This is currently a simple cyclic list. It might make sense to sort this by
+# priority and to add some user payload.
+#
+module Thread_List ref is
```
michaellilltokiwa: I think this should be in its own file.

michaellilltokiwa (approved):

maxteufel (approved):

modules/base/src/concur/thread.fz
fridis: Fundamental functions like `wait` and `signal` IMHO have to just work, we cannot reasonably handle an error here. 

But what we should have at some point is a way to interrupt these with a signal like `SIG_INT` and jump to a surrounding handler effect executing all `finally`s on the way. 

BTW, there are versions of `wait` and similar functions in `feeze/src/java/dev/flang/util/Threads.java`. Maybe we should copy this file to the Fuzion repo. 

I have added #7075 for this.

modules/base/src/concur/thread.fz
```diff
@@ -105,3 +107,65 @@ is
...
+
+  # is this thread parked?
+  #
+  # this is modified only in `park` and `unpark` while protected by `cond`.
+  #
+  parked := false
+
+
+  # park this thread atomically after storing `condition` in `park_condition` and
+  # execuding `code`.
```
maxteufel:
```suggestion
  # executing `code`.
```

modules/base/src/concur/thread.fz
```diff
@@ -105,3 +107,65 @@ is
...
+  # threads that are parked may use these link fields to form a `Thread_List` of
+  # parked threads.  Since a thread can never be parked more than once, these links
+  # can be used exclusively by the mechanism using park.
+  #
+  module prev thread := thread.this
+  module next thread := thread.this
+
+
+  # for a parked this, this holds a condition that must hold for the
+  # thread to be unparked.
```
maxteufel: for a parked thread, this [...]?

for a parked `thread.this`, this [...]?

not sure what is meant here

PR #7074, opened 2026-05-12
net/connection: instantiate mutate effect inside thread

maxteufel:
We cannot pass an instance of mutate from outside anymore since cross-thread mutate is no longer allowed.

N.B.: instatiate is the right word here, because thread assignment to a mutate happens at time of instatiation, not instation!

### DIFF
--- a/modules/webserver/src/webserver.fz
+++ b/modules/webserver/src/webserver.fz
@@ -46,22 +46,20 @@ public Webserver(/*NYI: UNDER DEVELOPMENT: custom thread pool TP type : concur.t
           while is_running.read do
             accept_res := net.server.env.accept.bind unit conn->
 
-              lm : racy_mutate is
-
-              _ := conn.in_thread_pool unit concur.thread_pool lm lm ()->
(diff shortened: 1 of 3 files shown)


modules/webserver/src/webserver.fz
```diff
@@ -46,22 +46,20 @@ public Webserver(/*NYI: UNDER DEVELOPMENT: custom thread pool TP type : concur.t
           while is_running.read do
             accept_res := net.server.env.accept.bind unit conn->
 
-              lm : racy_mutate is
-
-              _ := conn.in_thread_pool unit concur.thread_pool lm lm ()->
+              _ := conn.in_thread_pool unit concur.thread_pool racy_mutate ()->
```
michaellilltokiwa: I think we can just use the normal `mutate` here now, since the instance is created inside of the thread?

maxteufel: needed to do this to fix a failing test anyway.

michaellilltokiwa (approved):

PR #7079, opened 2026-05-13
fe: raise error if inferred lambda result type, refers to itself

### LINKED ISSUE #4428: AoC: type inference failure for cotype failed in case of inferred result type from lambda being inner type of lambda
I ran into this during [6 dec 24 Advent of Code](https://github.com/fridis/fuzion_aoc/tree/main/2024/06):

This code
```
f(x ()->T) =>
f ()->
  a is
    x => a
  a.x
```
results in 
```
> ~/fuzion/work/build/bin/fz type_infer_error5.fz 

/home/fridi/fuzion/advent_of_code/fuzion_aoc/2024/06/type_infer_error5.fz:2:3: error 1: Failed to infer actual type parameters
f ()->
  a is
    x => a
  a.x

In call to 'Function.type', no actual type parameters are given and inference of the type parameters failed.
Expected type parameters: 'THIS#TYPE, R, A...'
Type inference failed for one type parameter 'R'

one error.
```
which is strange since the error complains about the cotype's type parameter. Would be better to complain about the original `Function` type's type parameter or --even better-- to just infer the type correctly even though the outer type is a generated lambda. 

A related problem occurs with the following code
```
f(x ()->T) =>
f ()->
  a is
    x => a
  a
```
which complains that argument type when calling `f` is wrong:
```
 > ~/fuzion/work/build/bin/fz type_infer_error6.fz 

/home/fridi/fuzion/advent_of_code/fuzion_aoc/2024/06/type_infer_error6.fz:2:3: error 1: Incompatible types when passing argument in a call
f ()->
  a is
    x => a
  a

Actual type for argument #1 'x' does not match expected type.
In call to          : 'f'
expected formal type: 'Function λ.this.call.this.a'
actual type found   : 'universe.this.λ'
assignable to       : 'Any',
                      'Function universe.this.λ.call.this.a',
                      'universe.this.λ'
for value assigned  : 
()->
  a is
    x => a
  a

To solve this, you could change the type of the target 'x' to 'universe.this.λ' or convert the type of the assigned value to 'Function λ.this.call.this.a'.

one error.
```
In both cases, the underlying problem seems to be that the inferred type for `T` depends on the type of the feature created for the lambda, which makes this an unusable type outside of the lambda. We should either handle this case in a way that it does not produce these errors, or produce an error message that complains about this situation and explains the problem better.

michaellilltokiwa:
fixes #4428

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -2167,6 +2167,15 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
   }
 
 
+  /**
+   * return true if this == f or if this is an inner feature of f.
+   */
+  public boolean isSelfOrInnerFeatureOf(Feature f)
+  {
+    return this == f || outer() != null && outer().isSelfOrInnerFeatureOf(f);
+  }
+
+
 
 
 }
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2669,6 +2669,11 @@ public class AstErrors extends ANY
     );
   }
 
+  public static void inferredResultTypeOfLambdaIllegal(Function fun, AbstractType t)
+  {
+    error(fun.pos(), "Result type of lambda must not refer to itself.", "The inferred result type of the lambda: " + s(t));
+  }
+
 }
 
 /* end of file */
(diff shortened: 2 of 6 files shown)


maxteufel (approved):
The error is better but I agree with @fridis saying in the original issue that this should just work.

michaellilltokiwa:
@maxteufel Which PR are you referring to?

michaellilltokiwa:
superseded

PR #7087, opened 2026-05-19
Modified Jitter so that testing threads send raw data back to the main thread.

Nimux2:
Changes in this branch:

- Worker/main separation: the worker thread only collects raw timing samples and returns them to the caller via a `concur.Channel`. The main thread consumes those lists and builds the histograms. This decouples measurement from analysis and simplifies future multi-threaded extensions.
- Dedicated-thread path: the example can run the test on a spawned worker thread (default) or on the main thread.
- CPU affinity: the worker tries to pin itself to CPU 0. Success or failure is logged as part of the run output.
- Real-time scheduling (SCHED_FIFO) is implemented similarly to CPU affinity, and can be disabled if necessary.
- Precision note: This version is significantly more accurate/efficient than the original (thread-less).


examples/jitter/jitter.fz
```diff
@@ -68,101 +68,165 @@ jitter =>
...
+           for p := period_default, p.times 0.5 do
+             run_for_period p
 
-      e error => say $e
-      void    => # will not happen
+        _ error => results.get.reverse_list
+        void    => results.get.reverse_list
+
+  # offer a dedicated-thread run option
+  run_on_dedicated_thread => true
```
maxteufel: just throw out this option and make a separate `jitter_threaded.fz`. this will keep the code cleaner.

maxteufel (changes requested):

examples/jitter/jitter.fz
```diff
@@ -110,7 +105,8 @@ jitter =>
       # time for next release used by state effect
       next_release(t time.instant) is
 
-      state (next_release start) ! ()->
+      # NYI: BUG: #5634: use `infix !` in this example once #5634 is fixed
+      state _ _ (next_release start) ()->
```
maxteufel: that's not in the latest version of `jitter.fz`, this works by now

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,208 @@
...
+        period := samples_arr[i].values.0
+        period_samples := samples_arr[i].values.1
+        cycle_samples := samples_arr[i].values.2
+        say "testing period $period for $test_duration..."
+        h  := time.histogram.new "period jitter $period, log count / linear time" period period*2 1 0
+        h2 := time.histogram.new "cycle jitter $period, log count / log time"     nil    nil      1 0
+        ps_arr := period_samples.as_array
+        for j := 0, j+1 while j < ps_arr.length do h.add ps_arr[j]
+        cs_arr := cycle_samples.as_array
+        for j := 0, j+1 while j < cs_arr.length do h2.add cs_arr[j]
```
maxteufel: this could be done in a single loop

maxteufel (changes requested):

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,209 @@
...
+        period_samples := samples_arr[i].values.1
+        cycle_samples := samples_arr[i].values.2
+        say "testing period $period for $test_duration..."
+        h  := time.histogram.new "period jitter $period, log count / linear time" period period*2 1 0
+        h2 := time.histogram.new "cycle jitter $period, log count / log time"     nil    nil      1 0
+        ps_arr := period_samples.as_array
+        cs_arr := cycle_samples.as_array
+        for j := 0, j+1 while j < ps_arr.length do
+          h.add ps_arr[j]
+          h2.add cs_arr[j]
```
maxteufel: yes! but `ps_arr` and `cs_arr` are just taken from `samples_arr` which is an array of tuples!

it will be easier to read if you iterate over the elements of that:

```
for
  sample in samples
  # write code here
do
  # the loop content here
```

maxteufel (changes requested):

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,208 @@
...
+        period_samples := sample.values.1
+        cycle_samples := sample.values.2
+        say "testing period $period for $test_duration..."
+        h  := time.histogram.new "period jitter $period, log count / linear time" period period*2 1 0
+        h2 := time.histogram.new "cycle jitter $period, log count / log time"     nil    nil      1 0
+        ps_arr := period_samples.as_array
+        cs_arr := cycle_samples.as_array
+        for j := 0, j+1 while j < ps_arr.length do
+          h.add ps_arr[j]
+          h2.add cs_arr[j]
```
maxteufel: you can still do

```
for
  ps in period_samples
  cs in cycle_samples
do
  h.add # ...
```

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,208 @@
...
+        say "(Hint: run with sudo or use setcap cap_sys_nice=ep ./jitter)"
+
+  # wait for thread to send results
+  match result_ch.next
+    samples list =>
+      say "received samples"
+      for sample in samples do
+        period := sample.values.0
+        period_samples := sample.values.1
+        cycle_samples := sample.values.2
```
maxteufel: `period, period_samples, cycle_samples := sample.values` should work here

maxteufel (changes requested):

PR #7092, opened 2026-05-19
list: redef finite to yes if cycled list is empty

### LINKED ISSUE #7090: cycled empty `Sequence` should be finite
Try this
```
 > ./build//bin/fz -e 'say (Sequence unit).empty.cycle.count' 
0
```
so this sequence is empty, which is definitely finite, but:
```
 > ./build//bin/fz -e 'say (Sequence unit).empty.cycle.finite' 
no
```

maxteufel:


simonvonhackewitz (approved):
Does it need a regression test?

modules/base/src/list.fz
```diff
@@ -572,7 +572,7 @@ is
   #
   public redef cycle Sequence A =>
     _ : Sequence A is
-      public redef finite trit => trit.no
+      public redef finite trit => cycle_list_finite
```
michaellilltokiwa: How about?
```suggestion
      public redef finite trit => is_empty ? trit.yes : trit.no
```

maxteufel: it would need to be:
```suggestion
      public redef finite trit => list.this.is_empty ? trit.yes : trit.no
```
due to ambiguity with the anonymous Sequence. do you still think this is more readable than my suggestion?

michaellilltokiwa:
I currently think we should add:
```
  public bool.as_trit =>
    if bool.this
      trit.yes
    else
      trit.no
```
Then the code would be `list.this.is_empty.as_trit`

michaellilltokiwa (approved):

modules/base/src/Sequence.fz
```diff
@@ -480,7 +480,7 @@ is
   # is empty, returns 'nil'
   #
   public cycle Sequence T
-   post debug: result.finite.is_no
+   post debug: is_empty && result.finite.is_yes || result.finite.is_no
```
fridis, line 483:
```suggestion
   post debug: is_empty && result.finite.is_yes || !is_empty && result.finite.is_no
```

maxteufel: addressed in #7105

PR #7099, opened 2026-05-20
concur/thread: keep track of scheduling policy and priority

maxteufel:


modules/base/src/concur/thread.fz
```diff
@@ -63,17 +73,17 @@ is
   # priority only applies to some scheduling policies and is
   # currently limited to a value between 0 and 99.
   #
-  public set_policy(policy choice
+  public set_policy(policy0 choice
```
michaellilltokiwa, line 76: I think as is this is not thread safe. Imagine two threads calling `set_policy` at the same time. Maybe just put this whole code block in `cond.synchronized ()->`

modules/base/src/concur/thread.fz
fridis, line 128: This should also be `cond.synchronized`.

modules/base/src/concur/thread.fz
```diff
@@ -35,6 +35,16 @@ is
...
 
+  public policy choice
+    concur.scheduling_policy.sched_other
+    concur.scheduling_policy.sched_fifo
+    concur.scheduling_policy.sched_rr
+  :=
+     concur.scheduling_policy.sched_other
+
+
+  public priority i32 := 0
```
fridis: These should both be non-public since reading a field that may be changed by another thread may result in outdated (or even out-of-thin-air) values. 

Instead, there should be one function that returns a tuple of these two and that is made atomic using `cond.synchronized`.

fridis (approved):

modules/base/src/concur/thread.fz
maxteufel: addressed in #7106 since it is a feature untouched by this pull request.

fridis (approved):

PR #7102, opened 2026-05-20
rt: make posix.c compile without systemtap installed

### LINKED ISSUE #7101: posix.c does not compile anymore on my machine
```
./build/include/posix.c:65:10: fatal error: 'sys/sdt.h' file not found
   65 | #include <sys/sdt.h>  // dtrace_probe
      |          ^~~~~~~~~~~
1 error generated.
```

michaellilltokiwa:
fixes #7101


fridis (approved):
ok, this looks nicer than the linux-specific check. 

But nevertheless, I think it would be good if `DTRACE_PROBE` was available on the systems we work with if possible.

PR #7104, opened 2026-05-20
Dockerfile: Add installation of systemtap-sdt-dev

fridis:
required for `DTRACE_PROBE`

[no ci]


maxteufel (approved):

michaellilltokiwa:
Not sure if its enough to just add this to builder and not runner but lets try.

maxteufel:
It is; all the header file does is define macros that add additional assembler code to the binary, there is no library involved.

PR #7107, opened 2026-05-20
lexer: change string interpolation according to #6670

### LINKED ISSUE #6670: Use `"x=$x x+y=$(x+y)"` instead of `"x=$x x+y={x+y}"` in Strings, drop support for `"{x,y}"`
I sometimes think that the dollar interpolation option in Strings is not useful enough to warrant it. Even worse if we later want to change the dollar expression to something more complex we have to replace the dollar by braces. I could also imagine that readability is better when using just one mechanism. Another benefit is that there would not be a need to escape dollar signs anymore.

```
"{"*** FAILED".bold.red} out on {file}"
```
instead of 
```
"{"*** FAILED".bold.red} out on $file"
```

michaellilltokiwa:
Support only `$x` or `$(expr)` as noted in the linked issue.


src/dev/flang/parser/Lexer.java
```diff
@@ -3248,6 +3250,12 @@ else if (!skipped(pos))
                 }
               pos = advance(pos, 1);
               p = raw(pos);
+              if (foundStart && p == '(')
```
fridis:
```suggestion
              if (end(t)==StringEnd.DOLLAR && p == '(')
```
should IMHO work and then you can remove `foundStart`.

fridis (commented):
Looks good. Just a small suggestions to remove the redundant local var `foundStart`.

I did not check all the changes to string literals in the fuzion code, though...

src/dev/flang/parser/Lexer.java
```diff
@@ -325,13 +325,13 @@ public String toString()
...
-            case t_stringBQ          : result = "string constant after {<expr>}"             ; break;
-            case t_stringBD          : result = "string constant after {<expr>} ending in $" ; break;
-            case t_stringBB          : result = "string constant after {<expr>} ending in {" ; break;
+            case t_StringDB          : result = "string constant after $<id> ending in $("   ; break;
+            case t_stringPQ          : result = "string constant after $(<expr>)"            ; break;
+            case t_stringPD          : result = "string constant after $(<expr>) ending in $"; break;
+            case t_stringPB          : result = "string constant after $(<expr>) ending in $("; break;
             case t_eof               : result = "end-of-file"                                ; break;
             default                  : result = super.toString()                             ; break;
             }
```
simonvonhackewitz: Now the alignment is off.

I think the modern switch case syntax could simplify this a bit.
```suggestion
          result = switch (this) {
              case t_op         -> "operator";
              case t_comma      -> "comma ','";
              case t_period     -> "period '.'";
              case t_lparen     -> "left parenthesis '('";
              case t_rparen     -> "right parenthesis ')'";
              case t_lbrace     -> "left curly brace '{'";
              case t_rbrace     -> "right curly brace '}'";
              case t_lbracket   -> "left bracket '['";
              case t_rbracket   -> "right bracket ']'";
              case t_semicolon  -> "semicolon ';'";
              case t_question   -> "question mark '?'";
              case t_numliteral -> "numeric literal";
              case t_ident      -> "identifier";
              case t_stringQQ   -> "string constant";
              case t_stringQD   -> "string constant ending in $";
              case t_stringQB   -> "string constant ending in $(";
              case t_StringDQ   -> "string constant after $<id>";
              case t_StringDD   -> "string constant after $<id> ending in $";
              case t_StringDB   -> "string constant after $<id> ending in $(";
              case t_stringPQ   -> "string constant after $(<expr>)";
              case t_stringPD   -> "string constant after $(<expr>) ending in $";
              case t_stringPB   -> "string constant after $(<expr>) ending in $(";
              case t_eof        -> "end-of-file";
              default           -> super.toString();
          };
```

maxteufel (approved):

PR #7110, opened 2026-05-21
base: Use mapping to orderable type instead of comparator for sorting

### LINKED ISSUE #7091: `Sequence.sort_by` should receive a feature that maps to an orderable type
Should be like this
```
  public sort_by(O type : property.orderable, by T->O) container.sorted_array T
  => ...
```
Such that we can do 
```
  ["hi","Hello","Hola"].sort_by (.byte_count) |> say
```
instead of
```
  ["hi","Hello","Hola"].sort_by (a,b->a.byte_count <= b.byte_count) |> say
```
Other features in the base module should be checked if they could be changed accordingly.

simonvonhackewitz:
fix #7091

### DIFF
--- a/benchmarks/priority_queue_benchmark.fz
+++ b/benchmarks/priority_queue_benchmark.fz
@@ -36,7 +36,7 @@ priority_queue_benchmark =>
         m : mutate is
         m ! ()->
 
-          q := container.Binary_Heap_Queue m i32 .empty_min_first (size + 10).as_i64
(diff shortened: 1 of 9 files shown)


michaellilltokiwa (approved):

maxteufel:
This still lacks documentation that now the smallest element according to the defined ordering is returned, and that for reverse ordering some special wrapping must be used (e.g. in `empty_by`).

PR #7113, opened 2026-05-21
ast: `setFunctionReturnType` in loop only when `successBlock` is present

### LINKED ISSUE #6964: Wrong error `Loop results in two incompatible types` produced either when not justified or when rudandant
This is related to #6263, I have tried to reduce that example to the bare minimum:
```
_ ()->i32 := ()->
    for a in 0..0 do
      for b in 0..0 do
    else
      0
```
which produces an error that is IMHO not justified and that is also not detailed enough to actually explain what the problem is:
```
 > FUZION_DISABLE_ANSI_ESCAPES=true ./build//bin/fz reg_issue6263.fz

/home/fridi/fuzion/clean/fuzion.1/reg_issue6263.fz:3:7: error 1: Loop results in two incompatible types.
      for b in 0..0 do
------^
The incompatible types are: 'i32','unit'

one error.
```
we can see that this error is not justified if we remove the surrounding lambda `()->`:
```
_ i32 :=
    for a in 0..0 do
      for b in 0..0 do
    else
      0
```
then it works
```
 > ./build//bin/fz reg_issue6263ok.fz
 >
```
So, the original problem should IMHO not have produced an error. 

I suspect type propagation to or type inference from the lambda to somehow mess up the loop code such that we get into this error case. 

If I write code that actually justifies an error for two incompatible loop result types as follows
```
_ i32 :=
    for a in 0..0 do
      for b in 0..0 do
      until b=3 then
        unit
      else
        3.14
    else
      0
```
then I do get a proper error for the incompatible types
```
 > FUZION_DISABLE_ANSI_ESCAPES=true ./build//bin/fz reg_issue6263mustfail.fz

/home/fridi/fuzion/clean/fuzion.1/reg_issue6263mustfail.fz:4:14: error 1: Incompatible types in branches of if expression
      until b=3 then
-------------^
block returns value of type 'unit' at /home/fridi/fuzion/clean/fuzion.1/reg_issue6263mustfail.fz:5:9:
        unit
--------^^^^
block returns value of type 'f64' at /home/fridi/fuzion/clean/fuzion.1/reg_issue6263mustfail.fz:6:7:
      else
------^
```
There are only minor problems with this error, it should refer to `loop` and not `if expression` and it would be great if it would show the source position of `3.14` and not `else` (best would be to show `else 3.14` such that it also works for an empty `else`, and similar for the positive result `unit` that gets a bit messed up if we leave out `unit` or `then` or both).

Additionally, there is a redundant error for the Loop results that refers to `**error**` and that should not be shown. 
```

/home/fridi/fuzion/clean/fuzion.1/reg_issue6263mustfail.fz:3:7: error 2: Loop results in two incompatible types.
      for b in 0..0 do
------^
The incompatible types are: '**error**','f64'

2 errors.
```

### LINKED ISSUE #7088: Unexpected propagation of expected type from effect instatement into nested loop
This is a simplified example of a problem in AoC solution [2023/18/part1.fz](https://github.com/fridis/fuzion_aoc/blob/995a015a8b51b73d64994b0c0fb5249b3fbeab15/2023/18/part1.fz#L31).

The expected type `i32` from `lm.instate_self` is propagated to the nested loop. Which is wrong in this particular case, as the inner loop should result in `unit`.

```
part1 =>
  lm : mutate is
  lm.instate_self i32 ()->
  
    for x in [0,1,2] do
      for i in 0..3 do unit
    else
      0

say part1
```

```
❯ fz /tmp/ex.fz | cat

/tmp/ex.fz:6:7: error 1: Loop results in two incompatible types.
      for i in 0..3 do unit
------^
The incompatible types are: 'i32','unit'

one error.
```

michaellilltokiwa:
fixes #6964

It does not make sense to set/propagate result type to loop that does not produce a result.


maxteufel (dismissed):

michaellilltokiwa:
- 6263 failed compilation, but I think, it should not have failed. It compiles now but requires stdin, hence I changed from `simple.mk` to `compile.mk`.
- 6540, here the error message changed slightly, it was not acceptable before and isn't now. I will create an issue for this.

maxteufel (approved):

PR #7125, opened 2026-05-26
Modified Jitter so that testing threads send raw data back to the main thread. 

Nimux2:
Changes in this branch:

- Worker/main separation: the worker thread only collects raw timing samples and returns them to the caller via a `concur.Channel`. The main thread consumes those lists and builds the histograms. This decouples measurement from analysis and simplifies future multi-threaded extensions.
- Dedicated-thread path: the example can run the test on a spawned worker thread (default) or on the main thread.
- CPU affinity: the worker tries to pin itself to CPU 0. Success or failure is logged as part of the run output.
- Real-time scheduling (SCHED_FIFO) is implemented similarly to CPU affinity, and can be disabled if necessary.
- Precision note: This version is significantly more accurate/efficient than the original (thread-less).


Nimux2:
@maxteufel Could you review this ?

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,206 @@
...
+  test_jitter(# the clock used to schedule the releases
+              C type : time.Clock,
+
+              # the clock used to measure the times
+              TIMER type : time.Clock)
+  =>
+
+    mutate ! ()->
+
+      results := mutate.env.new (list (tuple time.duration (list time.duration) (list time.duration))) nil
```
maxteufel: I suggest dropping the use of mutate here. results could be sent directly to the channel, or is there a reason for that to not be possible here?

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,206 @@
...
+  if enable_rt
+    match thrd.set_policy(concur.scheduling_policy.sched_fifo, false, 99)
+      unit => say "Worker thread configured for SCHED_FIFO priority 99"
+      e error =>
+        say "Worker thread priority setup failed: $e"
+        say "(Hint: run with sudo or use setcap cap_sys_nice=ep ./jitter)"
+
+  # wait for thread to send results
+  match result_ch.next
+    samples list =>
```
maxteufel: In the case I suggest above, you would not get one single list but many data points separately on the channel.

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,206 @@
...
+            err := error "*** overrun: run started at {actual_release-expected_rel}  and took {actual_finish-actual_release}, which is larger than period $period"
+            (exception jitter_too_high).env.cause err
+
+      match exception jitter_too_high .on ()->
+
+           for p := period_default, p.times 0.5 do
+             run_for_period p
+
+        _ error => results.get.reverse_list
+        void    => results.get.reverse_list
```
maxteufel: `void` case should not have any code. this is a case that can never ever happen.

maxteufel (changes requested):

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,208 @@
...
+            say "$s (never called)"
+
+      # perform jitter test for given period: collect raw samples
+      run_for_period(period time.duration)
+      =>
+
+        start := C.env.read + start_delay
+        end := start + test_duration
+
+        state (sample_acc nil nil nil) ! ()->
```
maxteufel: I meant that collecting any sort of mutable state is bad design here if possible to avoid it.

You can send to a channel (`result_ch`) more than once, and it would be very neat if you collected the tuple for one run, sent it to a channel, and then build the list/iterate through those results in the other thread, rather than building up a list here and then sending it to the other thread only when you've built the list.

maxteufel (changes requested):

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,208 @@
...
+            if actual_finish >= next_rel
+              err := error "*** overrun: run started at {actual_release-expected_rel}  and took {actual_finish-actual_release}, which is larger than period $period"
+              (exception jitter_too_high).env.cause err
+
+      _ := match exception jitter_too_high .on ()->
+
+           for p := period_default, p.times 0.5 do
+             run_for_period p
+
+        _ error => unit
```
maxteufel: in case of error you can still say what kind of error happened

maxteufel (commented):
There is still mutability in the `run_for_period` feature in the form of usage of the `state` effect. Please explain what you are trying to do here. What do you need the data from previous runs here for?

Nimux2:
@maxteufel `state` was already used in the original, from what I understand it's used to compute the current cycle's duration based on the previous one.

maxteufel:
@Nimux2 yes, but you use more `state`, see `state_acc`. none of that data is relevant to the current cycle anymore as long as it has been sent to the other thread already.

maxteufel (approved):
nice!

maxteufel:
I'd love to see a small final comparison of the both jitter versions alongside the conclusions you draw from that.

Nimux2:
@maxteufel 
Old version :
- Single thread, no affinity/priority
- Histograms mutated in-place during release
- Result display after each period's run_periodic call returns
- Accuracy of 3781µs

New version :
- Dedicated worker + main thread
- Histograms folded from streamed tuples on main thread
- Inter-thread communication through concur.Channel of tuples
- Result display incrementally, the moment each period's done_flag record is consumed
- CPU affinity + SCHED_FIFO attempt
- Accuracy of 42µs

I think that thanks to a worker thread exclusively focused on collecting data, that thread can achieve higher performances and thus be more accurate.

maxteufel:
what is your definition of accuracy here?

Nimux2:
> what is your definition of accuracy here?

The rate at which the worker thread can run and collect data.

maxteufel:
i still don't understand, this isn't an output of the jitter program? how did you calculate this?

Nimux2:
> i still don't understand, this isn't an output of the jitter program? how did you calculate this?

The histograms have additional details regarding each testing period.
Here's for the old version :
```
testing period 3906µs for 2100ms...

---  period jitter 3906µs, log count / linear time  ---            

Histogram here (it's unreadable if just copy-pasted)

 3781µs  | 3627µs  | 5962µs  |  165µs  |   536   |

```
 
 
 And for the new one
```
testing period 15ms for 2100ms...
period jitter:
            ---  period jitter 15ms, log count / linear time  ---             

Histogram

  42µs   |  27µs   |  99µs   | 9631ns  |   134   |
```

maxteufel:
i don't think comparing jitter average for different periods is a good idea.

Nimux2:
> i don't think comparing jitter average for different periods is a good idea.

Right, my bad. I just picked the last period of each. The equivalent period for the old jitter would be :
```
              ---  cycle jitter 15ms, log count / log time  ---               

Histogram

  15ms   |  15ms   |  15ms   |  113µs  |   133   |
```

Do you think something else would be more appropriate as measurement though ?

maxteufel:
that's a better comparison yeah

examples/jitter/jitter_threaded.fz
```diff
@@ -0,0 +1,253 @@
...
+#
+# -----------------------------------------------------------------------
+
+# jitter -- demo measuring periodic task jitter using time.histogram
+#
+# See Makefile for how to start fuzion to run this example.
+#
+# NYI: UNDER DEVELOPMENT: Currently, this supports only one single test using
+# `time.nano.run_periodic`.  Once we have support for periodic tasks run by dedicated
+# threads, we should add jitter tests for those.
```
simonvonhackewitz: I think the comments above are still referring to jitter.fz

Nimux2: Good catch, I've fixed it now.

Nimux2:
@maxteufel Do you think this is ready to be merged ?

maxteufel:
yes, hence my approval. however I'd like the rest of the team @tokiwa-software/developers to have a look too.

Nimux2:
> yes, hence my approval. however I'd like the rest of the team @tokiwa-software/developers to have a look too.

That's fine, I just can't merge it myself.

PR #7127, opened 2026-05-27
time/histogram: make `histogram_mutate` public for use in other threads

### LINKED ISSUE #7089: time/histogram: `histogram_mutate` causes invalid mutate across thread boundaries
error involving invalid mutate.

@Nimux2 can you please look at this?

maxteufel:
Since it causes histograms being used from other threads break.

### DIFF
--- a/modules/base/src/time/histogram.fz
+++ b/modules/base/src/time/histogram.fz
@@ -438,7 +438,7 @@ is
 
 # a local mutate instance we are using to modify our state
 #
-histogram_mutate : mutate
+public histogram_mutate : mutate
 is
 
   # default implementation of mutate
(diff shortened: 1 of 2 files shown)


modules/base/src/time/histogram.fz
```diff
@@ -438,11 +438,5 @@ is
...
 #
-histogram_mutate : mutate
+public histogram_mutate : mutate
 is
-
-  # default implementation of mutate
-  #
-  # this will get instated automatically at startup
-  #
-  public redef fixed type.default_value option time.histogram_mutate => time.histogram_mutate
```
fridis, line 448: this should stay.

fridis (changes requested):
For now, it should be sufficient to add an explicit ` time.histogram_mutate ! ()->` when histogram is used in any but the main thread, so please remove the ones you added that are used only in the main thread.

For the main thread, the `default_value` should remain. 

Eventually, there should be a `default_thread_local_value` such that each thread would automatically get its private `histogram_mutate`.

fridis (commented):
IMHO much nicer like this.

fridis (approved):

PR #7134, opened 2026-06-01
base: improve error message in `os.process`

michaellilltokiwa:


simonvonhackewitz:
This would produce messages like `error(code=9): error reading from stdout of process 12345`.
Is it intentional that the word "error" appears twice here? It feels a bit redundant.

modules/base/src/os/process.fz
```diff
@@ -36,31 +36,31 @@ module:public process(public pid i64, module std_in, std_out, std_err i64, name
...
-  read_handler(desc i64) : io.Read_Handler is
+  read_handler(name String, desc i64) : io.Read_Handler is
     public redef read(max_count i32) choice (Sequence u8) io.end_of_file error =>
       arr := array u8 pipe_buffer_size i->0
       res := fzE_pipe_read desc arr.internal_array.data arr.count
       if res = -1
         # NYI: UNDER DEVELOPMENT: register failures
         _ := fzE_pipe_close desc
-        error "failed reading from stdout" fzE_last_error
+        error "error reading from {read_handler.this.name} of process {pid}." fzE_last_error
```
michaellilltokiwa:
```suggestion
        error "reading from {read_handler.this.name} of process {pid}." fzE_last_error
```

modules/base/src/os/process.fz
```diff
@@ -72,10 +72,10 @@ module:public process(public pid i64, module std_in, std_out, std_err i64, name
 
   # write bytes to standard in
   #
-  write_handler (desc i64) : io.Write_Handler is
+  write_handler (name String, desc i64) : io.Write_Handler is
     public redef write (bytes Sequence u8) outcome unit =>
       if fzE_pipe_write desc bytes.as_array.internal_array.data bytes.count = -1
-        error "failed writing to stdin" fzE_last_error
+        error "error writing to {write_handler.this.name} of process {pid}. Tried writing {bytes.count} bytes." fzE_last_error
```
michaellilltokiwa:
```suggestion
        error "writing to {write_handler.this.name} of process {pid}. Tried writing {bytes.count} bytes." fzE_last_error
```

modules/base/src/os/process.fz
```diff
@@ -36,31 +36,31 @@ module:public process(public pid i64, module std_in, std_out, std_err i64, name
...
   # standard in of the process
   #
-  std_in_write_handler => write_handler std_in
+  std_in_write_handler => write_handler "stdin" std_in
 
 
   # read bytes from standard out
   #
-  read_handler(desc i64) : io.Read_Handler is
+  read_handler(name String, desc i64) : io.Read_Handler is
```
fridis: I think it would be better to not add `name` as an argument field, but instead to have 
```
private desc_name(desc) =>
  if      desc = std_in  then "stdin"
  else if desc = std_out then "stdout"
  else if desc = std_err then "stderr"
  else "file_desc#$desc"
```
defined in `process` and use this to create the error message.

michaellilltokiwa: Yes, that's better

modules/base/src/os/process.fz
simonvonhackewitz, line 110: What about this error message?

simonvonhackewitz, line 170: Or this?

simonvonhackewitz (commented):
Sorry, I just realized how few lines have actually changed.

modules/base/src/os/process.fz
michaellilltokiwa: You mean if we should also add e.g. pid info to the error message?

modules/base/src/os/process.fz
```diff
@@ -51,6 +51,15 @@ module:public process(public pid i64, module std_in, std_out, std_err i64, name
   std_in_write_handler => write_handler std_in
 
 
+  # get a name for the given descriptor
+  #
+  desc_name(desc) =>
+    if      desc = std_in  then "stdin"
+    else if desc = std_out then "stdout"
+    else if desc = std_err then "stderr"
+    else "file_desc#$desc"
```
maxteufel: i think we can use `file_descriptor` in verbose here

maxteufel (changes requested):

modules/base/src/os/process.fz
michaellilltokiwa:
```suggestion
  desc_name(desc) =>
    if      desc = std_in  then "stdin"
    else if desc = std_out then "stdout"
    else if desc = std_err then "stderr"
    else "file_descriptor#$desc"
```

modules/base/src/os/process.fz
simonvonhackewitz: Yes, and also use `desc_name`.

michaellilltokiwa: I prefer to do this in a separate PR.

simonvonhackewitz (approved):

maxteufel (approved):

PR #7136, opened 2026-06-01
fe: confusing error: "expected formal type: 'T' actual type found : 'T'"

### LINKED ISSUE #6921: confusing error: expected formal type: 'T' actual type found   : 'T'
```
a(T type) is
  b T => abstract
  type.new(F type: i32->T, f F) a T=> a T
  public transpose a T => (a T).new _ i->b
```

```
/home/sam/playground/test.fz:6:20: error 1: Incompatible types in assignment
    (a T).new _ i->b

assignment to field : 'a.transpose.(λi->b).call.result'
expected formal type: 'T'
actual type found   : 'T'
assignable to       : 'T'
for value assigned  : 'b'
To solve this, you could change the type of the target 'a.transpose.(λi->b).call.result' to 'T' or convert the type of the assigned value to 'T'.

one error.

``

michaellilltokiwa:
fixes #6921

In essence this changes `toString` for a type parameter, if the outer is a co-type to also include this in the name. Normal type parameters are still stringified as just  the baseName of the feature.

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -2420,7 +2420,13 @@ there is no common super type of the two types (Types.t_ERROR)
     if (isGenericArgument())
       {
         var ga = genericArgument();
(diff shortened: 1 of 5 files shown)


maxteufel (approved):
tests ok; only wolfssl failing due to fuzion-lang.dev downtime.

PR #7150, opened 2026-06-03
ast: suppress followup error for result type `**error**` in native feat

### LINKED ISSUE #7149: Follow-up error for result type `**error**` should be suppressed
Try this: 
```
 > ~/fuzion/fuzion/build/bin/fz -e "x => native"

command line:1:1: error 1: Failed to infer result type for feature 'x'.
x => native

To solve this, please specify a result type explicitly.


error 2: Implementation restriction: Result type '**error**' is not (yet) allowed in native features.
To solve, this specify a legal type.

2 errors.

```
The second error should be suppressed in case there were previous errors and the result type is `Types.t_ERROR`.

simonvonhackewitz:
fix #7149

### DIFF
--- a/src/dev/flang/ast/AstErrors.java
+++ b/src/dev/flang/ast/AstErrors.java
@@ -2579,14 +2579,19 @@ public class AstErrors extends ANY
 
   public static void explicitTypeRequired(AbstractFeature f, AbstractType inf)
   {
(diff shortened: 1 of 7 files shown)


maxteufel (changes requested):
"To solve this, please specify a type explicitly." error resolution message.

maxteufel (approved):

PR #7151, opened 2026-06-03
tests/be/int: update test recordings

michaellilltokiwa:


simonvonhackewitz:
The highlighted source range was changed in #7143, but tests did not fail in the PR because the interpreter backend does not get tested in pull requests.

simonvonhackewitz (approved):

PR #7152, opened 2026-06-03
Add method `as_string_decimals` to convert a float to a string with a custom number of decimals

Nimux2:
Solution to [this](https://github.com/tokiwa-software/fuzion-idioms/pull/396) discussion about [Idiom 23](https://fuzion-lang.dev/idioms/idiom23).
I will open a pull request for the related Idiom once this one is merged, to avoid causing issues.

Here's an usage example : 
```
y := 3.14563

say (y.as_string_decimals 2)  # Output: 3.14

x := 3.1

say (x.as_string_decimals 2)  # Output: 3.10

z := 3

say (z.as_string_decimals 2)  # Output: 3.00
```

### DIFF
--- a/modules/base/src/f32.fz
+++ b/modules/base/src/f32.fz
@@ -198,15 +198,22 @@ public f32(public val f32) : float, java_primitive is
 
 
   # convert this to a string with a fixed number of digits after the decimal point.
-  #
+  # If the given values are unusable, the result is the same as `as_string`.
+  # Overflows are technically possible, but floats become inaccurate long before that.
   public redef as_string_decimals(fraction_digits i32) String =>
-    if fraction_digits <= 0
+    if is_NaN
+      "NaN"
+    else if val = f32.positive_infinity
+      "Infinity"
+    else if val = f32.negative_infinity
+      "-Infinity"
+    else if fraction_digits <= 0
       f32.this.as_string
     else
       scale := 10.0 ** fraction_digits.as_f64
 
       rounded_scaled :=
-        if val >= 0.0
+        if val.as_f64 >= 0.0
           ((val.as_f64 * scale) + 0.5).as_i64_lax
         else
           ((val.as_f64 * scale) - 0.5).as_i64_lax
--- a/modules/base/src/f32.fz
+++ b/modules/base/src/f32.fz
@@ -203,7 +203,17 @@ public f32(public val f32) : float, java_primitive is
     if fraction_digits <= 0
       f32.this.as_string
     else
-      s String := (num.ryū f32).as_string val false (num.ryū f32).rounding_conservative
+      scale := 10.0 ** fraction_digits.as_f64
+
+      rounded_scaled :=
+        if val >= 0.0
+          ((val.as_f64 * scale) + 0.5).as_i64_lax
+        else
+          ((val.as_f64 * scale) - 0.5).as_i64_lax
+
+      rounded := rounded_scaled.as_f64 / scale
+
+      s String := (num.ryū f32).as_string rounded.as_f32 false (num.ryū f32).rounding_even
       match s.find "."
         i i32 =>
           digits := s.substring i+1
--- a/modules/base/src/f32.fz
+++ b/modules/base/src/f32.fz
@@ -197,6 +197,23 @@ public f32(public val f32) : float, java_primitive is
     (num.ryū f32).as_string val
 
 
+  # convert this to a string with a fixed number of digits after the decimal point.
+  #
+  public redef as_string(fraction_digits i32) String =>
+    if fraction_digits <= 0
+      f32.this.as_string
+    else
+      s String := (num.ryū f32).as_string val false (num.ryū f32).rounding_conservative
+      match s.find "."
+        i i32 =>
+          digits := s.substring i+1
+          if digits.byte_count > fraction_digits
+            "{s.substring 0 i}.{digits.substring 0 fraction_digits}"
+          else
+            "{s.substring 0 i}.{digits.pad "0" fraction_digits}"
+        nil => s
+
+
   # -----------------------------------------------------------------------
   #
   # type features:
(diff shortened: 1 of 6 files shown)


michaellilltokiwa:
This is good for now, in the long run we probably want an as_string with quite a few formatting options `as_string(fo formatting_options) =>`

- precision
- width
- zero padding
- signed (if + should always be included)
- thousands separator 
- fraction separator
- ...

maxteufel:
This just seems to cut off digits, i.e.

```
$ fz -e 'say <| 3.49 .as_string_decimals 1'
3.4
```

I would expect the result here to be 3.5.

maxteufel:
Please document your assumptions. This does not work with special floats such as ±Infinity or NaNs.

Nimux2:
> Please document your assumptions. This does not work with special floats such as ±Infinity or NaNs.
@maxteufel Are there other outliers I should be aware of ?

modules/base/src/f32.fz
```diff
@@ -197,6 +197,40 @@ public f32(public val f32) : float, java_primitive is
...
+  # convert this to a string with a fixed number of digits after the decimal point.
+  # If the given values are unusable, the result is the same as `as_string`.
+  # Overflows are technically possible, but floats become inaccurate long before that.
+  public redef as_string_decimals(fraction_digits i32) String =>
+    if is_NaN
+      "NaN"
+    else if val = f32.positive_infinity
+      "Infinity"
+    else if val = f32.negative_infinity
+      "-Infinity"
```
maxteufel: you should rely on the `as_string` provided by ryū here instead of duplicating the string constants.

maxteufel (commented):
You might need to consider `0` and `-0` to be a special case too (the latter being equal to the former but should always be normalized to `0`).

Nimux2:
> You might need to consider `0` and `-0` to be a special case too (the latter being equal to the former but should always be normalized to `0`).

It's already normalized that way apparently, probably due to the rounding.

modules/base/src/int.fz
```diff
@@ -143,6 +143,10 @@ is
   public redef as_string String =>
     if ns then "$n" else "-$n"
 
+  # Convert this to a string with a fixed number of digits after the decimal point.
+
```
michaellilltokiwa:
```suggestion
    #
```

modules/base/src/i64.fz
```diff
@@ -126,6 +126,10 @@ public i64(public val i64) : num.wrap_around, has_interval, java_primitive is
   public as_int int =>
     int val
 
+  # Convert this to a string with a fixed number of digits after the decimal point.
+  public as_string_decimals(fraction_digits i32) String =>
```
michaellilltokiwa, line 131:
```suggestion
  # Convert this to a string with a fixed number of digits after the decimal point.
  #
  public as_string_decimals(fraction_digits i32) String =>
```

modules/base/src/i32.fz
```diff
@@ -106,6 +106,9 @@ public i32(public val i32) : num.wrap_around, has_interval, java_primitive is
   public as_f64 f64 => intrinsic
   public as_f32 f32 => as_f64.as_f32
 
+  # Convert this to a string with a fixed number of digits after the decimal point.
+  public as_string_decimals(fraction_digits i32) String =>
```
michaellilltokiwa, line 111:
```suggestion
  # Convert this to a string with a fixed number of digits after the decimal point.
  #
  public as_string_decimals(fraction_digits i32) String =>
```

modules/base/src/f64.fz
```diff
@@ -205,6 +205,34 @@ public f64(public val f64) : float, java_primitive is
     (num.ryū f64).as_string val
 
 
+  # convert this to a string with a fixed number of digits after the decimal point.
+  # If the given values are unusable, the result is the same as `as_string`.
+  # Overflows are technically possible, but floats become inaccurate long before that.
+  public redef as_string_decimals(fraction_digits i32) String =>
```
michaellilltokiwa, line 212:
```suggestion
  # Overflows are technically possible, but floats become inaccurate long before that.
  #
  public redef as_string_decimals(fraction_digits i32) String =>
```

modules/base/src/f32.fz
```diff
@@ -197,6 +197,34 @@ public f32(public val f32) : float, java_primitive is
     (num.ryū f32).as_string val
 
 
+  # convert this to a string with a fixed number of digits after the decimal point.
+  # If the given values are unusable, the result is the same as `as_string`.
+  # Overflows are technically possible, but floats become inaccurate long before that.
+  public redef as_string_decimals(fraction_digits i32) String =>
```
michaellilltokiwa, line 204:
```suggestion
  # Overflows are technically possible, but floats become inaccurate long before that.
  #
  public redef as_string_decimals(fraction_digits i32) String =>
```

modules/base/src/float.fz
```diff
@@ -75,6 +75,11 @@ public float : numeric is
       (float.this + float.this.one/float.this.two).floor
 
 
+  # convert this float to a string with a fixed number of digits after the decimal point.
+  #
+  public as_string_decimals(fraction_digits i32) String => abstract
```
michaellilltokiwa, line 82: Please add an NYI comment here that states that in the future this should be replaced by a more general as_string that takes formatting_options.

michaellilltokiwa (commented):
I think we can merge this as soon as some minor nitpicks are addressed.

modules/base/src/float.fz
```diff
@@ -75,6 +75,13 @@ public float : numeric is
       (float.this + float.this.one/float.this.two).floor
 
 
+  # convert this float to a string with a fixed number of digits after the decimal point.
+  # NYI : In the future, this should be replaced by a version of as_string with more formatting options,
+  # as described here : https://github.com/tokiwa-software/fuzion/pull/7152#issuecomment-4613267795
```
maxteufel: we have a particular format of `NYI: CATEGORY: ...` where `CATEGORY` in this case would probably be `ENHANCEMENT`.

maxteufel (approved):
Looks mostly fine now.

PR #7154, opened 2026-06-03
native: make `fzE_file_read` handle `EINTR`

### LINKED ISSUE #7135: Reading from stdout fails sometimes because of bad file descriptor
Since [flang_dev#679](https://git.tokiwa.software/tokiwa/flang_dev/pulls/679) the flang_dev Jenkins pipeline [fails sometimes](https://jenkins.tokiwa.software/job/flang.dev/job/main/1073/console), caused by exit code 9 "Bad file descriptor":

```
*** failed *** panic ***: `error(code=9): failed reading from stdout`
```

fridis:
Also removed always true comparison of unsigned `size_t` value `>= 0`. Check if `fread` reached `feof` and return 0 if that is the case or would block (`errno = EAGAIN`) and retry unless we have already received some data, then return that data. For other error cases, return `-1`.

The missing handling of `EINTR` caused the Fuzion version for the `feeze_recorder` to terminate occasionally out of thin air since it thought its `stdin` input stream reached `end_of_file`. The proper handling of `feof` and `EAGAIN` was required for `feeze_recorder` and all tests to succeed.


maxteufel (dismissed):

michaellilltokiwa (approved):
Looks reasonable to me.

PR #7156, opened 2026-06-05
os: add basic support for creating pipes

maxteufel:
A small example for this:

```
$ cat ex_pipe.fz 
max_test_time := time.duration.s 10

envir_vars := envir.vars.env.map_of [ "PATH", "LANG" ]

Sequence.prefix ! outcome tokiwa.process_result =>
  seq := map x->x.as_string
  os.process.start seq.first.or_panic (seq.drop 1) envir_vars
            .bind (p -> tokiwa.record_process p Sequence.this.as_string max_test_time nil)

ex_pipe =>
  os.pipe ! ()->
    lm : mutate is
    _ := lm ! ()->
      os.pipe.env.with_writer _ lm (()->
          (io.buffered lm).writer.env.write "hello, world!\n" .or_panic
          (io.buffered lm).writer.env.flush.or_panic
          say "written") (r->
          x := !["/usr/bin/cat", "/dev/fd/{r}"]
          say x)

_ := ex_pipe

$ fz -modules=tokiwa ex_pipe.fz
written
0: hello, world!
```

### DIFF
--- a/include/fz.h
+++ b/include/fz.h
@@ -474,6 +474,16 @@ int fzE_process_create(char * args[], size_t argsLen, char * env[], size_t envLe
 int64_t fzE_process_poll(int64_t p);
 
 /**
+ * open a new pipe, put read and write end into array passed here
+ *
+ * @param fds array to be filled with pipe's file descriptors
+ *
+ * @return non-zero error number, or 0 on success
+ *
+ */
+int fzE_pipe_create(int64_t * fds);
+
+/**
  * read nbytes bytes into `buf` from pipe `desc`.
  *
  * @return -1 error, bytes read on success
--- a/modules/base/src/os/pipe.fz
+++ b/modules/base/src/os/pipe.fz
@@ -1,0 +1,132 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature os.pipe
+#
+# -----------------------------------------------------------------------
+
+
+# effect for creating pipes
+#
+private:public pipe : effect is
+
+  # the underlying pipe's file descriptors
+  #
+  pipefd array i64 := [0, 0]
+
+
+  # create the pipe
+  #
+  check debug: pipefd.count = 2
+  status := fzE_pipe_create pipefd.internal_array.data
+
+
+  # has this pipe been used already?
+  #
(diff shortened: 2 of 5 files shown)


modules/base/src/os/pipe.fz
```diff
@@ -0,0 +1,85 @@
...
+public pipe : effect is
+
+  # the underlying pipe's file descriptors
+  #
+  pipefd array i64 := [0, 0]
+
+
+  # create the pipe
+  #
+  check pipefd.count = 2
```
michaellilltokiwa: this check seems useless to me?

modules/base/src/os/pipe.fz
```diff
@@ -0,0 +1,85 @@
...
+  status := fzE_pipe_create pipefd.internal_array.data
```
michaellilltokiwa, line 37: This creates the pipes in the features constructor, not on effect instantiation. This is not wrong but IMHO a little prone to misuse, since if the effect is not instated `finally` will never be called.

Another way to model this would be to mark the effect as `private:public`, then add a public type-feature taking a callback `code` that does the construction of the effect and then instates it and runs the code while pipe-effect is instated.

modules/base/src/os/pipe.fz
maxteufel: The intrinsic `fzE_pipe_create` does not take the length of the array. Since it should always be equal to 2, I want to document this requirement somewhere. (Bad things happen when the length is 0 or 1).

modules/base/src/os/pipe.fz
maxteufel: Using `private:public` seems a bit like a hack, but this is a good point. Maybe we need something like an opposite of `finally`, that hooks into effect instation?

modules/base/src/os/pipe.fz
michaellilltokiwa: Hmm, okay.
```suggestion
  check debug : pipefd.count = 2
```

modules/base/src/os/pipe.fz
michaellilltokiwa: @maxteufel The effect lifecycle is not well defined yet. I think there should already be issues discussing this problem and others.

modules/base/src/os/pipe.fz
```diff
@@ -0,0 +1,85 @@
...
+      if fzE_pipe_write desc bytes.as_array.internal_array.data bytes.count = -1
+        error "failed writing to {desc}" fzE_last_error
+      else
+        unit
+
+
+  # with writer for one pipe
+  #
+  public with_writer(T type, LM type : mutate, w () -> outcome unit, r i64 -> T) outcome T =>
+    if used
```
michaellilltokiwa: should probably be a precondition and field `used` should be public

modules/base/src/os/pipe.fz
```diff
@@ -0,0 +1,85 @@
...
+      error "cannot reuse pipe"
+    else if status != 0
+      error "creating pipe" status.as_i64
+    else
+      set used := true
```
michaellilltokiwa: consider making `used` an atomic. I find `set` even though done in base library kind of ugly

include/fz.h
```diff
@@ -473,6 +473,16 @@ int fzE_process_create(char * args[], size_t argsLen, char * env[], size_t envLe
  */
 int64_t fzE_process_poll(int64_t p);
 
+/**
+ * open a new pipe, put read and write end into array passed here
+ *
+ * @param fds array to be filled with pipe's file descriptors
+ *
+ * @return error number returned by pipe call, or 0 on success
```
michaellilltokiwa:
```suggestion
 * @return non-zero error number, or 0 on success
```

michaellilltokiwa (approved):
Minor suggestion for a comment.

PR #7157, opened 2026-06-08
base: add wrapper type `uid` to be used by `unique_id` effect

### LINKED ISSUE #6405: unique id, should probably not be an `u64` but some wrapper type
Goal is to reduce (accidental) misuse.

simonvonhackewitz:
fix #6405

@michaellilltokiwa is this what you were thinking?

What's the preferred way to create new instances? Provide a type feature `type.new` or make the constructor public?

### DIFF
--- a/modules/base/src/uid.fz
+++ b/modules/base/src/uid.fz
@@ -1,0 +1,42 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature uid
+#
+# -----------------------------------------------------------------------
+
+# uid -- unique identifier type
+#
+# Wraps a u64 in a dedicated type to prevent accidental mixing
+# with other numeric values while preserving ordering semantics.
+#
+# Instances are typically created by the unique_id effect.
+#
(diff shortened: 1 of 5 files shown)


modules/base/src/uid.fz
```diff
@@ -0,0 +1,52 @@
...
+# with other numeric values while preserving ordering semantics.
+#
+# Instances are typically created by the unique_id effect.
+#
+private:public uid(val u64) : property.orderable is
+
+
+  # Returns the underlying numeric identifier.
+  #
+  public as_u64 u64 => val
```
michaellilltokiwa: I think this is a leaky abstraction. I would not want to add this unless we have a clear reason to do so.

michaellilltokiwa:
> fix #6405
> 
> @michaellilltokiwa is this what you were thinking?
> 
> What's the preferred way to create new instances? Provide a type feature `type.new` or make the constructor public?

- Do you want to prevent inheritance from uid => `type.new`
- Do you need error handling before the constructor and want to return an outcome => `type.new`
- In case you want to allow inheritance you need to mark the constructor as public

maxteufel (approved):
I think using a `private:public` constructor and a type feature here is the way to go. Other than that, nice.

modules/base/src/uid.fz
```diff
@@ -0,0 +1,42 @@
...
+
+  # Returns the identifier as its decimal string representation.
+  #
+  public redef as_string String => val.as_string
+
+
+  # Defines a total ordering for uid values based on their
+  # underlying numeric representation.
+  #
+  public redef type.lteq(a, b uid.this) bool => a.val <= b.val
```
michaellilltokiwa: I think it is better to mark this as  `fixed`, to prevent accidental misuse.

modules/base/src/uid.fz
```diff
@@ -0,0 +1,42 @@
...
+# -----------------------------------------------------------------------
+
+# uid -- unique identifier type
+#
+# Wraps a u64 in a dedicated type to prevent accidental mixing
+# with other numeric values while preserving ordering semantics.
+#
+# Instances are typically created by the unique_id effect.
+#
+public uid(val u64) : property.orderable is
```
michaellilltokiwa, line 31: Can you add a comment that we may want to parametrize this in the future?
```suggestion
public uid(BT : property.orderable, val BT) : property.orderable is
```

simonvonhackewitz: What would be the benefit? It would result in several incompatible uid types, right?

michaellilltokiwa: Yes. Maybe your right, we don't need this for now...

PR #7159, opened 2026-06-09
ast: refine code `constraintAssignableFrom`

michaellilltokiwa:
Just made the code more readable for now.

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -677,19 +677,14 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
 
 
   /**
-   * Check if a type parameter actual can be assigned to a type parameter with
+   * Check if actual can be assigned to a type parameter with
    * constraint this.
    *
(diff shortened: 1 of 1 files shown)


src/dev/flang/ast/AbstractType.java
```diff
@@ -703,36 +698,65 @@ boolean constraintAssignableFrom(Context context, AbstractType actual)
 
     if (!result && !isGenericArgument())
       {
-        if (actual.isGenericArgument())
-          {
-            result = constraintAssignableFrom(context, actual.genericArgument().constraint(context));
-          }
-        else
+        // NYI: BUG: likely unsound
```
maxteufel: please add the reference to the issue that was mentioned in the old code here, and any possible other references that say how this could be possibly unsound.

michaellilltokiwa: I removed the mentioned issue since this was actually wrong/closed already but now added two different ones.

maxteufel (approved):

PR #7172, opened 2026-06-10
fe: suppress visibility reporting for call to outer ref

### LINKED ISSUE #7171: Too restrictive visibility reported for feature called in pre-condition.
Take this code that declares only `public` features
```
public muta is
  public x bool => true
  public new is
    public get unit
      pre
        x
    =>
```
nevertheless, the compiler complains about too restrictive visibily of a feature `muta.new._` which does not seem to be part of the source coude:
```
 > ./build/bin/fz -modules=base test_visi.fz 

/home/fridi/fuzion/fuzion/test_visi.fz:6:9: error 1: Called feature in precondition has more restrictive visibility than feature.
        x

To solve this, increase the visibility of 'muta.new._' or do not use this feature in the precondition.

one error.

```

michaellilltokiwa:
fixes #7171


src/dev/flang/fe/SourceModule.java
```diff
@@ -1952,13 +1952,15 @@ public void action(AbstractCall c)
...
-              && !(c.target() instanceof Current)
+              || (c.target() instanceof Current)
               // type param is known by caller
-              && !c.calledFeature().isTypeParameter()
+              || c.calledFeature().isTypeParameter()
+              // e.g. call to outerref is fine
+              || c.calledFeature().featureName().isInternal()
               // the called feature must be at least as visible as the feature.
-              && c.calledFeature().visibility().eraseTypeVisibility().ordinal() < f.visibility().eraseTypeVisibility().ordinal())
+              || c.calledFeature().visibility().eraseTypeVisibility().ordinal() >= f.visibility().eraseTypeVisibility().ordinal()))
```
maxteufel, line 1963: logical note: negating the entire chain, flipping AND to OR, and negating all inner statements. looks good!

maxteufel (approved):

src/dev/flang/fe/SourceModule.java
michaellilltokiwa: Yes, I used https://en.wikipedia.org/wiki/De_Morgan%27s_laws

PR #7176, opened 2026-06-10
base: add `atomic.update`

michaellilltokiwa:
to be able to update an atomic value a lambda that gets called with the respective old value until `compare_and_set` is successful

### DIFF
--- a/modules/base/src/concur/atomic.fz
+++ b/modules/base/src/concur/atomic.fz
@@ -166,7 +166,7 @@ is
   #
   # returns the (old_value, new_value)
   #
-  public update(compute_new T->T) (T,T) =>
+  public update(F type : T->T, compute_new F) (T,T) =>
     for current_value := read
         new_value     := compute_new current_value
     while !(compare_and_set current_value new_value)
--- a/modules/base/src/concur/atomic.fz
+++ b/modules/base/src/concur/atomic.fz
@@ -150,11 +150,7 @@ is
   public incr T ! atomic_access
     pre T : integer
   =>
-    for current_value := read
-        new_value := current_value+1
-    while !(compare_and_set current_value new_value)
-    else new_value
-
+    (update o->o+1).1
 
 
   # atomic decrement
@@ -162,10 +158,19 @@ is
   public decr T  ! atomic_access
     pre T : integer
   =>
+    (update o->o-1).1
+
+
+  # use `compute_new` to compute a new value of the atomic
+  # do this until a compare_and_set is successful.
+  #
+  # returns the (old_value, new_value)
+  #
+  public update(compute_new T->T) (T,T) =>
     for current_value := read
-        new_value := current_value-1
+        new_value     := compute_new current_value
     while !(compare_and_set current_value new_value)
-    else new_value
+    else (current_value, new_value)
 
 
   # create a new atomic of type T


maxteufel (approved):

modules/base/src/concur/atomic.fz
```diff
@@ -150,22 +150,27 @@ is
...
+    (update o->o-1).1
+
+
+  # use `compute_new` to compute a new value of the atomic
+  # do this until a compare_and_set is successful.
+  #
+  # returns the (old_value, new_value)
+  #
+  public update(compute_new T->T) (T,T) =>
     for current_value := read
```
fridis: Would be good to specialize this for `compute_new`, i.e., do this instead
```suggestion
  public update(F type : T->T, compute_new F) (T,T) =>
for current_value := read
```

fridis (approved):
good, just a small suggestion for performance. 

I plan to eventually add all the features from `atomic` to `mutate.new` and implement them to match the underlying mutate, i.e., no synchronization in `mutate`, use of a mutex in `blocking_mutate` and a new `atomic_mutate` that does the same as `concur.atomic` currently does.

modules/base/src/concur/atomic.fz
maxteufel: how does this help with performance?

fridis: A normal function type is a `ref` type, i.e., `compute_new` is a `ref` value and `compute_new.call` is a dynamically bound call. 

If we use a type parameter `F`, then `update` gets specialized for each different actual type. The type of `i->i+1` in a call `counter.update i->i+1` is a value type that is a child of `Function`. Since `update` gets specialized, that value does not need to get boxed to get a reference to a heap allocated instance. Furthermore, if the lambda does not access the closure —which is the case for `i->i+1`—, there is no need for an outer reference and the type is effectively a unit type, so the argument `compute_new` gets removed by the back-end. Finally, `compute_new.call` does not need dynamic binding, but is a direct call that can even be inlined easily, so the call overhead disappears completely. 

Originally, I thought that it would be good to have the compiler do this optimization automatically. I still think that should happen, but I also think that it would be good to not having to rely on this optimization. 

BTW, the same optimization works for any reference type, e.g., `bla(s String) => ...` could also benefit from being `bla(S type : String, s S) => ...` if the value passed to this function is a value type that inherits from `String`. 

What I am unsure about is what syntax to use and whether we should change the syntax in a way that discourages `ref` types. We already can write 
```
  public update(F type : T->T, compute_new F) (T,T) =>
```
or
```
  public update(compute_new F : T->T) (T,T) =>
```
or
```
  public update(compute_new _ : T->T) (T,T) =>
```
a possibility to simplify the syntax might be to allow
```
  public update(compute_new : T->T) (T,T) =>
```
but I am not sure if this would cause more confusion than benefit.

modules/base/src/concur/atomic.fz
```diff
@@ -150,22 +150,27 @@ is
...
   =>
+    (update o->o-1).1
+
+
+  # use `compute_new` to compute a new value of the atomic
+  # do this until a compare_and_set is successful.
+  #
+  # returns the (old_value, new_value)
+  #
+  public update(compute_new T->T) (T,T) =>
```
simonvonhackewitz:
```suggestion
  public update(F type : T->T, compute_new F) (T,T) =>
```

PR #7182, opened 2026-06-12
base: remove workaround for #7171 since that issue is fixed

fridis:
The visibility test was too strict and did not permit outer refs in the precondition.


maxteufel (approved):

maxteufel:
needs test rerecording

PR #7189, opened 2026-06-15
base: cleanup and additional functions for `Channel`

fridis:
I am trying to bring this closer to golang's channel functionality.

Apart from adding more detailed documentation, this adds the following features:

- added `Channel.prefix <-` as an alternative to `next`.

- added `Sending_Channel.is_closed`/`Channel.is_closed`

- added `Channel.as_list` that permits iteration of the elements

- added `channel(T type)` feature to create a channel. I would like to make `Channel` a template feature and return different implementations.

Also changed the following

- `Sending_Channel.infix <-`/`Channel.infix <-` no longer returns `bool`, but results in a precondition fault in case the Channel is closed.

What I would like to do next is marry this with `blocking_mutate`.

### DIFF
--- a/modules/base/src/concur/Sending_Channel.fz
+++ b/modules/base/src/concur/Sending_Channel.fz
@@ -25,6 +25,25 @@
 #
 public Sending_Channel(T type) ref is
 
-  public infix <- (to T) bool => abstract
+  # send a value to this Channel.
+  #
+  # The channel must be open. Sending a value to a closed channel will cause a fault
+  # (pre_fault if `safety` is enabled, otherwise `panic`).
+  #
+  # In case the Channel is full, the attempt to send data will block until the Channel
+  # is no longer full or closed.
+  #
+  public infix <- (to T) unit
+  pre
+    safety : !is_closed
+  =>
+    abstract
 
-  public close unit => abstract
+
+  public close unit
+  post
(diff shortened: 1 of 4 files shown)


michaellilltokiwa (approved):
Good, the test seems to need an update?

modules/base/src/concur/Sending_Channel.fz
```diff
@@ -25,6 +25,25 @@
...
+  #
+  # The channel must be open. Sending a value to a closed channel will cause a fault
+  # (pre_fault if `safety` is enabled, otherwise `panic`).
+  #
+  # In case the Channel is full, the attempt to send data will block until the Channel
+  # is no longer full or closed.
+  #
+  public infix <- (to T) unit
+  pre
+    safety : !is_closed
```
maxteufel, line 38: this feels like a correctness problem: if the channel is already closed when we are trying to send, this is banned by precondition. however, the comment reads as if something can be sent to a full channel, which blocks, and maybe by then the channel is closed. what are the consequences of this? this will not cause a fault.

fridis: This was addressed in the previous comment. What is not so nice is that there are two different errors, I have changed this in #7196 to result in `panic` in both cases.

PR #7190, opened 2026-06-15
base: add `mutate.dynamic_array`, a mutable array that can grow/shrink

### LINKED ISSUE #7027: Consider adding a flexible growing/shrinking mutable array and use it in `Binary_Heap_Queue`
> I am still very unhappy about this. `mutate.array` is a fundamental and supposedly very simple data structure. Adding functionality here to shrink the size and to store the minimum makes this more complex and less efficient for _all_ uses of `mutate.array`, just to make the implementation of one use, `Binary_heap_Queue` a bit more convenient. 
> 
> This is not how this should be done. 
> 
> Instead, if you really need this, create a flexible array or similar that uses `mutate.array` and add your favorite functionality there!
> 
> For now, you can use this for your assignment, but I would like this to be cleaned up eventually.

_Originally posted by @fridis in https://github.com/tokiwa-software/fuzion/pull/6873#discussion_r3110105531_

simonvonhackewitz:
fix #7027

There are still a few NYIs left; let me know if you have any thoughts on them.


modules/base/src/mutate/array2.fz
```diff
@@ -73,7 +73,7 @@ is
   #
   public redef as_array universe.array T =>
     check_and_replace
-    array (length0.as_i32 * length1.as_i32) i->data[i]
+    array T (length0.as_i32 * length1.as_i32) i->data[i]
```
michaellilltokiwa:
```suggestion
    array _ (length0.as_i32 * length1.as_i32) i->data[i]
```

michaellilltokiwa (commented):
Good, I would like to see labels on newly added NYI though

michaellilltokiwa (approved):

simonvonhackewitz:
superseded by #7249

PR #7193, opened 2026-06-16
realtime: add a very basic API for tasks

maxteufel:
Nothing major yet; but a barebones implementation of the basics of what I have in mind from the suggestions by @fridis. Please tell me if this is what you have in mind or a completely different direction.

### DIFF
--- a/modules/base/src/concur/task_pool.fz
+++ b/modules/base/src/concur/task_pool.fz
@@ -1,0 +1,54 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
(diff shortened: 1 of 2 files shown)


modules/base/src/concur/task_pool.fz
```diff
@@ -0,0 +1,54 @@
...
+  # returns false if the pool is already shutting down
+  #
+  public submit(P type : time.Period, t concur.task P) unit =>
+
+    nt := concur.threads.env.spawn ()->
+      now := time.nano.env.read
+      end_of_time := time.instant u64.max
+      _ := time.nano.env.run_periodic now t.period end_of_time t.code
+
+    set thrds := list nt thrds
```
michaellilltokiwa: I would prefer using an atomic here and then use, `thrds.write`

modules/base/src/realtime/task.fz
```diff
@@ -0,0 +1,29 @@
...
+#  Source code of Fuzion standard library feature concur.task
+#
+# -----------------------------------------------------------------------
+
+# task -- a piece of code to be executed periodically and concurrently
+# with other tasks.
+#
+public task(P type : time.Period,
+            module code Function unit,
+            module period P) is
```
fridis, line 29: I think `P` and `period` could be public. Eventually, I would like to have `period` split up into a `trigger` , which is one (or maybe several) event(s) that triggers a release, and `release parameters` that more like describe when this happens.  There might be events like a periodic timer that can be used to produce release parameters, but in general, this might not be the case.

fridis (approved):
Good start. 

Maybe we should have more low-level infrastructure in `concur` and have a new feature `universe.realtime` that provides a higher level of abstraction.

michaellilltokiwa (approved):

Issue #7177, opened 2026-06-10
test stdin fails sometimes


michaellilltokiwa:
```
for i in {1..100}; do make c -C build/tests/stdin/; done
```
```
make: Leaving directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
make: Entering directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
cat stdintest.fz.stdin | ../../bin/check_simple_example c "../../bin/fz " stdintest.fz || exit 1
RUN stdintest.fz using  c  backend PASSED.
make: Leaving directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
make: Entering directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
cat stdintest.fz.stdin | ../../bin/check_simple_example c "../../bin/fz " stdintest.fz || exit 1
RUN stdintest.fz using  c  backend PASSED.
make: Leaving directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
make: Entering directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
cat stdintest.fz.stdin | ../../bin/check_simple_example c "../../bin/fz " stdintest.fz || exit 1
RUN stdintest.fz using  c  backend PASSED.
make: Leaving directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
make: Entering directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
cat stdintest.fz.stdin | ../../bin/check_simple_example c "../../bin/fz " stdintest.fz || exit 1
RUN stdintest.fz using  c  backend PASSED.
make: Leaving directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
make: Entering directory '/home/sam/openvscode-server-fuzion/vscode-fuzion/fuzion/build/tests/stdin'
cat stdintest.fz.stdin | ../../bin/check_simple_example c "../../bin/fz " stdintest.fz || exit 1
RUN stdintest.fz using  c  backend *** FAILED out on stdintest.fz
7,8c7,8
< %s😀: %s😀
< %s😀: %s😀
---
[...]

PR #7200, opened 2026-06-17
base: change array length type to `i64` and use type parameter for index access

### LINKED ISSUE #6424: change array length type to `i64`
I really like the idea behind #6244, even though it can probably be improved (#6423) that make it possible to use different integer types to index an array. This should make it possible to change the array length type to 64 bits while leaving the default integer type as `i32`. 

What is a bit tricky: I would like to see code like
```
  bytes := array (u8 99) id
  shorts := array (i16 99) id
  int32s := array (i32 99) id
  longs := array (u64 20) x->(x << x)
```
To create values of type `array u8`, `array i64`, `array i32` and `array u64`, respectively. 

This should, however, be straightforward using type parameters as suggested in #6423:
```
public array(I idx, T type, length I, init I -> T) array T =>
  (array T).new length init 
``` 
Even though the array length can never be negative, I would not go to `u64` to allows index calculations like `n := j-i+1` to get the number of indices between `i` and `j` (inclusive) or things like `for i := a.length-1, i-1 while i>=0 do say a[i]`

simonvonhackewitz:
fix #6424

### DIFF
--- a/benchmarks/read_10mb_file.fz
+++ b/benchmarks/read_10mb_file.fz
@@ -19,7 +19,7 @@ read_10mb_file =>
     io.file.use file io.file.mode.append ()->
       sz := (io.file.stat file false).or_panic.size
       io.file.open.mmap 0 sz ()->
-        array u8 sz.as_i32 i->
+        array u8 _ sz.as_i32 i->
           io.file.mapped_buffer.env[i.as_i64]
 
   read_file2(file path) outcome (Sequence u8) =>
@@ -28,7 +28,7 @@ read_10mb_file =>
 
   f := path.of "/tmp/file"
 
-  a := array u8 1E7 x->65
+  a := array u8 _ 1E7 _->(u8 65)
   s := _ : String is
          public redef utf8 Sequence u8 => a
 
--- a/modules/base/src/handles.fz
+++ b/modules/base/src/handles.fz
@@ -164,7 +164,7 @@ is
   #
   update0 (
     # a handle created by `new`
-    x i32,
+    x i64,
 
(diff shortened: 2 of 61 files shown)


simonvonhackewitz:
test failure due to #7177

benchmarks/read_10mb_file.fz
```diff
@@ -28,7 +28,7 @@ read_10mb_file =>
 
   f := path.of "/tmp/file"
 
-  a := array u8 1E7 x->65
+  a := array u8 _ 1E7 _->(u8 65)
```
michaellilltokiwa, line 31: ~Seems like a regression that type inference does not work anymore?~

modules/base/src/handles.fz
```diff
@@ -183,11 +183,11 @@ is
 # shorthand for creating and installing an empty set of handles of given type.
 #
 public handles(T type, rr ()->unit) unit =>
-  (handles unit (array T 0 x->do) oneway_monad_mode.inst) ! rr
+  (handles unit (array T).empty oneway_monad_mode.inst) ! rr
```
michaellilltokiwa, line 186: Why did you change the semantics here?

michaellilltokiwa (commented):
Good work!
Only one minor question.

modules/base/src/handles.fz
simonvonhackewitz: Did I? I replaced an array of size 0 with an empty one. Also, I didn't really understand what the `do` is there.

michaellilltokiwa (approved):

PR #7206, opened 2026-06-17
be/c: posix, do not heap allocate `pthread_t`

### LINKED ISSUE #7029: `pthread_t` is just a thread id, shold not be heap allocated
This happens in `include/posix.c` where thread results and arguments for intrinsics are wrapped into heap allocated data. It should be possible to just cast to / from `void *` instead.

michaellilltokiwa:
fixes #7029


fridis (approved):
looks much better!

PR #7209, opened 2026-06-18
fe: fix looking up type for generic of inherits call

### LINKED ISSUE #37: Inner feature inheriting from generic outer feature results in endless loop or recursion in front end
This small example
```
> cat a.fz
a<T> is
  b : a<b> is
```
causes an endless loop or recursion in the front end:
```
 > timeout 3s ../build/bin/fz a || echo killed
killed
```

michaellilltokiwa:
fixes #37


src/dev/flang/fe/SourceModule.java
```diff
@@ -1215,7 +1215,9 @@ private List<FeatureAndOuter> lookup0(AbstractFeature outer, String name, Expr u
               {
                 if ((use == null || (hidden != featureVisible(use.pos()._sourceFile, v))) &&
                     !(use instanceof Call c && !c.isInheritanceCall() && v.isChoice()) &&
-                    (use == null || /* NYI: do we have to evaluate inScope for all possible outers? */ inScope(use, v)))
+                    (use == null || /* NYI: do we have to evaluate inScope for all possible outers? */ inScope(use, v)) &&
+                    // see #37 where this is necessary
+                    v != curOuter)
```
michaellilltokiwa, line 1220:
```suggestion
                    v != curOuter || use instanceof Call c && c.isInheritanceCall())
```

PR #7211, opened 2026-06-18
modules: do some missed string interpolation changes

michaellilltokiwa:


modules/terminal/src/terminal.fz
```diff
@@ -140,7 +140,7 @@ public terminal is
     public clear_line                           String => "2K"
     public erase_end_of_line                    String => "K"
     public clear_screen                         String => "2J"
-    public put_cursor(line, column i32)         String => "$(line);$(column)H" # or "{line};{column}f"
+    public put_cursor(line, column i32)         String => "$(line);$(column)H"
```
simonvonhackewitz: Why did you remove the comment?

michaellilltokiwa: It is outdated, just `$` does not work either, because of the `H` suffix.

simonvonhackewitz: Is it? I think the suffix `f` is an alternative to `H`, so the comment could just be updated to the new syntax.

michaellilltokiwa: Ah, sorry. I did not fully understand the comment, will update.

michaellilltokiwa:
```suggestion
    public put_cursor(line, column i32)         String => "$(line);$(column)H" # or "$(line);$(column)f"
```

simonvonhackewitz (approved):

PR #7214, opened 2026-06-18
ast: further simplify result type logic in `Call`/`AbstractCall`

michaellilltokiwa:


fridis (approved):
nice reduction by 50 lines!

PR #7228, opened 2026-06-22
fe: support auto_unwrap of match subjects and if conditions

### LINKED ISSUE #7227: support auto_unwrap of match subjects and if conditions
```
match mut (option String) nil
  nil => say "nil"
  String => say "string"

if mut bool false
  say "true"
else
  say "false"
```
currently:
```
/home/sam/playground/test.fz:1:7: error 1: 'match' subject type must be a choice type
match mut (option String) nil

Matched type: 'mutate.new (option String)', which is not a choice type


/home/sam/playground/test.fz:5:4: error 2: If condition must be assignable to type 'bool'
if mut bool false

Actual type is 'mutate.new bool'

2 errors.
```

michaellilltokiwa:
fixes #7227

This allows to do a match on e.g. a mutable variable without the need for explicit unwrapping:
```fuzion
x := mut (option i32) nil

# some code that may compute a result
x <- 42

match x
  nil => say "no result found"
  n i32 => say n
```

### DIFF
--- a/src/dev/flang/ast/Expr.java
+++ b/src/dev/flang/ast/Expr.java
@@ -711,6 +711,33 @@ public abstract class Expr extends ANY implements HasSourcePosition
 
 
   /**
+   * Do automatic unwrapping of features inheriting {@code unwrap}
+   * if unwrapped type is a choice.
+   *
+   * @param res the resolution instance
+   *
+   * @param context the source code context where this Expr is used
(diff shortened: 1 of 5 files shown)


fridis (approved):
Nice. 

Would be good to have a realistic small example in the description of the PR that could be added to the next news.

PR #7237, opened 2026-06-24
macOS: fix `_unlocked` versions not present on macOS

michaellilltokiwa:
[ci skip]

### DIFF
--- a/include/posix.c
+++ b/include/posix.c
@@ -87,6 +87,16 @@ static_assert(SIGALRM == 14, "signal definition different than expected");
 static_assert(SIGTERM == 15, "signal definition different than expected");
 static_assert(sizeof(pthread_t) <= sizeof(void *), "pthread_t must be smaller or equal to pointer size");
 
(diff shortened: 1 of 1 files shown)


maxteufel:
does this cause macOS file operations to be locked then?

michaellilltokiwa:
@maxteufel Yes, I expect so.

maxteufel:
not sure if that is a good idea then. At best, the locking is done twice (once inside Fuzion, once by the standard library). At worst, something bad happens because of an interaction we didn't think of.

michaellilltokiwa:
@maxteufel Hmm, this is why I added a `NYI: DOCUMENTATION`. I think it is fine for now to use the locked versions on macOS since on windows the behavior might be similar (not at all sure though). Caveat: when using the native/intrinsics in Fuzion we must not assume these are locked.

I'm open to better/different suggestions of course.

maxteufel (approved):

PR #7238, opened 2026-06-24
tests: add regression test for already fixed issue #4471

### LINKED ISSUE #4471: AoC: Type inference for `[]` could be improved
Ran into this during the [10 Dec Advent of Code](https://github.com/fridis/fuzion_aoc/blob/main/2024/10/part2.fz): This fails to infer the type of `[]` even though it could be determined done automatically
```
 > ./build/bin/fz -e 'say (s 10); s(n i32) Sequence i32 => n>0 ? container.set_of_ordered [n] : []'

command line:1:75: error 1: Failed to infer type of expression.
say (s 10); s(n i32) Sequence i32 => n>0 ? container.set_of_ordered [n] : []

Expression with unknown type: '[]'

one error.
```
so we have to give it explicitly
```
> ./build/bin/fz -e 'say (s 10); s(n i32) => n>0 ? container.set_of_ordered [n] : id (Sequence i32) []'
{10}
```
here is a similar example where type inference works nicely
```
> ./build/bin/fz -e 'say (s 10); s(n i32) => n>0 ? [n]++(s n-1) : []'
[10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
```
The following is a little harder and also does not work:
```
 > ./build/bin/fz -e 'say (s 10); s(n i32) => n>0 ? container.set_of_ordered [n] : []'

command line:1:62: error 1: Failed to infer type of expression.
say (s 10); s(n i32) => n>0 ? container.set_of_ordered [n] : []

Expression with unknown type: '[]'

one error.
```

michaellilltokiwa:
fixes #4471

The issue also mentions this as not working:
`> ./build/bin/fz -e 'say (s 10); s(n i32) => n>0 ? container.set_of_ordered [n] : []'`

But I think this should not work anyway. How should the compiler know how to box `set_of_ordered` and `array`?


maxteufel (approved):

fridis:
> The issue also mentions this as not working: `> ./build/bin/fz -e 'say (s 10); s(n i32) => n>0 ? container.set_of_ordered [n] : []'`
> 
> But I think this should not work anyway. How should the compiler know how to box `set_of_ordered` and `array`?

You are right, the inferred result type is usually ambiguous. We could just require the most specific `ref` type if there is just one such `ref` type, but this is maybe too tricky, maybe even dangerous since it would result in `Any` for something like `n>0 ? 42 : "neg"`, which is probably worse than an error.

PR #7244, opened 2026-06-24
base: make `close` on mutable variable require `is_exclusive`

fridis:
Otherwise, we have racy accesses to the `open` flag when used with `blocking_mutate`.


michaellilltokiwa:
I think this has to be done for the reading of open as well? We probably need a private field `_open` and and a public feature `open` that is protected by the same precondition as `close`.

maxteufel (approved):

PR #7247, opened 2026-06-24
base: add `codepoint.as_u8`, hide `String.type.minus_char` etc.

michaellilltokiwa:
I think the convenience functions are better not exposed...

### DIFF
--- a/modules/base/src/codepoint.fz
+++ b/modules/base/src/codepoint.fz
@@ -86,7 +86,7 @@ is
 
   # shortcut to get this codepoint as a u8
   #
-  public as_u8
+  public as_u8 u8
     pre codepoint.utf8_encoded_in_one_byte.contains val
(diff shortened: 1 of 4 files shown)


maxteufel (approved):

modules/base/src/codepoint.fz
```diff
@@ -84,6 +84,12 @@ is
   #
   public is_not_a_character bool => codepoint.not_a_character.contains val
 
+  # shortcut to get this codepoint as a u8
+  #
+  public as_u8 u8
+    pre codepoint.utf8_encoded_in_one_byte.contains val
```
fridis, line 90: The precondition should IMHO be 

```suggestion
    pre debug: val.fits_in_u8
```
Since codepoints 128 through 255 are not encoded in one byte.

PR #7249, opened 2026-06-24
base: add mutateble dynamic arrays, that can grow/shrink

### LINKED ISSUE #7027: Consider adding a flexible growing/shrinking mutable array and use it in `Binary_Heap_Queue`

simonvonhackewitz:
Alternative to approach to #7190
fix #7027

~~Requires #7256 to get fixed.~~ ✅️

The idea is to avoid storing an `i64` for the minimum length of the internal array in each instance.

Defines `Abstract_Dynamic_Array` that serves as an interface. It is implemented by `Dynamic_Array`, which always grows/shrinks.
Alternative behavior (like `Fixed_Capacity_Array`) can then be implemented by inheriting from `Dynamic_Array` and only redefining `may_grow` and `may_shrink`.

Currently, all of these are implemented as refs. ~~Ideally, they would be value types instead, if possible, to avoid heap allocations.~~
It might actually be better for them to be refs, so the array does not get copied when passed around, and modifications are made to the same instance. I think this is usually the expected behavior.


modules/base/src/mutate/array3.fz
```diff
@@ -71,7 +71,7 @@ is
   #
   public redef as_array universe.array T =>
     check_and_replace
-    array (length0 * length1 * length2) i->data[i]
+    array _ _ (length0 * length1 * length2) i->data[i]
```
michaellilltokiwa: I think I would prefer to fully qualify the call.
```suggestion
    universe.array (length0 * length1 * length2) i->data[i]
```

michaellilltokiwa (commented):
Looks good! Will look at it again tomorrow.

maxteufel (changes requested):
there could be a small test for the dynamic array

michaellilltokiwa (approved):

maxteufel (approved):

PR #7251, opened 2026-06-24
base: make Nullary inherit from `auto_unwrap`

michaellilltokiwa:
6634 needed to be rerecorded. It would be better if `mayUnwrapTarget` in `Call` would return false and to not unwrap the target if we can't find a feature for the calls name. Will create an issue for this.

### DIFF
--- a/tests/reg_issue6634/reg_issue6634.fz.expected_err
+++ b/tests/reg_issue6634/reg_issue6634.fz.expected_err
@@ -3,7 +3,7 @@
       (()-> for i:=1E5,i-°1 while i>0 do)*1E5
 -----------------------------------------^
 Feature not found: 'infix/infix_right *' (one argument)
-Target feature: 'ex.(λ(()-> for i:=1E5,i-°1 while i>0 do))'
+Target feature: 'unit'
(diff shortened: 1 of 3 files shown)


maxteufel (approved):

tests/reg_issue6634/reg_issue6634.fz.expected_err
```diff
@@ -3,7 +3,7 @@
       (()-> for i:=1E5,i-°1 while i>0 do)*1E5
 -----------------------------------------^
 Feature not found: 'infix/infix_right *' (one argument)
-Target feature: 'ex.(λ(()-> for i:=1E5,i-°1 while i>0 do))'
+Target feature: 'unit'
```
fridis, line 6: I think this change deserves an issue that probably applies to all `auto_unwrap` features. Ideally, this would list both, the original target feature (which is the lambda `ex.(λ(()-> for i:=1E5,i-°1 while i>0 do))` in this case) and the unwrapped variant `unit`. Listing only the unwrapped type may be strange. 
I have added #7257 for this.

fridis (approved):
Nice. Just not so happy about the change in the error message of reg_issue6634, but added new issue #7257 for that.

PR #7253, opened 2026-06-25
fe: enhance errors for loop termination condition resulting in non boolean

michaellilltokiwa:


src/dev/flang/ast/AbstractMatch.java
maxteufel, line 1: the `switch` here feels a bit verbose and i wonder if it would make sense to move this into the error message method instead.

maxteufel (approved):

PR #7254, opened 2026-06-25
fe: fix missing source position in "Illegal this-type" error message

michaellilltokiwa:


simonvonhackewitz (approved):

fridis (commented):
I think we need the same logic in Feature.java at
```
        checkLegalNativeResultType(res, resultTypePos(), resultType());
```
Not sure if we could move this logic into `resultTypePos()` and never return `null`.

maxteufel (approved):

PR #7260, opened 2026-06-25
ast: refactor `applyTypePars_` a bit

michaellilltokiwa:
also removed some potentially wrong code:
`result = result.applyTypePars(target.feature().outer().thisType());`


maxteufel:
can you elaborate more on why you believe the code mentioned is potentially wrong?

michaellilltokiwa:
I realize I should have copied more of the concerning code:
```
if (target.isThisType())
          {
            // see #659 for when this is relevant
            result = result.applyTypePars(target.feature().outer().thisType());
          }
```
This is done for this types only. I don't see why we need to applyTypePars for the outer of the feature of a this type.
I think this usually should just does nothing... So it has not been a problem yet.

maxteufel (approved):

PR #7263, opened 2026-06-26
fe: fix this-type not properly adjusted to cotype-target

### LINKED ISSUE #7256: `Illegal '.this' type: 'my_mut.this'`
```
public my_mut : mutate is

  public abstract_a (public T type) ref is

  public arr (public T type) ref : abstract_a T is
    public type.empty abstract_a T =>
      my_mut.this.env.arr T

foo unit =>
  m : my_mut is
  m ! ()->
    _ := (m.arr path).empty
```

Requires #7254 to get a source position
```
❯ fz /tmp/ex.fz | cat

/tmp/ex.fz:12:5: error 1: Illegal '.this' type: 'my_mut.this'
    _ := (m.arr path).empty
----^
To solve this, either
  - move the code inside of the feature the type refers to
or
  - change the type to a legal '.this' type.

one error.
```

michaellilltokiwa:
fixes #7256

```
public my_mut : mutate is

  public abstract_a (public T type) ref is

  public arr (public T type) ref : abstract_a T is
    public type.empty abstract_a T =>
      my_mut.this.env.arr T

foo is
  m : my_mut is
  m ! ()->
    say <| (m.arr path).empty

  a(M type : m, mm M) =>
    mm ! ()->
      say <| (M.arr path).empty
  a m

foo2 : foo is

_ := foo2
```

In the example type.empty has result type : `my_mut.this.abstract_a T`
For the result type of the call `(m.arr path).empty`,  `my_mut.this` needs to be replaced by `foo.this.m`.
This was not yet done, this patch changes this.


src/dev/flang/ast/Call.java
```diff
@@ -2801,6 +2830,13 @@ private void setResultType(Resolution res, Context context)
             _type = t2;
           });
       }
+
+    // if (_type != null)
```
fridis: Please add a comment why this code is here (or remove the code).

src/dev/flang/ast/Call.java
```diff
@@ -1439,14 +1439,43 @@ private AbstractType adjustResultType(Resolution res, Context context, AbstractT
...
+   */
+  private static AbstractType extractRelayedTypeFromCoType(Context context, AbstractType thisType, AbstractType coTypeTarget)
+  {
+    if (PRECONDITIONS) require
+      (thisType.isThisType(),
+       coTypeTarget.isCotypeType());
+
+    return coTypeTarget.generics().get(0).isThisTypeInCotype()
+      ? coTypeTarget.generics().get(0).genericArgument().outer().selfType().actualType(thisType, context)
+      : coTypeTarget.generics().get(0).actualType(thisType, context);
```
fridis:
```suggestion
    var g0 = coTypeTarget.generics().get(0);
    var g0a = g0.isThisTypeInCotype() ? g0.genericArgument().outer().selfType() 
                                      : g0;
    return g0a.actualType(thisType, context);
```
Also, would be great to add some examples what actually happens here, maybe using the example from the added regression test and where this actually makes a difference.

src/dev/flang/ast/ResolvedNormalType.java
```diff
@@ -166,7 +166,9 @@ protected ResolvedNormalType(List<AbstractType> g,
     if (POSTCONDITIONS) ensure
       (_feature == null /* artificial built in type */
        || feature().generics().sizeMatches(generics())
-       || generics().isEmpty() /* e.g. an incomplete type in a match case */);
+       || generics().isEmpty() /* e.g. an incomplete type in a match case */,
+        generics().stream().allMatch(x -> !x.isCotypeType()),
+        o == null || o.backingFeature().isUniverse() || _feature == null || !_feature.isCotype() || o.isGenericArgument() || o.isCotypeType());
```
fridis: I am confused about that this actually checks. A comment might be good.

fridis (commented):
Some comments.

src/dev/flang/ast/Call.java
```diff
@@ -1439,14 +1439,43 @@ private AbstractType adjustResultType(Resolution res, Context context, AbstractT
...
+   * @param coTypeTarget
+   * @return
+   */
+  private static AbstractType extractRelayedTypeFromCoType(Context context, AbstractType thisType, AbstractType coTypeTarget)
+  {
+    if (PRECONDITIONS) require
+      (thisType.isThisType(),
+       coTypeTarget.isCotypeType());
+
+    return coTypeTarget.generics().get(0).isThisTypeInCotype()
```
simonvonhackewitz: This fixed building when merging this PR into #7249 
```suggestion
    return coTypeTarget.isThisType()
      ? thisType
      : coTypeTarget.generics().get(0).isThisTypeInCotype()
```

src/dev/flang/ast/Call.java
```diff
@@ -1439,14 +1439,61 @@ private AbstractType adjustResultType(Resolution res, Context context, AbstractT
...
+
+
+  /**
+   * use RELAY#TYPE to find the actual type of
+   * the this-type.
+   *
+   * Example: where `my_mut.this` in the result type of `type.empty`
+   * needs to be replaced by `foo.this.m`.
+   *
+   * ```
```
simonvonhackewitz: I think the right way in JavaDoc is to use `@snippet` or `@code`:

```
{@snippet :
…
}
```
```
<pre>{@code
…
}</pre>
````

File: src/dev/flang/ast/Call.java [hunk 6]
```diff
@@ -1439,14 +1439,61 @@ private AbstractType adjustResultType(Resolution res, Context context, AbstractT
...
+      : t4;
+  }
+
+
+  /**
+   * use RELAY#TYPE to find the actual type of
+   * the this-type.
+   *
+   * Example: where `my_mut.this` in the result type of `type.empty`
+   * needs to be replaced by `foo.this.m`.
```
michaellilltokiwa:
```suggestion
   * Example: where {@code my_mut.this} in the result type of {@code type.empty}
   * needs to be replaced by {@code foo.this.m}.
```

src/dev/flang/ast/Call.java
```diff
@@ -1435,13 +1435,63 @@ private AbstractType getResultType0(Resolution res, Context context, boolean urg
...
+    if (PRECONDITIONS) require
+      (thisType.isThisType(),
+       coTypeTarget.isCotypeType());
+
+    return coTypeTarget.isThisType()
+      ? thisType
+      : coTypeTarget.generics().get(0).isRelayTypeInCotype()
+      ? coTypeTarget.generics().get(0).typeParameter().outer().selfType().actualType(thisType, context)
+      : coTypeTarget.generics().get(0).actualType(thisType, context);
   }
```
fridis: I still have some doubt that this works, having exactly one call to `outer()` is IMHO suspicious, what if we need `outer().outer()`?

Can you try this example: 
```
my_mut is

  abstract_a is
  inner is
    type.empty my_mut.this.abstract_a => panic "ignore"

foo is
  m : my_mut is
    inner2 : inner is
  _ := m.inner.type.empty
  i : m.inner
  _ := i.type.empty
  i2 : m.inner2
  _ := i2.type.empty
```

src/dev/flang/ast/Call.java
```diff
@@ -1435,13 +1435,63 @@ private AbstractType getResultType0(Resolution res, Context context, boolean urg
...
+  {
+    if (PRECONDITIONS) require
+      (thisType.isThisType(),
+       coTypeTarget.isCotypeType());
+
+    return coTypeTarget.isThisType()
+      ? thisType
+      : coTypeTarget.generics().get(0).isRelayTypeInCotype()
+      ? coTypeTarget.generics().get(0).typeParameter().outer().selfType().actualType(thisType, context)
+      : coTypeTarget.generics().get(0).actualType(thisType, context);
```
fridis: Just cosmetics: I suggest to try to avoid copied code like `coTypeTarget.generics().get(0)` (3x) or `.actualType(thisType, context)` (2x): 

```suggestion
    AbstractType result = thisType; 
    if (!coTypeTarget.isThisType())
      {
        var g0 = coTypeTarget.generics().get(0);
        var g0a = g0.isRelayTypeInCotype() ? g0.typeParameter().outer().selfType()
                                           : g0;
        result = g0a.actualType(thisType, context);
      }
    return result; 
```

fridis (changes requested):
Getting better, but I still have some doubt and suggest to check with another example. 

Also a cleanup suggestion.

src/dev/flang/ast/Call.java
michaellilltokiwa: I created an issue for this since I believe this is not currently possible anyway.

fridis (approved):
good.

PR #7266, opened 2026-06-26
base: add `Unbuffered Channel` for channels that do not buffer data

fridis:
Such a channel blocks on sending data unless a receiving thread is ready to take the data.
    
This works with a small state machine that records if there is a blocked receiver or a blocked sender and that unblocks these once the corresponding sender or receiver, respectively, has come.
    
Also added a new effect `channel_mutate T` that is used for unbuffered channel and that should be used for a buffered channel in the future as well.


modules/base/src/concur/Channel.fz
```diff
@@ -119,4 +119,266 @@ public Channel(public T type/*, buf_size i32*/) ref : Iterator T, Sending_Channe
...
+  if capacity > 0 then
+    # NYI: UNDER DEVELOPMENT use `container.Ring_Buffer`
+    Channel T
+  else
+    Unbuffered_Channel (channel_mutate T) T
+
+
+# mutate instance used for channels of given type
+#
+# We strictly do not need the type here, we could instead just
```
michaellilltokiwa: I think better:
- "Strictly, we do not need the type here..."
or
- "We do not  strictly need the type here..."

modules/base/src/concur/Channel.fz
```diff
@@ -119,4 +119,266 @@ public Channel(public T type/*, buf_size i32*/) ref : Iterator T, Sending_Channe
...
+        closed    => true
+
+    may_send =>
+      match unbuffered_channel_state.this
+        ready     => true
+        sending   => false
+        sent      => false
+        receiving => true
+        received     => false
+        closed    => true   # this does not sound logical, but it is: send will not block, but cauee a fault instead!
```
michaellilltokiwa:
```suggestion
        closed    => true   # this does not sound logical, but it is: send will not block, but cause a fault instead!
```

modules/base/src/concur/Channel.fz
```diff
@@ -119,4 +119,266 @@ public Channel(public T type/*, buf_size i32*/) ref : Iterator T, Sending_Channe
...
+
+  # ready -- channel state indicating that we are unused, no sender nor receiver
+  #
+  universe.ready is
+
+  # receiving -- channel state indicating that a thread is blocked receiving
+  #
+  universe.receiving is
+
+  # sent -- channel state indicating that data was sent to blocked receiving threda
```
michaellilltokiwa:
```suggestion
  # sent -- channel state indicating that data was sent to a blocked receiving thread
```

modules/base/src/concur/Channel.fz
```diff
@@ -119,4 +119,266 @@ public Channel(public T type/*, buf_size i32*/) ref : Iterator T, Sending_Channe
...
+        sent      => false
+        receiving => true
+        received     => false
+        closed    => true   # this does not sound logical, but it is: send will not block, but cauee a fault instead!
+
+    is_received =>
+      match unbuffered_channel_state.this
+        received     => true
+        *         => false
+    is_sent =>
```
michaellilltokiwa, line 283:
```suggestion
    
    is_sent =>
```

michaellilltokiwa (approved):

modules/base/src/concur/Channel.fz
fridis:
```suggestion
# Strictly, we do not need the type here, we could instead just
```

PR #7270, opened 2026-06-29
realtime: add abstraction for a scheduler

maxteufel:


modules/base/src/realtime/scheduler.fz
```diff
@@ -0,0 +1,52 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature realtime.scheduler
+#
+# -----------------------------------------------------------------------
+
+# scheduler primitive for realtime tasks
+#
+public scheduler (s Scheduler_Handler) : effect
```
fridis: The handler-pattern is no longer needed since we can instead just have a `ref` effect:
```suggestion
public Scheduler ref  : effect
```

fridis (changes requested):
Yes, a good start. 

It will get more interesting if you add release parameters, budgets, deadlines, etc.

michaellilltokiwa (approved):

fridis (approved):

PR #7271, opened 2026-06-29
fe: cleanup `Call.type`

michaellilltokiwa:
The control flow in Call.type was IMHO too complicated. This should improve this somewhat.


maxteufel (approved):
nice cleanup, a lot more readable

PR #7294, opened 2026-07-01
base: make `Mutable_Hash_Map` thread-safe

simonvonhackewitz:


michaellilltokiwa:
Good! I think some test would be nice as well. But this can be in another PR.

michaellilltokiwa (approved):

PR #7297, opened 2026-07-01
base: `Map.as_mutable_map` add default implemenation

michaellilltokiwa:


modules/base/src/container/hash_map.fz
```diff
@@ -206,12 +206,6 @@ is
       .map (.or_else (panic "filter failed"))
 
 
-  # NYI: implement this when #4642 is done
-  # create a mutable map from this
-  #
-  # public redef as_mutable_map container.Mutable_Map PK V =>
```
simonvonhackewitz, line 212: I think this should be redefined to return a `Mutable_Hash_Map` here (with precondition relaxed to hashable), because the keys are not guaranteed to be orderable, but they are hashable.

michaellilltokiwa: Good idea to return a `Mutable_Hash_Map` here but we can not relax the precondition for the constraints currently.

simonvonhackewitz: Why not?

michaellilltokiwa: I tried it now. It seems to work, I was wrong apparently...

modules/base/src/container/hash_map.fz
```diff
@@ -206,10 +206,14 @@ is
       .map (.or_else (panic "filter failed"))
 
 
-  # NYI: implement this when #4642 is done
   # create a mutable map from this
   #
-  # public redef as_mutable_map container.Mutable_Map PK V =>
+  public redef as_mutable_map(LM type : mutate) container.Mutable_Hash_Map LM HK V
+    pre else HK : property.hashable
```
fridis: `HK : property.hashable` is `true` since this is required in line 29 above. 

```suggestion
    pre else true
```
or
```suggestion
    pre else HK : property.hashable  # always true
```

modules/base/src/container/Map.fz
```diff
@@ -80,9 +80,10 @@ public Map(public K type : property.equatable, public V type) ref is
 
   # create a mutable map from this
   #
-  # NYI: maybe add default implementation when there is a naive implementation only requiring property.equatable
-  #
-  public as_mutable_map(LM type : mutate) container.Mutable_Map K V => abstract
+  public as_mutable_map(LM type : mutate) container.Mutable_Map K V
+    pre K : property.orderable
```
fridis: why not allow both

```suggestion
    pre K : property.orderable || K : property.hashable
```

modules/base/src/container/Map.fz
```diff
@@ -80,9 +80,10 @@ public Map(public K type : property.equatable, public V type) ref is
...
+  =>
+    (container.mutable_tree_map LM K V).from items
```
fridis: and then do 

```suggestion
   if      K : property.orderable then (container.mutable_tree_map LM K V).from items
   else if K : property.hashable  then (container.mutable_tree_map LM K V).from items
   else compile_time_error...
```

fridis (changes requested):
some minor suggestions.

modules/base/src/container/Map.fz
michaellilltokiwa: Yeah, that seems best...

simonvonhackewitz (approved):

PR #7300, opened 2026-07-02
docs: use term "Type Feature" instead of "Type Function"

simonvonhackewitz:


simonvonhackewitz:
Again I canceled the Linux tests since the only change was to the docs.

michaellilltokiwa (approved):

PR #7301, opened 2026-07-02
base: add `Sequence.collect`, a combined filter and map

michaellilltokiwa:


modules/base/src/Sequence.fz
```diff
@@ -1304,6 +1304,17 @@ is
   */
 
 
+  # filter and map in one pass.
+  #
+  # map items to a switch, return only "good" values
+  #
+  # NYI: UNDER DEVELOPMENT: can this somehow work?
+  #
+  # public collect(A type, B type, S type: switch A B, fn T -> S) Sequence A =>
```
fridis, line 1313: I would even like to have something like this: 
```
  public collect(A type,
                 B type,
                 S type : switch A B,
                 F type : T -> S,
                 fn F) Sequence A =>
```

modules/base/src/Sequence.fz
```diff
@@ -1304,6 +1304,17 @@ is
...
+  #
+  public collect(A type, B type, fn T -> switch A B) Sequence A =>
```
fridis, line 1315: As long as we cannot use `S type : switch A b`, we should IMHO use `option` as follows: 

```suggestion
  public collect(A type, fn T -> option A) Sequence A =>
```

modules/base/src/container/Mutable_Hash_Map.fz
```diff
@@ -246,9 +246,9 @@ is
   #
   public redef items Sequence (tuple HK V) =>
     LM.env.exclusive ()->
-      contents.get.as_array
-        .filter (o -> o ? tuple => true | * => false)
-        .map (x -> x ? t tuple => t | * => panic "filter failed" )
+      contents.get
+        .as_array
+        .collect (tuple HK V) nil (o -> o ? t tuple => t | * => nil)
```
fridis, line 251: would be good if the type parameters would not be needed here.

fridis (changes requested):
Nice, but this shows clearly that the restriction of not allowing `choice` as a type parameter is wrong. 

It is IMHO important that actual type parameters are inferred automatically.

modules/base/src/Sequence.fz
```diff
@@ -1304,6 +1304,17 @@ is
   */
 
 
+  # filter and map in one pass.
+  #
```
simonvonhackewitz, line 1308: I think this is not needed, as the comment below explains it better
```suggestion
```

simonvonhackewitz (commented):
I find the name `collect` a bit ambiguous, my first association was something like `fold`.
The only alternatives I found are `compact_map`, `filter_map` and `choose`. Not sure if they are any better.

fridis (commented):
I think that `flat_map` is actually more powerful and exists already: 
```
./build/bin/fz -e "say <| (i8 4 .. 100).flat_map (x->x*?x)"
[16, 25, 36, 49, 64, 81, 100, 121]
```
So, instead of `collect`, just use `flat_map` and you are good, works with `switch` as well. 

We might rename `flat_map` as `collect`. I am not particular happy about the name `flat_map`, is too Haskellish and IMHO sounds scary to newbys.  Looks as if F#  calls its version of `flat_map` `collect`.

michaellilltokiwa:
`flat_map id` is what I was looking for. I did not remember that switch inherits from Sequence...

PR #7303, opened 2026-07-02
base: remove `String.type.from_mutable_array`

michaellilltokiwa:
We use `FingerTree` for string concat

### DIFF
--- a/modules/base/src/String.fz
+++ b/modules/base/src/String.fz
@@ -970,33 +970,6 @@ public String ref : property.equatable, property.hashable, property.orderable is
     join elems "\n"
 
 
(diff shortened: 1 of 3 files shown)


simonvonhackewitz (approved):
Will get me a merge conflict in #7249, but OK.

simonvonhackewitz:
I might have merged this too eagerly. I think it does not make sense to something that is no longer in the base lib in the string test?

fridis:
> I might have merged this too eagerly. I think it does not make sense to something that is no longer in the base lib in the string test?

I agree. But for reg_issue1862, we should keep that function in.

PR #7307, opened 2026-07-02
build: add timeout to Jenkins pipelines

simonvonhackewitz:
[skip ci]

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,10 @@
 pipeline {
   agent any
+
+  options {
+    timeout(time: 30, unit: 'MINUTES')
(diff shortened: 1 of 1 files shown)


michaellilltokiwa (commented):
I don't think we need this in this pipeline. I think this never failed?

simonvonhackewitz:
~~Closed in favor of the “Global Build Timeout” setting.~~

Option "Global Build Timeout" comes from the plugin Build Timeout which [isn’t applicable to pipelines](https://plugins.jenkins.io/build-timeout/).

maxteufel (approved):

PR #7319, opened 2026-07-03
ast: resolve NYI comment hinting on a potential performance problem

michaellilltokiwa:


fridis (approved):
Is this actually called repeatedly such that caching makes a difference?

michaellilltokiwa:
@fridis I think this might very well be since this is used e.g. in IR.toStack which may be called for a features code for each of its specializations.

PR #7324, opened 2026-07-03
parser: fix sourceposition of ternary operator call

### LINKED ISSUE #4429: AoC: Strange subsequent error `Illegal forward or cyclic type inference`
This produces two errors:
```
 > ~/fuzion/work/build/bin/fz -e 'for x in true ? [0] : 1..2 do'

command line:1:15: error 1: Failed to infer type of expression.
for x in true ? [0] : 1..2 do

Expression with unknown type: ''


<source position not available>: error 2: Illegal forward or cyclic type inference
The definition of a field using ':=', or of a feature or function
using '=>' must not create cyclic type dependencies.
Referenced feature: 'loop' at <source position not available>:

2 errors.
```
the first error is ok, but it could be a little more explicit about what went wrong and not mentaion `''` as the culprit. 

bu the second error is fully strange, refering to `:=` and `=>` that are nowhere to be seen in the source code.

michaellilltokiwa:
fixes #4429

### DIFF
--- a/src/dev/flang/parser/Parser.java
+++ b/src/dev/flang/parser/Parser.java
@@ -1820,7 +1820,7 @@ exprNoColon : operatorExpr          // may not contain {@code :} unless enclosed
             matchOperator(":", "expr of the form >>a ? b : c<<");
             Expr g = operatorExpr();
             i.end();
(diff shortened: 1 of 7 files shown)


maxteufel (approved):

fridis (approved):
This is good, but it does not fix #4429 yet, the error is still very strange! So, please do not close that issue yet!

PR #7326, opened 2026-07-03
base: add `short_circuit` effect to allow early return with result

### LINKED ISSUE #6380: AoC: `effect.abort` usage cleanup
As a solution of [2025 Dec 8 part 2](https://github.com/fridis/fuzion_aoc/blob/main/2025/08/part2.fz), we could use `effect.abort` as follows:
```
  ae(r i64) : effect is

  part2a =>
    code =>
      _ := d.foldf (array n _->1) cl,dd->
        cl0 := dd.connect cl
        if cl0[0]=n then
          ae.abort (ae input[dd.i][0]*input[dd.j][0])
        cl0
      i64 -1  # no circuit possible

    ae.instate i64 (ae 0) code (.r.val)
```
But I am not fully happy with this, in particular

- providing the dummy argument `ae 0` when instating the effect
- it would be better to have a base lib effect to return a result via `abort`
- using `instate` as in the example makes it hard to provide the `code` as a lamba expression. 
- would be nice if instead of `ae.abort (ae result)`, we would do `ae.env.abort result` or similar. 
-

michaellilltokiwa:
fixes #6380


modules/base/src/eff/short_circuit.fz
```diff
@@ -0,0 +1,53 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion standard library feature eff.short_circuit
+#
+# -----------------------------------------------------------------------
+
+# short_circuit -- short_circuit is an effect to model a computation
+# that may return early from somewhere in the code
+#
```
fridis, line 26: Please add example code how this is used here.

modules/base/src/eff/short_circuit.fz
```diff
@@ -0,0 +1,62 @@
...
+# synthetic usage example, in a loop early return when result 42 is found:
+#
+#      say <| eff.short_circuit (option i32) .run ()->
+#        for i in 1..100 do
+#          if i=42
+#            (eff.short_circuit (option i32)).env.return i
+#        nil
```
fridis: Better create a specific type, e.g., 
```suggestion
#      sc : eff.short_circuit (option i32) is
#      say <| sc.run ()->
#        for i in 1..100 do
#          if i=42
#            sc.env.return i
#        nil
```
This will avoid confusion in case of nested use of this effect.

maxteufel (approved):

fridis (approved):
Just a small suggestion to improve the example given in the comment.

PR #7337, opened 2026-07-07
tests: add reg test for #4505

### LINKED ISSUE #4505: AoC: `Sequence.as_string sep` combined with `String.split s` has quadratic performance
The performance for a sequence of 2000 is 15s
```
> time ~/fuzion/clean/fuzion/build/bin/fz -e 'say (["1"," "].cycle.take 1000 .as_string "" .split " ").count'
501

real	0m6,232s
user	0m15,430s
sys	0m0,653s
```
while the performance for 4000 elements almost quadruples to 57s:
```
 > time ~/fuzion/clean/fuzion/build/bin/fz -e 'say (["1"," "].cycle.take 2000 .as_string "" .split " ").count'
1001

real	0m38,676s
user	0m57,146s
sys	0m1,348s
```

michaellilltokiwa:
fixes #4505


michaellilltokiwa:
failing due to merge incogongruence

maxteufel (approved):

tests/reg_issue4505/reg_issue4505.fz
```diff
@@ -0,0 +1,63 @@
...
+real	0m23,933s
+user	0m40,594s
+sys	0m0,995s
+```
+
+*/
+
+reg_issue4505 =>
+
+  say (["1"," "].cycle.take 10000 .as_string "" .split " ").count
```
simonvonhackewitz, line 63: The issue was about performance. This regression test does not seem to check the performance.

How about this
```suggestion
  initial := time.stopwatch (()->say (["1"," "].cycle.take 1000 .as_string "" .split " ").count)
  doubled := time.stopwatch (()->say (["1"," "].cycle.take 2000 .as_string "" .split " ").count)

  check doubled <= initial.times 2.5
```

fridis: This would result in occasional test failures, which is not good. 

However, this would be a candidate for a benchmark once we have a good infrastructure for benchmarks. But even then, there should be no failure, but just a graph that shows how the performance evolves over time.

michaellilltokiwa: @simonvonhackewitz It indirectly does check performance since this test would time out if performance was as bad as the issue states.

PR #7342, opened 2026-07-08
fe: Loop, mark successBlock/elseBlock as `final`

### LINKED ISSUE #2842: Loop.java has some hardly readable code.
Lots of fields are not final and replaced while the loop is constructed. This makes this code very hard to reason about.

michaellilltokiwa:
fixes #2842

Code could still be cleaner but I think it's okay for now.


maxteufel (approved):

maxteufel:
can be improved more but already a nice step into the right direction :+1:

PR #7343, opened 2026-07-08
cleanup: update "Source of ..." comments

michaellilltokiwa:
[ci skip]


maxteufel (approved):

simonvonhackewitz:
Do we really need these comments at all? What are their benefits?

fridis:
> Do we really need these comments at all? What are their benefits?

They are there to see what this is in case this information is not available, e.g, in a printout or log file or similar.

PR #7346, opened 2026-07-08
build: fix recording of tests that require input via stdin

simonvonhackewitz:
`check_simple_examples.fz` passes file `$test.stdin`, its content is then sent to stdin by `tokiwa.record_process` if the file exists

`record_simple_example.fz` lacks that behavior, so recording such tests stalls indefinitely

[skip ci]

### DIFF
--- a/bin/record_simple_example.fz
+++ b/bin/record_simple_example.fz
@@ -24,6 +24,10 @@
 
 ########### Extensions ###############
 
+# used for protecting spaces when splitting command string
+#
+protected_space := codepoint codepoint.utf8_encoded_in_four_bytes.as_list.last.or_panic
+
 # envir vars as map
 #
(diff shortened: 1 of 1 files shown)


simonvonhackewitz:
Should we put features like `String.execute` / `String.prefix !!` and `Sequence.execute`/`Sequence.prefix !` in module tokiwa? I think they are used in several places.

michaellilltokiwa (approved):
Nice! There are probably some more discrepancies between check_simple_example and record_simple_example

michaellilltokiwa:
> Should we put features like `String.execute` / `String.prefix !!` and `Sequence.execute`/`Sequence.prefix !` in module tokiwa? I think they are used in several places.

Yes, but the `prefix ...` feature names where a mistake. Also maybe find a better name than `execute`

PR #7347, opened 2026-07-08
base: convert Unique_Id and env.Vars effects to not use Handlers

michaellilltokiwa:
Ultimate goal is to get rid of all `_Handler` stuff in base and always just `ref`-effects


tests/i18n_default_test/i18n_default_test.fz
```diff
@@ -23,14 +23,14 @@
...
-    h : envir.Vars_Handler is
+    custom_env_vars : envir.Vars is
       public redef get(v String) option String =>
         if v = "LANG"
           "de_DE.UTF-8"
         else
           nil
 
-    (envir.vars h) ! ()->
+    custom_env_vars ! ()->
```
maxteufel: in my own tests i've had problems with using `ref` effects and the short-hand for instation. it might make sense to use the long form here, even if it does not lead to problems yet.

maxteufel (approved):

tests/i18n_default_test/i18n_default_test.fz
michaellilltokiwa: @maxteufel I actually does lead to problems here as it fails the test.
custom_env_vars is instated for type `i18n_default_test.this.custom_env_vars` but should be instated for type `envir.Vars`, so it will actually be used when doing `envir.Vars.env...`.

PR #7359, opened 2026-07-09
base: change nano to ref-effect removing `Nano_Time_Handler`

michaellilltokiwa:


modules/base/src/time/Clock.fz
```diff
@@ -69,7 +69,7 @@ public Clock ref : effect is
   # this will get instated automatically at startup
   #
   public redef fixed type.default_value option time.Clock =>
-    nano.default_value.bind (x -> id time.Clock x)
+    Nano.default_value.bind (x -> id time.Clock x)
```
maxteufel, line 72: is this good or should it rather be
```suggestion
    time.Nano.default_value.bind (x -> id time.Clock x)
```

maxteufel (approved):

modules/base/src/time/Clock.fz
michaellilltokiwa: Does not really matter since `default_value` is a type feature. So this should amount to the same outcome since `Nano` or `time.Nano` is used as the cotype.

PR #7365, opened 2026-07-09
base: make exit a ref-effect

### LINKED ISSUE #7357: exit effect's operations should be public
I currently cannot do
```
  exit.env.exit 1
```
I think it would be better to make `exit.exit` public and to remove `universe.exit(u8)` and  `universe.exit()` such that the it becomes clear when an exit is done that this is an effect. Furthermore, the exit effect should be a `ref`, i.e., to exit with error code one, we would need to do this:
```
  Exit.env.exit 1
```

michaellilltokiwa:
fixes #7357


maxteufel:
`exit` feels so fundamental that to me it is wrong to remove the shorthand `exit => Exit.env.exit`.

michaellilltokiwa:
@maxteufel `exit` does not feel that fundamental to me. It computes `void` which is arguably useless, from the perspective of computation it does not deliver any result. :shrug:

maxteufel (approved):

PR #7372, opened 2026-07-10
convert stat effect to ref-effect, removing Handler

michaellilltokiwa:


maxteufel (commented):
just as an idea here, what if `stat` and `lstat` were different handlers of the same effect?

michaellilltokiwa:
@maxteufel I split up the effect now, I think there may be cases where want to have custom versions of just lstat or stat, so different handlers may not suffice.

maxteufel (approved):

PR #7376, opened 2026-07-10
base: Add POSIX clocks and use them for timeouts in `blocking_mutate.wait`

fridis:
For now, `time.Monotonic` and `time.Wallclock` correspond to `CLOCK_MONOTONIC` and `CLOCK_REALTIME`. 

`blocking_mutate` was extended with a mechanism to add a timeout to `wait` and `exclusive_when` that can be used as follows: 
```
       mutate_type.env.exclusive ()->
         if mutate_type.env.timeout t
                           .wait condition
         then
           # condition was met before timeout
         else
           # timeout occured before condition became true
 ```
 or
```
       match mutate_type.env.timeout t
                            .exclusive_when condition ()->
                               # condition was met before timeout
         nil => # timeout occured before condition became true
         r T => # r was returned by code passed to `exclusive_when`.
```
 or
```
       mutate_type.env.timeout t
                      .exclusive_when condition ()->
                         # condition was met before timeout
                      .or_else ()->
                        # timeout occured before condition became true
```

### CODE REVIEW

File: modules/base/src/time/posix_clock.fz [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+#  Author: Fridtjof Siebert (siebert@tokiwa.software)
+#
+# -----------------------------------------------------------------------
+
+# time.posix_clock -- template for POSIX based clocks like Monotonic, etc.
+#
+# This is not meant to be used directly, but for other posix-based clocks
+# to inherit from.
+#
+public posix_clock : time.Clock is
```
maxteufel: maybe this can be `module:public`?

modules/base/src/concur/Thread.fz
```diff
@@ -126,42 +126,107 @@ is
...
+  #
+  # result is `true`  if this returns due to a call to `unpark`, it is `false` if there
+  # was no timeout.
+  #
+  module park_until(C type : time.Clock,
+                    F type : ()->unit,
+                    timeout   time.instant,
+                    condition ()->bool,
+                    code      F) bool =>
+    # NYI: UNDER DEVELOPMENT: We are locking both `mtx` and `c` here. This is redudant
```
maxteufel:
```suggestion
    # NYI: UNDER DEVELOPMENT: We are locking both `mtx` and `c` here. This is redundant
```

modules/base/src/concur/Thread.fz
```diff
@@ -126,42 +126,107 @@ is
...
 
-  # park this thread atomically after storing `condition` in `park_condition` and
+  # park this thread atomically after storing `condition` as `park_condition` and
   # executing `code`.
   #
   # The idea is that `code` may release a mutex in a way that ensures that this thread
   # is parked before it is released.
   #
-  module park(F type : ()->unit, condition ()->bool, code F)
+  module park_using_clock(C type : time.Clock, F type : ()->unit, condition ()->bool, code F)
```
maxteufel: name confusing since clock is not being "used" in a necessary way for the user.

modules/base/src/time/posix_clock.fz
fridis: yes.

modules/base/src/concur/Thread.fz
fridis: will change this (at all 3 occurrences...)

modules/base/src/concur/Thread.fz
fridis: I will just rename this back to `park` and remove the new `park`.

modules/base/src/concur/blocking_mutate.fz
```diff
@@ -144,3 +216,16 @@ public blocking_mutate : mutate is
...
     mtx.destroy
+
+
+# blocking_mutate -- a multi-thread variant mutate using time.Monotonic
+#
+# This mutates ensure exclusive access to shared memory via using locks and
+# blocking.
+#
+# Accidental race accesses are not permitted by this mutate, i.e., accessing
+# mutable elements, e.g, via `get` or `put` on a mutable var `mutate.new T`,
```
```suggestion
# mutable elements, e.g. via `get` or `put` on a mutable var `mutate.new T`,
```

modules/base/src/concur/Thread.fz
```diff
@@ -126,42 +126,102 @@ is
...
   # executing `code`.
   #
   # The idea is that `code` may release a mutex in a way that ensures that this thread
   # is parked before it is released.
   #
   module park(F type : ()->unit, condition ()->bool, code F)
   =>
-    cond.synchronized ()->
+    # NYI: UNDER DEVELOPMENT: We are locking both `mtx` and `c` here. This is redundant
+    # since `cond C` is based on `mtx`.  However, this does not seem to be the case for the
```
michaellilltokiwa, line 159: I don't understand the comment, I don't see the connection between `mtx` and `c`, maybe I'm missing something?

modules/base/src/posix.fz
```diff
@@ -0,0 +1,57 @@
...
+#
+public posix is
+
+
+  # a clock identifier
+  #
+  public clockid is
+
+
+  # the wall clock (NOTE: This is usually not useful for real-tiem code)
```
michaellilltokiwa:
```suggestion
  # the wall clock (NOTE: This is usually not useful for real-time code)
```

modules/base/src/concur/Thread.fz
```diff
@@ -126,42 +126,102 @@ is
...
-  # thread in park mode.
+  # thread in park mode using the given clock for timeouts
   #
-  cond := concur.sync.condition.new.or_else (panic "failed to create condition variable")
+  cond(C type : time.Clock) => (A1_park_condition C).env.cond
 
-  # is this thread parked?
+
+  # is this thread parked? This is an option of a tuple containing the condition
+  # variable and a Nullaray function that checks the condition that must hold for
```
michaellilltokiwa:
```suggestion
  # variable and a Nullary function that checks the condition that must hold for
```

modules/base/src/concur/Thread.fz
```diff
@@ -198,3 +251,51 @@ is
...
+#
+# NYI: CLEANUP: This could be just a field in thread as follows
+#
+#    mtx := concur.sync.mtx_init.or_else (panic "failed to create mutex")
+#
+# but there is currently no easy way to access this field while `concur.Threads.env.current`
+# cannot be used becuse `concur.Threads` is not yet instated.
+#
+# NYI: BUG: #7400 This must be instated before `concur.Threads`. Currently, the order
+# is not be dependency, but alphabetically, so we add prefix 'A_' as a workaround.
```
michaellilltokiwa: maybe like this?
```suggestion
# is sorted by dependency, but alphabetically, so we add prefix 'A_' as a workaround.
```

modules/base/src/concur/Thread.fz
fridis: A condition variable is connected to a mutex, `c.synchronized` uses `mtx.synchronized` (see code in condition.fz). The mutex is used for exclusive access, the condition variable to signal that a condition has occurred. 

So this code is locking `mtx` twice. however, I did not get this to work when locking only on `mtx`, which I think should be sufficient. I would like to revisit this and remove the double locking if possible. But this most likely requires some cleanup unrelated to this PR like removing `condition.synchronized` and using `mtx.synchronized` directly instead, which requires moving the `is_locked` field as well and maybe some more changes...

maxteufel (approved):

PR #7377, opened 2026-07-10
fix occasional "process not in set of running processes"

michaellilltokiwa:
running_processes was just overwritten, we must update

### DIFF
--- a/modules/base/src/os/Process_effect.fz
+++ b/modules/base/src/os/Process_effect.fz
@@ -48,7 +48,10 @@ public Process ref : effect is
   # this will get instated automatically at startup
   #
   public redef fixed type.default_value option os.Process =>
-    _ : os.Process is
+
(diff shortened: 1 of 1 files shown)


maxteufel (approved):

modules/base/src/os/Process_effect.fz
```diff
@@ -48,7 +48,10 @@ public Process ref : effect is
   # this will get instated automatically at startup
   #
   public redef fixed type.default_value option os.Process =>
-    _ : os.Process is
+
+    # this needs to be a ref since it contains an atomic field
+    #
+    P_ref ref : os.Process is
```
maxteufel: not allowed (type in type feature)

maxteufel (changes requested):

maxteufel (approved):

PR #7382, opened 2026-07-13
base: `io.dir.read` effect change to ref-style from _Handler style

michaellilltokiwa:
~seems to trigger a JVM-be-bug currently~

Added workaround for the JVM-bug


maxteufel:
workaround should be removed since bugfix is merged

maxteufel (approved):

PR #7384, opened 2026-07-13
web: `parse_url`, return proper error if port value exceeds 16-bits

michaellilltokiwa:


modules/web/src/web/shared.fz
```diff
@@ -109,7 +108,11 @@ module parse_url(url String) outcome web.parsed_url
               p_str String =>
                 match p_str.parse_u32
                   e error => e
-                  p u32 => (host_and_port_seq[0], id (option u16) p.as_u16, tpl.1)
+                  p u32 =>
+                    if !p.fits_in_u16
```
michaellilltokiwa:
```suggestion
                    if p.fits_in_u16
```

maxteufel (approved):

PR #7385, opened 2026-07-13
base: remove outdated NYI

michaellilltokiwa:


modules/base/src/fuzion/sys/internal_array.fz
```diff
@@ -62,15 +62,10 @@ module internal_array(T type, module data Array T, module length i32) is
...
 
-  # NYI: UNDER DEVELOPMENT: change to, does not work yet in jvm backend Code size of bytecode created for... is 68325, the maximum allowed length is 65535.
-
-  # module set [ ] (i idx, o T) unit
   module set [ ] (i i32, o T) unit
     pre
       safety  : 0 ≤ i < length
       debug 2 : {ensure_not_frozen; true}
-  # post   NYI
-    # debug: array.this[i] == o
```
simonvonhackewitz, line 73: Why did you remove the postcondition?

michaellilltokiwa: I don't see how this makes sense, an equality comparison for arbitrary values?

simonvonhackewitz: It's not arbitrary, it checks that element at index i is actually o, after storing o at index i.

michaellilltokiwa: I meant the type of the value could be anything, but we don't have an equality comparison for arbitrary values and never will, I think. So this check is not really possible to implement.

maxteufel (approved):

modules/base/src/fuzion/sys/internal_array.fz
simonvonhackewitz: 🙈

PR #7388, opened 2026-07-13
base: move `path` to `io.path`

michaellilltokiwa:


maxteufel:
if anything, to me `path` feels more like it belongs to `os`. but i don't mind the top-level position for now.

michaellilltokiwa:
@maxteufel Hmm, yes. I'm indifferent about `os.path` vs `io.path`.
@tokiwa-software/developers Your opinions?

simonvonhackewitz:
I'd lean more toward “io.” Is it currently heavily dependent on the operating system? If so, maybe we should make it more general.

maxteufel (approved):

PR #7392, opened 2026-07-13
check_simple_example: pass cpu_time_limit to `record_process`

michaellilltokiwa:


bin/check_simple_example.fz
```diff
@@ -212,7 +212,7 @@ else
     # stack_size_limit, cpu_time_limit
 
     bt := concur.threads.env.spawn ()->
-      time.Nano.env.sleep <| time.duration.s cpu_time_limit
+      time.Nano.env.sleep <| cpu_time_limit
```
simonvonhackewitz:
```suggestion
      time.Nano.env.sleep cpu_time_limit
```

bin/check_simple_example.fz
```diff
@@ -127,8 +127,8 @@ String.execute =>
 
 
 # limit cpu time and stack size for executing test
-cpu_time_limit   := u64 600  # NYI: UNDER DEVELOPMENT: we should decrease this value
-stack_size_limit := 1024
+cpu_time_limit   := time.duration.min 10  # NYI: UNDER DEVELOPMENT: we should decrease this value
+stack_size_limit := units.data_size.b 1024
```
simonvonhackewitz: Where is the stack size limit set? I think it is not done using this field.

bin/check_simple_example.fz
simonvonhackewitz:
```suggestion
    say "$("*** CANCELLED:".bold.red) test $file exceeded stack size limit of $stack_size_limit"
```

bin/check_simple_example.fz
michaellilltokiwa: It is not yet set. Missing an NYI: UNDER DEVELOPMENT:

simonvonhackewitz: like this?
```suggestion
stack_size_limit := units.data_size.b 1024 # NYI: UNDER DEVELOPMENT: stack limit is not yet set
```

simonvonhackewitz (approved):

Issue #7187, opened 2026-06-15
`sum` called on non-numeric `Sequence` produces error that lacks useful reference to source position


fridis:
Try this
```
 > ~/fuzion/fuzion/build/bin/fz -e 'say [unit].sum'

{base.fum}/Sequence/statistic.fz:97:5: error 1: Actual type parameter 'unit' does not satisfy constraint 'numeric'
    T : numeric


one error.
```
It would be great if this would include a reference to `[unit]` such that one knows where is the actual problem in the source code.

Nimux2:
I'm currently working on this.

fridis:
> I'm currently working on this.

I think this might be hard, this source position is not easily available when that error is produced.

PR #7394, opened 2026-07-14
Improved error messages to better point to source

Nimux2:
Second attempt at solving issue #7187.

For comparison with fz -e 'say [unit].sum' : 
Before : 
```

{base.fum}/Sequence/statistic.fz:97:5: error 1: Actual type parameter 'unit' does not satisfy constraint 'numeric'
    T : numeric


one error.
```

After : 
```

{base.fum}/Sequence/statistic.fz:97:5: error 1: Actual type parameter 'unit' does not satisfy constraint 'numeric'
    T : numeric

Constraint defined at: {base.fum}/unit.fz:84:8:
public unit : property.orderable, property.hashable is

    T : numeric


one error.
```

### DIFF
--- a/src/dev/flang/fuir/FuirErrors.java
+++ b/src/dev/flang/fuir/FuirErrors.java
@@ -116,10 +116,42 @@ public class FuirErrors extends AstErrors
   }
 
 
-  public static void unmetTypeContraint(SourcePosition pos, AbstractType tp, Clazz constraint)
-  {
-    error(pos, "Actual type parameter " + s(tp) + " does not satisfy constraint " + s(constraint._type), "");
-  }
+    public static void unmetTypeContraint(SourcePosition pos, AbstractType tp, Clazz constraint)
+    {
+        // Build the error message
+        StringBuilder fullMsg = new StringBuilder();
+        fullMsg.append("Actual type parameter '").append(s(tp)).append("' does not satisfy constraint '")
+            .append(s(constraint._type)).append("'");
+        
+        // Show the source code with underlining at the error position
+        if (!pos.isBuiltIn())
+        {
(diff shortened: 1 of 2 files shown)


src/dev/flang/fuir/FuirErrors.java
```diff
@@ -116,10 +116,49 @@ public static void abstractFeatureNotImplemented(AbstractFeature featureThatDoes
...
+        {
+            // If it's the same position, just mention it briefly
+            fullMsg.append("\nConstraint defined at this location");
+        }
+        
+        error(pos, fullMsg.toString(), "");
+    }
+
+    // Helper to get source line
+    private static String getSourceLine(SourcePosition pos)
```
michaellilltokiwa: This does not seem to be used (anymore)?

Nimux2: I must have forgotten to remove it after the last changes. I'll quickly push a fix.

maxteufel:
The output that we imagine should look more like this:
```
$ fz -noANSI -e 'say [unit].sum'

command line:1:6: error 1: Actual type parameter 'unit' does not satisfy constraint 'numeric'
say [unit].sum
-----^^^^

one error.
```

michaellilltokiwa:
@Nimux2 This is a hard issue, I tried solving this a few days ago and failed. Good that you tried anyway but maybe it is not worth it currently to try and solve it.

Nimux2:
> @Nimux2 This is a hard issue, I tried solving this a few days ago and failed. Good that you tried anyway but maybe it is not worth it currently to try and solve it.

I think you're right. I'll shelve this for the time being.

PR #7395, opened 2026-07-14
fe: remove restriction constraint must not be choice

michaellilltokiwa:


fridis (commented):
This is still a draft, but I comment anyway: 

I think this makes sense, the reason for the restriction was probably just that it used to be impossible to inherit from a choice feature, which made using `choice` as a type constraints pretty useless.

maxteufel (approved):

PR #7397, opened 2026-07-14
fe: add effect `native_access` to denote access of native code

michaellilltokiwa:
Idea is that `fz -effects` shows if applications uses any native code.


maxteufel (approved):
nice!

PR #7398, opened 2026-07-14
fe: fix type pars not handed down via targets of inheritance calls

### LINKED ISSUE #7436: type par not properly replaced leading to require condition in clazz constructor
```
buffered_ (LM type : mutate) is
  Writer_ ref : effect is
    buffer := container.Ring_Buffer u8 LM 10
    a =>
      say buffer

lm : mutate is

std_in_writer(LM type : mutate) : (buffered_ LM).Writer_ is

b (LM type : mutate) =>

  (buffered_ LM).Writer_.instate _ (std_in_writer LM) ()->
    (buffered_ LM).Writer_.env.a

lm ! ()->
  b lm
```

michaellilltokiwa:
This is/was an issue that came up in #7355, where type pars where not handed down properly like in:
`writer (desc File_Descriptor) : io.buffered file_mutate .Writer is`

fixes #7436

### DIFF
--- a/src/dev/flang/ast/AbstractFeature.java
+++ b/src/dev/flang/ast/AbstractFeature.java
@@ -1484,6 +1484,7 @@ public abstract class AbstractFeature extends Expr implements Comparable<Abstrac
    */
   private static List<AbstractType> handDownListThroughInheritsCall(List<AbstractType> l, int select, AbstractCall c)
   {
+    // NYI: BUG: it is not really sufficient to only do this for targets that are calls
     return (c.target() instanceof AbstractCall ac
         ? handDownListThroughInheritsCall(l, ac.select(), ac)
         : l)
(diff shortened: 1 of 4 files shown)


maxteufel (approved):

fridis (commented):
Is this pure cleanup or a semantic change? If it is the latter, is there an example or test that shows the difference? If so, it would be good to add a test!

michaellilltokiwa:
@fridis I found an example (in #7436) outside of base that currently crashes the compiler. Also added a test that this change fixes the issue.

fridis (approved):
Looks good for now. 

It would be good to work on tests inheriting with target being a match like 
```
  x(U type, i i32) is
    y is
      z(a U) is
        say "$a type $U i $i"
  q(T type, v T) : (
    match envir.Args.env.count>3
      true  => x T 1
      false => x T 2
    ).y.z v is
  _ := q "bla"
```

PR #7402, opened 2026-07-15
Move helper features `String`/`Sequence.execute` from scripts to module `tokiwa`

### LINKED ISSUE #7348: Move features `String.execute` and `Sequence.execute` into module tokiwa
These features are used in several scripts in `bin/`.

>> Should we put features like `String.execute` / `String.prefix !!` and `Sequence.execute`/`Sequence.prefix !` in module tokiwa? I think they are used in several places.
>
>Yes, but the `prefix ...` feature names where a mistake. Also maybe find a better name than `execute`

_Originally posted by @michaellilltokiwa in https://github.com/tokiwa-software/fuzion/issues/7346#issuecomment-4914295847_

simonvonhackewitz:
fix #7348

`check_simple_example` got extremely slow. Running `make -C build/tests/hello` (witch check_simple_example already compiled) takes 15 MINUTES instead of like 12 seconds. But it seems to work.

Running it in jvm got also slower, though I think not that extreme.
```
fz -debug=0 -modules=terminal,tokiwa bin/check_simple_example.fz jvm "fz" tests/hello/HelloWorld.fz
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,7 @@ $(BUILD_DIR)/%.md: $(FZ_SRC)/%.md
 
 $(FUZION_EBNF): $(FUZION_BASE) $(FZ_SRC)/bin/ebnf.fz
 	mkdir -p $(@D)
-	$(FZ) $(FZ_SRC)/bin/ebnf.fz $(JAVA_FILES_PARSER) > $@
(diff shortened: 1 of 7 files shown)


michaellilltokiwa:
problem is likely that `tokiwa.protected_space` is expensive to calculate and executed frequently while before it was only executed once.

maxteufel (approved):

PR #7404, opened 2026-07-15
cleanup: improve comments, simplify code, fix typos and formatting

simonvonhackewitz:


tests/simple.mk
maxteufel, line 1: why the changes in this file?

maxteufel (approved):
Looks good except mentioned.

tests/simple.mk
simonvonhackewitz: So that the output aligns nicely on the terminal.

maxteufel: okay, but that doesn't help with the alignment of `effect` "backend".

simonvonhackewitz: yes, but that is not included in target `all`, usually just `jvm`, `c` and `int` are run after each other

PR #7413, opened 2026-07-20
be/jvm: callNative, store result of arguments expressions in locals

### LINKED ISSUE #7387: require condition in `ClassFile.ByteCodeSizeEstimate.setLabel`
see changes in #7382

michaellilltokiwa:
fixes the bug that came up in:
https://github.com/tokiwa-software/fuzion/actions/runs/29237320990/job/86774908481

`fzE_dir_read io.dir.open.env.dd internal_array.data`

Specifically the expression `io.dir.open.env.dd` triggered the problem since a Labels positions was found to be already set when it was expected to not yet be set.

fixes #7387

### DIFF
--- a/src/dev/flang/be/jvm/CodeGen.java
+++ b/src/dev/flang/be/jvm/CodeGen.java
@@ -668,15 +668,18 @@ class CodeGen
       + ")"
       + nativeResultTypeDescriptor(rt).descriptor();
 
+    var argSlots = new List<Integer>();
     var localSlotsOfMemorySegments = new List<Integer>();
-    return Expr
-          .getstatic(_names.javaClass(cc),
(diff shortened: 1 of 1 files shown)


fridis (approved):
If I understand this correctly, the problem is that code for array arguments was created twice, once when converting them to memory segments, and then again when copying the memory segment back to the arguments. 

This patch avoids this by evaluating the arguments once and storing them into local variables that are used for conversion to and from memory segments. 

If this is correct, this explanation would have helped me at lot to understand this PR...:-). 

Anyway, looks good provided that my understanding is correct.

michaellilltokiwa:
@fridis Yes, this correct. Sorry for not explaining more succinctly.

PR #7416, opened 2026-07-20
base: effect instation, fluent interface, instate effects based on type

michaellilltokiwa:
depends on #7438

### DIFF
--- a/modules/base/src/effect.fz
+++ b/modules/base/src/effect.fz
@@ -232,15 +232,16 @@ public effect is
   # fluent interface for using multiple effects together
   #
   public and(R type, E type : effect, e Lazy E) Functional R =>
-    as_functional R
-      .compose (e.as_functional R)
+    as_functional effect.this R
+      .compose (e.as_functional E R)
 
 
   # this effect reinterpreted as a Functional
   # which allows composition of effects
   #
-  public as_functional(R type) Functional R =>
-    code -> instate_self code
+  public as_functional(E type : effect.this, R type) Functional R =>
+    code ->
+      E.instate _ effect.this code
 
 
   # has an effect of this type been instated?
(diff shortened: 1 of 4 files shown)


modules/base/src/effect.fz
```diff
@@ -232,15 +232,16 @@ public effect is
...
+    as_functional effect.this R
+      .compose (e.as_functional E R)
 
 
   # this effect reinterpreted as a Functional
   # which allows composition of effects
   #
-  public as_functional(R type) Functional R =>
-    code -> instate_self code
+  public as_functional(E type : effect.this, R type) Functional R =>
```
maxteufel, line 242: this is not marked as `fixed` unlike implied by the pull request title

michaellilltokiwa: @maxteufel Thanks for pointing this out. The title is out of date :frowning_face:

maxteufel (approved):

src/dev/flang/ast/AbstractType.java
```diff
@@ -640,6 +640,11 @@ else if (!actual.isRef())
             result = isAssignableFrom(actual.asRef(true), context, false, allowTagging, assignableTo);
           }
       }
+    // NYI: UNDER DEVELOPMENT: check if sound!
+    if (result.no() && target_type.selfOrConstraint(context).isThisType() && target_type.selfOrConstraint(context).compareTo(actual_type) == 0)
+      {
+        result = YesNo.yes;
+      }
```
fridis, line 647: This seems wrong to me: if the target_type is a type parameter with constraint `E type : x.this` and `actual_type` is `x.this` and we have types `x`, `y : x` and `z : y` we might have `E` being `z` and `x.this` being `y`, which are incompatible value types:
```
  x is
    f(E type : x.this) E => x.this
  y : x is
  z : y is
  say (type_of (y.f z))
```
where the result type of `y.f z` is `z` but the returned value is of type `y` (and vice-versa, the result type of `z.f y` is`y` but the returned value is of type `z`.

fridis (commented):
I fear this is not sound...

PR #7418, opened 2026-07-21
Edited Jitter to make use of `task` and `task_pool`

### LINKED ISSUE #7440: `time.Clock.run_periodic` does not handle an infinite period
The precondition allows for an infinite period and the comment documents this argument as well defined. However, I don't see in the code where this is handled. `p.as_duration` is not allowed to be called by its precondition if `p.is_infinity` is true.

Nimux2:
Now Jitter offloads most of the thread-related work through using the new API (task and task_pool). Task_pool was also slightly modified to give the option of running a given task only once, instead of looping it.


modules/base/src/concur/task_pool.fz
```diff
@@ -41,6 +41,16 @@ is
...
+  # submit a task to the thread pool and run it once
+  #
+  public submit_once(P type : time.period, t realtime.task P) unit ! atomic_access =>
+
+    nt := concur.Threads.env.spawn ()->
+      _ := t.code.call
+
+    thrds.write (list nt thrds)
+
+
```
maxteufel: this is not necessary. as per the API docs for `run_periodic`, using an infinite period will execute the code exactly once or not at all (depending on the parameters of `run_periodic`, in our case, exactly once).

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,22 @@ jitter_threaded =>
...
-  consume_and_display empty_acc
-  say "Dedicated thread completed."
+  say "Starting jitter test using task_pool..."
+  result_ch := concur.Channel (tuple time.duration time.instant time.instant bool)
+
+  jitter_worker =>
+    test_jitter time.Nano time.Nano result_ch
+    result_ch.close
+
+  jitter_task := realtime.task time.duration jitter_worker period_default
```
maxteufel: there should be a comment here that due to lack of implementation in the API, this does not set any priority or scheduling policy

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,22 @@ jitter_threaded =>
...
+
+  _ := concur.task_pool ! ()->
```
maxteufel: use `realtime.Scheduler.instate` with a `concur.task_pool` type parameter instead.

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,22 @@ jitter_threaded =>
...
+    pool := concur.task_pool.from_env
+    _ := pool.submit_once jitter_task
+    say "received samples"
+    consume_and_display empty_acc
```
maxteufel: then you can avoid using this pattern and use `realtime.Scheduler.env.submit` instead. This makes it easier to swap for a different implementation of Scheduler later on.

maxteufel (changes requested):
good but some nitpicks

modules/base/src/concur/task_pool.fz
Nimux2: run_periodic still uses p.as_duration in its debug pre/post checks, so an infinite period trips a divide-by-zero precondition. I’ve tweaked task_pool to handle infinite periods directly instead of routing them through run_periodic. It works for now, but I could try reworking Clock.fz instead, to properly handle such cases.
The error caused :
*** failed precondition: `debug: (numeric.this /! other)`

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,22 @@ jitter_threaded =>
...
+    test_jitter time.Nano time.Nano result_ch
+    result_ch.close
+
+  jitter_task := realtime.task time.frequency jitter_worker (time.frequency 0 (time.duration.ms 1))
+
+  realtime.Scheduler.instate concur.task_pool ()->
+    _ := realtime.Scheduler.env.submit jitter_task
+    say "received samples"
+    consume_and_display empty_acc
+    panic "Task-pool worker completed."
```
maxteufel: this should not `panic`. if we need to exit as a workaround, just use `exit`.

modules/base/src/concur/task_pool.fz
```diff
@@ -34,9 +34,12 @@ is
   public redef submit(P type : time.period, t realtime.task P) unit ! atomic_access =>
 
     nt := concur.Threads.env.spawn ()->
-      now := time.Nano.env.read
-      end_of_time := time.instant u64.max
-      _ := time.Nano.env.run_periodic now t.period end_of_time t.code
+      if t.period.is_infinity
+        _ := t.code.call
```
maxteufel: okay, but why not fix the actual problem in `run_periodic` instead?

modules/base/src/time/Clock.fz
```diff
@@ -105,18 +105,31 @@ public Clock ref : effect is
   pre
     debug: p.is_infinity || p.as_duration > time.duration.ns 0
   post
-    debug: ((last >= first): result <= ((last - first) / p.as_duration) + 1)
-    debug: ((last <  first): result = 0)
+    debug: p.is_infinity || ((last >= first): result <= ((last - first) / p.as_duration) + 1)
```
maxteufel:
```suggestion
    debug: ((last >= first): if p.is_infinity then result = 0 else result <= ((last - first) / p.as_duration) + 1)
```
can you check if this works instead? it would be the nicer alternative imo.

modules/base/src/time/Clock.fz
```diff
@@ -105,18 +105,31 @@ public Clock ref : effect is
...
+    debug: p.is_infinity || ((last < first): result = 0)
```
maxteufel: `p.is_infinity` is unnecessary here since `last < first` is well defined in that case as well, and the result should still be 0.

modules/base/src/time/Clock.fz
Nimux2: Shouldn't it be `if p.is_infinity then result = 1` instead ? The code would run once in this case.

maxteufel: yes, my bad

maxteufel (approved):

Nimux2:
@maxteufel could you merge this ?

fridis (changes requested):
This runs the original jitter test that is based on `Clock.run_periodic` in a task created by `realtime.task`. This can be done, but it still measures the jitter of `Clock.run_periodic`. 

What I consider more interesting is the jitter of a periodic task scheduled by `realtime.Scheduler.env.submit`. So, we need a variant of `test_jitter.run_for_period` that submits taks performing `payload` for different periods to `realtime.Scheduler.env`.

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,81 @@ jitter_threaded =>
...
+
+        # atomic for next expected release
+        next_rel := concur.atomic time.instant .new start
+
+        # task to be submitted to the scheduler
+        task_code =>
+          actual := time.Nano.env.read
+          if actual <= end
+            expected := next_rel.read
+            next_rel.write (expected + period)
```
maxteufel: this can lead to a race condition in case of an overrun

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,81 @@ jitter_threaded =>
...
+              (flow.exception jitter_too_high).env.cause <| error "release time jitter exceeded period"
+            payload
+            finish := time.Nano.env.read
+            if finish >= expected + period
+              err := error "*** overrun: run started at {actual-expected} and took {finish-actual}, which is larger than period $period"
+              (flow.exception jitter_too_high).env.cause err
+
+        # submit the periodic task to the scheduler
+        t := realtime.task time.duration task_code period
+        realtime.Scheduler.env.submit t
```
maxteufel: this will just keep running the task, even if it should have stopped.

this is an API problem, since we don't have an end time implemented yet even though we could. please do point out these limitations in the future, rather than working around them.

Nimux2:
@fridis do you think this is ready to be merged now ?

fridis (approved):

examples/jitter/jitter_threaded.fz
maxteufel: still not solved since the `state` effect is not thread-safe either; use blocking mutate.

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,92 @@ jitter_threaded =>
...
+        for
+          i := 0, i+1
+          s := i, s +° i
+        while i < 1000
+        else
+          if s = s +° 1
+            say "$s (never called)"
+
+      # Create and instate a blocking_mutate effect for the entire period loop.
+      bm := concur.blocking_mutate
```
maxteufel: no, you're supposed to create a type _inheriting_ from `concur.blocking_mutate` here. this is assigning it to a field.

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,92 @@ jitter_threaded =>
...
+
+      bm ! ()->
+        run_for_period(period time.duration)
```
maxteufel: what is the reason you swapped function definition and effect instation? i preferred the previous way you did this.

maxteufel (changes requested):

examples/jitter/jitter_threaded.fz
Nimux2: There wasn't any complex reasoning, I just had some difficulties making Jitter work with the new changes and this swap ended up working. I think I've managed to get it to work properly now.

maxteufel (approved):
i'm fine with this now!

maxteufel (changes requested):
i have just attempted merging and testing this; unfortunately in the new `jitter_threaded.fz`, the period jitter *always* seems to be around 500 ms, this seems like some wrong calculation or something. the existing `jitter.fz` seems is much lower and this is what i'd expect here as well.

Nimux2:
@fridis As Merle is on holiday, could you maybe review and merge this PR ?

fridis (changes requested):
I could not compile this when merging into my repository. Can you please merge main and make sure it still compiles.

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,97 @@ jitter_threaded =>
...
+
+        # Send end-of-period marker
+        marker := time.Nano.env.read
+        sample_ch <- (period, marker, marker, true)
+
+      # Instate the blocking_mutate effect around the period loop
+      bm ! ()->
+        for p := period_default, p.scale 0.5
+        while p.as_duration >= time.duration.ms 1 do
+          run_for_period p
```
fridis: The original `jitter.fz` does not terminate the loop early, but it reduces the period until there is an overrun or the jitter exceeds the period, here the code form `jitter.fz`: 
```
    match flow.exception jitter_too_high .on ()->

         for p := period_default, p.scale 0.5 do
           run_for_period p

      e error => say $e
      void    => # will not happen
```
Having an arbitrary limit of `1ms` means you do not measure the minimum period without excessive jitter, but you just stop at 1ms.

examples/jitter/jitter_threaded.fz
```diff
@@ -239,9 +216,97 @@ jitter_threaded =>
...
+
+  jitter_worker =>
+    test_scheduler_jitter result_ch
+    result_ch.close
+
+  jitter_task := realtime.task time.frequency jitter_worker (time.frequency 0 (time.duration.ms 1)) nil
+
+  # Instate blocking_mutate for the task_pool constructor (internal use)
+  concur.blocking_mutate ! ()->
+    realtime.Scheduler.instate (concur.task_pool concur.blocking_mutate) ()->
```
fridis: It is generally better to define your own mutate (as you do above)  and not just use the base lib's mutate. Otherwise, unrelated code that may use the same mutate type for a different purpose may get mixed up. 

So better declare a local mutate similar to `bm` above (but chosing a different name). You may also use only one mutate for both purposes and remove the code that instates `bm` above (provided that the effect worm hole works well, need to check).

fridis (changes requested):
The purpose of the jitter test is to run periodic tasks at smaller and smaller periods to find the period for which the jitter exceeds the length of the period. 

This version misses the detection of when the jitter has become too high, so it just always runs for all the periods. It arbitrarily stops at a period of 1ms and reports no final result. 

With the JVM backend on Linux, it is, of course, hard to reach periods below 1ms, but this test is supposed to be the basis to measure improvements when we run on a low-latency kernel, using the C backend, with realtime priorities, a realtime GC, CPU affinities and other measures that may reduce the jitter. There, periods in the order of tens of µsecs should be achievable.

maxteufel:
It keeps going for way longer now than the regular `jitter.fz`. This seems a bit too good to me. Are you sure the overrun detection is working properly?

```
testing period 59ns for 2100ms...
period jitter:
            ---  period jitter 59ns, log count / linear time  ---             
count
 64K_|_______|_______|_______|_______|_______|_______|_______|_______|_      ▉
 32K |       |       |       |       |       |       |       |       |       ▉
 16K |       |       |       |       |       |       |       |       |       ▉
  8K |       |       |       |       |       |       |       |       |       ▉
  4K_|_______|_______|_______|_______|_______|_______|_______|_______|_      ▉
  2K |       |       |       |       |       |       |       |       |       ▉
  1K |       |       |       |       |       |       |       |       |       ▉
512  |       |       |       |       |       |       |       |       |       ▉
256 _|_______|_______|_______|_______|_______|_______|_______|_______|_      ▉
128  |       |       |       |       |       |       |       |       |       ▉
 64  |       |       |       |       |       |       |       |       |       ▉
 32  |       |       |       |       |       |       |       |       |       ▉
 16 _|_______|_______|_______|_______|_______|_______|_______|_______|_      ▉
  8  |       |       |       |       |       |       |       |       |       ▉
  4  |       |       |       |       |       |       |       |       |       ▉
  2  |       |       |       |       |       |       |       |       |       ▉
  1 _|_______|_______|_______|_______|_______|_______|_______|_______|_      ▉
     |       |       |       |       |       |       |       |       |       |
    1ns     16ns    30ns    45ns    59ns    74ns    88ns   103ns   118ns     >

  864ms  | 2479µs  | 1945ms  |  602ms  |  74501  |

cycle jitter:
              ---  cycle jitter 59ns, log count / log time  ---               
count
 64K_|_______|_______|__▉____|_______|_______|_______|_______|_______|_      |
 32K |       |       |  ▉    |       |       |       |       |       |       |
 16K |       |       |  ▉    |       |       |       |       |       |       |
  8K |       |       |  ▉    |       |       |       |       |       |       |
  4K_|_______|_______|__▉____|_______|_______|_______|_______|_______|_      |
  2K |       |       |  ▉    |       |       |       |       |       |       |
  1K |       |       |  ▉    |       |       |       |       |       |       |
512  |       |       |  ▉▉   |       |       |       |       |       |       |
256 _|_______|_______|__▉▉___|_______|_______|_______|_______|_______|_      |
128  |       |       |  ▉▉   |       |       |       |       |       |       |
 64  |       |       |  ▉▉   |       |       |       |       |       |       |
 32  |       |       |  ▉▉▉  |       |       |       |       |       |       |
 16 _|_______|_______|__▉▉▉__|_______|_______|_______|_______|_______|_      |
  8  |       |       |  ▉▉▉▉ |       |       |       |       |       |       |
  4  |       |       |  ▉▉▉▉ |       |       |       |       |       |       |
  2  |       |       |  ▉▉▉▉ |      ▉|     ▉ |       |       |       |       |
  1 _|_______|_______|__▉▉▉▉▉|______▉|_▉__▉▉_|_______|_______|_______|_      |
     |       |       |       |       |       |       |       |       |       |
    1ns     90ns   4979ns  273µs    15ms   825ms    45s     41m     38h      >

  26µs   |  15µs   |  275ms  | 1389µs  |  74501  |

testing period 29ns for 2100ms...
period jitter:

error 1: Postcondition `debug: indx (result + time.duration.ns 1) = i+1` does not hold after call
Call stack:
fuzion.sys.fatal_fault#2 at {base.fum}/fuzion/sys.fz:55
[..]
fuzion.runtime.postcondition_fault#1 at {base.fum}/fuzion/runtime/post_fault.fz:56
time.histogram.post largest at {base.fum}/time/histogram.fz:248
time.histogram.largest#1 at {base.fum}/time/histogram.fz:246
time.histogram.precall largest at {base.fum}/time/histogram.fz:244
time.histogram.as_string.loop.(λ(i -> largest i .as_string.pad_center ...).call#1 at {base.fum}/time/histogram.fz:403
(list i64).map_to_list#2 String at {base.fum}/list.fz:242
((list i64).map_to_list#2 String).(λmap_to_list).call at {base.fum}/list.fz:242
((list.type String).of#4 String).anonymous.tail at {base.fum}/list.fz:764
((list String).map_to_list#2 String).(λmap_to_list).call at {base.fum}/list.fz:242
((list.type String).of#4 String).anonymous.tail at {base.fum}/list.fz:764
(list String).reduce#3 (container.expanding_array String) at {base.fum}/Sequence.fz:704
(list String).as_array at {base.fum}/list.fz:228
(list String).as_array_backed at {base.fum}/Sequence.fz:238
String.type.join#2 at {base.fum}/String.fz:949
(list String).as_string#1 at {base.fum}/list.fz:420
time.histogram.as_string.loop at {base.fum}/time/histogram.fz:403
time.histogram.as_string at {base.fum}/time/histogram.fz:385
time.histogram.prefix $ at {base.fum}/Any.fz:43
const_string.infix +#1 at {base.fum}/String.fz:66
jitter_threaded.print_period_histograms#3 at fuzion/examples/jitter/jitter_threaded.fz:165
jitter_threaded.fold_record_scheduler#5 at fuzion/examples/jitter/jitter_threaded.fz:245
jitter_threaded.consume_and_display_scheduler#1 at fuzion/examples/jitter/jitter_threaded.fz:266
jitter_threaded.(λ()-> realtime.Scheduler.instate (concu...).call.(λ()-> realtime.Scheduler.env.submit jit...).call at fuzion/examples/jitter/jitter_threaded.fz:360
(instate_helper unit realtime.Scheduler).call_code.call at {base.fum}/effect.fz:405
instate_helper unit realtime.Scheduler at {base.fum}/effect.fz:400
realtime.type.Scheduler.type.instate#4 unit at {base.fum}/effect.fz:170
realtime.type.Scheduler.type.instate#3 unit at {base.fum}/effect.fz:191
jitter_threaded.(λ()-> realtime.Scheduler.instate (concu...).call at fuzion/examples/jitter/jitter_threaded.fz:357
(instate_helper unit jitter_threaded.m).call_code.call at {base.fum}/effect.fz:405
instate_helper unit jitter_threaded.m at {base.fum}/effect.fz:400
jitter_threaded.type.m.type.instate#4 unit at {base.fum}/effect.fz:170
jitter_threaded.type.m.type.instate#3 unit at {base.fum}/effect.fz:191
jitter_threaded.m.instate_self#2 unit at {base.fum}/effect.fz:212
jitter_threaded.m.infix !#2 unit at {base.fum}/effect.fz:229
jitter_threaded at fuzion/examples/jitter/jitter_threaded.fz:356
```

maxteufel:
Additionally, it should not run into this postcondition.

maxteufel (approved):
This now fails to distinguish the cases of overrun (actual finish happened after next release) and jitter exceeded period (actual release happens after supposed next release). But okay. I don't find any other obvious bugs anymore.

maxteufel:
I believe you technically introduced a data race again in ad5f3b8, since two threads can at the same time read `error_cell` and then put in two different things at the same time. This is not grave here, but it is not nice.

maxteufel:
now you have implemented a deadlock in 76e3715: if two threads run into this at the same time, they one of their `compare_and_set` will succeed, and the other one will run into an infinite loop because never again will `error_cell` be `nil` (which is a requirement for `compare_and_set` to succeed).

maxteufel:
The deadlock would be fixed now, however note that `done` is not mutable. The second definition of the field just creates a new field invisible to the outside loop, so it has no effect.

maxteufel:
Good.

Nimux2:
@maxteufel Can this be merged ?

maxteufel (approved):
@Nimux2 Yes it can be merged.

PR #7438, opened 2026-07-22
relax isAssignableFrom if target_type is typepar with exactly matching t…

### LINKED ISSUE #7437: relax isAssignalbeFrom if expected is typepar with exactly matching this-type constraint
The following is currently not allowed, see error below. I think we should probably allow it.
```
e is

  type.i(_ e.this) => unit

  af(E type : e.this) =>
    E.i e.this
```
```

/home/sam/playground/test.fz:6:9: error 1: Incompatible types when passing argument in a call
    E.i e.this

Actual type for argument #1 '_' does not match expected type.
In call to          : 'e.type.i'
expected formal type: 'E'
actual type found   : 'e.this'
assignable to       : 'ref e'
for value assigned  : 'e.this'
To solve this, you could change the type of the target '_' to 'e.this' or convert the type of the assigned value to 'E'.

one error.
```

michaellilltokiwa:
…his-type constraint

fixes #7437

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -633,6 +633,20 @@ public abstract class AbstractType extends ANY implements Comparable<AbstractTyp
       {
         result = isAssignableFrom(actual.selfOrConstraint(context).asRef(true), context, allowBoxing, allowTagging, assignableTo);
       }
+    /**
+     * This allows code like this:
+     *
+     *     e is
+     *
+     *       type.i(_ e.this) => unit
+     *
+     *       af(E type : e.this) =>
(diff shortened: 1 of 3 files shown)


fridis (commented):
I think `E` and `e.this` cannot be treated as the same type in `af`: if I add a feature `f : e` as follows
```
e is
  type.i(_ e.this) => unit
  af(E type : e.this) =>
    E.i e.this

f : e is

e.af f
```
then in the call `e.af f`, we have `E` is `f` while `e.this` is `e` and both are incompatible value types.

michaellilltokiwa:
@fridis I see, hmm. Closing this then...

PR #7441, opened 2026-07-23
fe: `handDownListThroughInheritsCall` use target-type for applying type pars

michaellilltokiwa:
instead of this `instanceof` hack


fridis (approved):
looks good.

PR #7445, opened 2026-07-23
time/frequency: fix case of infinite period multiplied by zero

### LINKED ISSUE #7444: time/frequency: `infix *` does not handle the case of `f = 0` and `n = 0`
This is well-defined according to the definition of `infix *` in `time/period.fz`.

maxteufel:
Fix #7444

### DIFF
--- a/modules/base/src/time/frequency.fz
+++ b/modules/base/src/time/frequency.fz
@@ -73,10 +73,14 @@ is
   #
   public redef infix *(n u64) time.duration
   =>
-    ns := base.nanos.as_u128
-    nn := n         .as_u128
-    ff := f         .as_u128
-    time.duration (ns * nn / ff).as_u64
+    if n = 0
+      # by API definition
+      time.duration.zero
+    else
+      ns := base.nanos.as_u128
+      nn := n         .as_u128
+      ff := f         .as_u128
+      time.duration (ns * nn / ff).as_u64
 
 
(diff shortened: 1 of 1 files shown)


michaellilltokiwa:
Does it make sense to add this?
```diff
diff --git a/modules/base/src/time/period.fz b/modules/base/src/time/period.fz
index 5881a22b0..3a777bfa0 100644
--- a/modules/base/src/time/period.fz
+++ b/modules/base/src/time/period.fz
@@ -60,5 +60,7 @@ is
   public infix *(n u64) time.duration
   pre
     debug: !is_infinity || n=0
+  post
+    n!=0 || result.nanos = 0
   =>
     abstract
```

michaellilltokiwa (approved):

fridis:
The postcondition 
```
  post
    n!=0 || result.nanos = 0
```
should be 
```
  post
    debug: (n=0 : result.nanos=0)
```
or (after `time.duration` was made `equatable` or even `orderable`)
```
  post
    debug: (n=0 : result=time.duration.zero)
```
But all of this can be done in separate PRs.

maxteufel:
Note that
```
    n!=0 || result.nanos = 0
```
and
```
    n=0 : result.nanos=0
```
are equivalent (since A implies B is equivalent to not A or B).

I prefer the latter too though and have opened #7446.

PR #7452, opened 2026-07-27
base: remove unnessary type param `MA` from `monad`

michaellilltokiwa:
I think `MA` was only needed before we had the `fixed` modifier?

Also implemented `>>=` by using `>>=~` which leaves us with three abstract features in monad, one type constructor and two operations.


maxteufel (approved):

fridis (approved):
Yes, this was probably written before we had support for `this`-types.

PR #7454, opened 2026-07-27
fe: improve type inference for nested InlineArrays

michaellilltokiwa:
the inner array may be a `Sequence` and not an `array`

### DIFF
--- a/modules/lock_free/src/lock_free/Map.fz
+++ b/modules/lock_free/src/lock_free/Map.fz
@@ -118,7 +118,7 @@ is
   public redef as_string String => "container_node[gen=$gen]($(array.as_string ", "))"
 
   as_list(f Indirection_Node CTK CTV -> Main_Node CTK CTV) =>
-    array.flat_map /* NYI: UNDER DEVELOPMENT: type inference */ (tuple CTK CTV) (.as_list f)
+    array.flat_map /* NYI: BUG: type inference */ (tuple CTK CTV) (.as_list f)
          .as_list
 
 
--- a/src/dev/flang/ast/InlineArray.java
+++ b/src/dev/flang/ast/InlineArray.java
@@ -176,7 +176,15 @@ public class InlineArray extends ExprWithPos
       {
         _type = super.type();
       }
-    return _type;
+
+    return _type == null || _type.isArtificialType() || !_type.containsArtificialType()
+      ? _type
+      : !urgent
+      ? null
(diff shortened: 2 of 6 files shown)


modules/lock_free/src/lock_free/Map.fz
```diff
@@ -118,7 +118,7 @@ is
   public redef as_string String => "container_node[gen=$gen]($(array.as_string ", "))"
 
   as_list(f Indirection_Node CTK CTV -> Main_Node CTK CTV) =>
-    array.flat_map /* NYI: UNDER DEVELOPMENT: type inference */ (tuple CTK CTV) (.as_list f)
+    array.flat_map /* NYI: BUG: type inference */ (tuple CTK CTV) (.as_list f)
```
maxteufel, line 121: this will lead to merge conflict with #7457.

src/dev/flang/ast/InlineArray.java
```diff
@@ -176,7 +176,15 @@ else if (_type == null && urgent)
       {
         _type = super.type();
       }
-    return _type;
+
+    return _type == null || _type.isArtificialType() || !_type.containsArtificialType()
```
maxteufel, line 180: `_type.isArtificialType` should be unnecessary here since this is already checked in `containsArtificialType`.

src/dev/flang/ast/InlineArray.java
```diff
@@ -200,9 +208,12 @@ else if (_type == null && urgent)
   @Override
   Expr propagateExpectedType(Resolution res, Context context, AbstractType t, Supplier<String> from)
   {
+    var bt = t.isNormalType() && Types.resolved.f_array.inheritsFrom(t.feature()) && t.feature().typeArguments().size()==1
```
maxteufel: what does `bt` mean in this context?

maxteufel (changes requested):

src/dev/flang/ast/InlineArray.java
michaellilltokiwa: hmm, but it is: isArtificialType and NOT containsArtificialType :shrug:

src/dev/flang/ast/InlineArray.java
michaellilltokiwa: I changed the name since it was confusing.

maxteufel (approved):

PR #7459, opened 2026-07-28
realtime/task: allow setting last time a task runs

maxteufel:
I'd love feedback on the API design here. This is good for now, but with tasks having more and more parameters I feel like this could become very problematic in the near future.

### DIFF
--- a/modules/base/src/realtime/task.fz
+++ b/modules/base/src/realtime/task.fz
@@ -26,4 +26,5 @@
 #
 public task(P type : time.period,
             module code Function unit,
-            module period P) is
+            module period P,
+            module last_start option time.instant) is
--- a/modules/base/src/concur/task_pool.fz
+++ b/modules/base/src/concur/task_pool.fz
@@ -34,9 +34,9 @@ is
   public redef submit(P type : time.period, t realtime.task P) unit ! atomic_access =>
 
     nt := concur.Threads.env.spawn ()->
+      actual_last := t.last_start.or_default (time.instant u64.max)
       now := time.Nano.env.read
-      end_of_time := time.instant u64.max
-      _ := time.Nano.env.run_periodic now t.period end_of_time t.code
+      _ := time.Nano.env.run_periodic now t.period actual_last t.code
 
     thrds.write (list nt thrds)
 


maxteufel:
@Nimux2 this is what you need in #7418. I also want to hear feedback on the API design from you.

modules/base/src/realtime/task.fz
```diff
@@ -26,4 +26,5 @@
 #
 public task(P type : time.period,
             module code Function unit,
-            module period P) is
+            module period P,
+            module last option time.instant) is
```
michaellilltokiwa: This name would leave me guessing what this means. Add a comment or choose a better name, `latest_start`/`last_start`?

michaellilltokiwa:
As for the API design, I don't mind many args on the constructor of task. We should have short-hands for commonly used ones though.

Nimux2:
> @Nimux2 this is what you need in #7418. I also want to hear feedback on the API design from you.

Do you want feedback about what you just did, or about the priority/affinity aspect from last week ?

maxteufel:
i want feedback from you about this particular change here, and how in the context of future changes (we will probably also add arguments `priority` and `scheduling_policy` here) this will be as a developer. it might, for example, be extremely annoying to write
```
t := realtime.task time.frequency (()->say "actual code here") (time.frequency 10 (time.duration.s 1)) nil 99 concur.scheduling_policy.sched_fifo
```
every time you want to have a task. i want to know if you think that my suggestion of just adding arguments here is good enough, or if we want to use some other way of implementing this in the future.

Nimux2:
> i want feedback from you about this particular change here, and how in the context of future changes (we will probably also add arguments `priority` and `scheduling_policy` here) this will be as a developer. it might, for example, be extremely annoying to write
> 
> ```
> t := realtime.task time.frequency (()->say "actual code here") (time.frequency 10 (time.duration.s 1)) nil 99 concur.scheduling_policy.sched_fifo
> ```
> 
> every time you want to have a task. i want to know if you think that my suggestion of just adding arguments here is good enough, or if we want to use some other way of implementing this in the future.

I've actually stumbled upon this issue while working on the API. I don't think it's too problematic. If there are too many arguments, it could be better to merge some of them into a shared structure to reduce the bloat. Something like : 
```
public task_options
(
  priority realtime.priority,
  affinity realtime.affinity,
  policy realtime.execution_policy
)
```

Otherwise they could also be spread out between `task` and `submit`, depending on what is appropriate. Keeping some of those in `submit` should allow for running the same task with varying priorities, or pinned to different CPUs. Not sure if that would be a common occurrence, but having the option can always be useful.

maxteufel:
but this would just shift the boilerplate to creating the `task_options` instance

Nimux2:
> but this would just shift the boilerplate to creating the `task_options` instance

This would still allow for easily reusable parameters, and adding new fields afterwards doesn't require editing every call site. I personally think it's generally cleaner and more readable (assuming parameters are grouped by theme and such).

michaellilltokiwa (approved):

PR #7462, opened 2026-07-28
base: rename `list_concat_monoid` as `concat_monoid`

michaellilltokiwa:


modules/base/src/Sequence.fz
```diff
@@ -1318,7 +1318,7 @@ is
 
   # monoid of Sequences with infix concatenation operation.
   #
-  public type.concat_monoid Monoid (Sequence T) =>
+  public fixed type.concat_monoid Monoid (Sequence T) /* NYI: UNDER DEVELOPMENT: better: Sequence.this */ =>
```
fridis, line 1321: Not sure I want to use `Sequence.this` here since that would require `concat_monoid` implementations for alll heirs of `Sequence`.

modules/base/src/list.fz
```diff
@@ -732,11 +732,7 @@ is
 
   # monoid of lists with infix concatenation operation.
   #
-  # NYI: Name should be `concat_monoid`, we use `list_concat_monoid`
-  # to avoid clash with inherited `Sequence.concat_monoid`. Maybe
-  # the inherited one should be renamed once renaming is supported.
-  #
-  public type.list_concat_monoid Monoid (list A) =>
+  public fixed type.concat_monoid Monoid (list A) =>
```
fridis, line 735: I think there might be a bug here that skips the type check for a `fixed` feature's result if the redefined feature is also `fixed` and now non-fixed parent exists. 

In any case, `Monoid (list A)` should probably not be comatible to `monoid (Sequence T)`.

fridis (changes requested):
I think the use of `fixed` here falsely disables some type checking that would be needed.

modules/base/src/Sequence.fz
michaellilltokiwa: @fridis I think, a monad needs to know how to construct the type so we need specific implementations for heirs anyway. Sequence.type.concat_monoid can still be used for any heir of Sequence if programmer chooses to.

fridis: ok, but then it should be ok for `my_sequence_type.concat_monoid` to cause an error that `abstract concat_monoid` is not implemented. 

Alternatively, we could make this a normal feature: 
```
public container.sequence_concat_monoid Monoid (Sequence T) => ...
```
but maybe the type feature is nicer.

michaellilltokiwa: @fridis  I don't see where the type parameter `T` comes from in the normal feature. I think a type feature is definitely nicer.

michaellilltokiwa: @fridis In the case `abstract not implemented` we could also show a hint that the abstract is actually a fixed type feature that just needs to be called differently.

PR #7467, opened 2026-07-28
fe: make type parameters with unspecified visibility inherit type visibility of outer

michaellilltokiwa:
fixes bug in test choice constraints

@tokiwa-software/developers Do you agree with this change in semantics for default visibility of type parameters?

### DIFF
--- a/src/dev/flang/ast/Feature.java
+++ b/src/dev/flang/ast/Feature.java
@@ -108,7 +108,9 @@ public class Feature extends AbstractFeature
   public Visi visibility()
   {
     return _visibility == Visi.UNSPECIFIED
-      ? Visi.PRIV
+      ? (isTypeParameter()
+          ? outer().visibility().typeVisibility()
+          : Visi.PRIV)
       : _visibility;
   }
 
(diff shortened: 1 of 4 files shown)


maxteufel (commented):
I'm fine with this but @fridis needs a say in this.

michaellilltokiwa:
@fridis I'm thinking maybe it makes sense to make this even more simple. Disallow explicit visibility on typepars altogether and make them public by default.

fridis (commented):
I am myself not decides, just trying to collect thoughts and pros and cons here: 

I think we need more tests and documentation for this. As I understand this PR, it will allow specifying type visibility for type parameters as in 
```
public f(public:module T type) is ...
public g(module:public U type) is ...
``` 
such that in another module, you could do 
```
  f.test => 
    say T               # ok, T called
    say (option T nil)  # not ok, T used as type
  g.test => 
    say U               # not ok, U called
    say (option U nil)  # ok, U used as type
```
I think this is maybe overkill and we should not distinguish type and call visibility for type parameters. 

Also, the automatic inheritance of type parameter visibility from the outer feature's typeVisibility is IMHO too complex. I think it would be fine to not inherit this and instead to use explicit `public` visibility in our base lib whenever the outer feature is public or defines a public type.  This, of course, has the disadvantage that it is easy to forget making the type parameter `public`, but this is a problem only in the rare case of adding inner features in other modules.  And four our base lib, we can take care.

PR #7479, opened 2026-07-30
fuir: AbstractInterpreter, remove `nop`, mark `drop` as abstract

michaellilltokiwa:
`nop` feels useless to me and `drop` should be abstract anyway

### DIFF
--- a/src/dev/flang/be/c/C.java
+++ b/src/dev/flang/be/c/C.java
@@ -122,10 +122,18 @@ public class C extends ANY
 
 
     /**
-     * no operation, like comment, but without giving any comment.
+     * drop a value, but process its side-effect.
+     *
+     * @param v an expression that calculates a value that is not needed, but
+     * where the calculation might have side-effects (like performing a call) that
+     * we do need.
+     *
+     * @param type clazz id for the type of the value
+     *
+     * @return code to perform the side effects of v and ignoring the produced value.
      */
     @Override
-    public CStmnt nop()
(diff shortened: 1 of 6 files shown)


fridis (commented):
For me, a `nop` is a very fundamental operation in any IR. It is a little surprising that is occurs at so few places here, but I would still like to keep it, even if it is really needed only for the assertion that a constant creates no code. 

Maybe we can use `nop()` more often instead of `CStmt.EMPTY`/`Expr.UNIT`. Maybe it would even make sense to rename `CStmt.EMPTY` as `CStmt.NOP` and `Expr.UNIT` as  `Expr.NOP` for consistency.

michaellilltokiwa:
@fridis At first it felt similar to me, that `nop` is indeed fundamental but thinking about it this is about emitting a `nop` instruction not about `nop` as a concept. So what is the point? I don't really see it currently?

michaellilltokiwa:
Not sure if this is true! AI claims the following:

> Most modern semantic IRs (including LLVM IR and MLIR) do not have a built-in, general-purpose nop instruction. Instead, optimization and rewriting passes typically erase operations that no longer contribute to the program's semantics.

fridis:
This `nop()` is not an IR instruction, but a function that produces an empty code sequence `CStmt.EMPTY` or `Expr.EMPTY`. I think it is useful to have this concept at the level of the `AbstractInterpreter`, even if it is currently used only for one `check`.

PR #7481, opened 2026-07-30
lock_free: make `Sieve_Cache` map implementation agnostic

michaellilltokiwa:


maxteufel (approved):

maxteufel:
reveals a bug; `Node` is still unambiguous in theory but this is wrongly claimed as a problem

PR #7483, opened 2026-07-30
fe: add (workaround) assignabilty rule `a.type.T` assignable to `a.T`

michaellilltokiwa:
I'm not 100% sure but the rule is probably unnecessary if replace type pars in type features by there cotype counter parts. 
I added an NYI  CLEANUP comment for this.

This allows implementing `Mutable_Map.type.from_entries`

### DIFF
--- a/modules/base/src/container/Mutable_Map.fz
+++ b/modules/base/src/container/Mutable_Map.fz
@@ -123,11 +123,6 @@ public Mutable_Map(public K type : property.equatable, public V type) ref is
   public type.empty Mutable_Map.this => abstract
 
 
-/*
-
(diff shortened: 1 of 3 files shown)


fridis (approved):
Looks good. 

I am unhappy about the special handling required for type parameters and their counterparts in cotypes, but I do not have a simple solution. Best would be to have only one central location when comparing types that checks that we have a type parameter vs. its cotype counterpart and returns that they are essentially the same. 

Alternatively, a more radical change would be not to have duplicate features for type parameters and their cotype siblings, but to use the same features for both. But this might lead to confusion and special handling at other locations.

PR #7484, opened 2026-07-30
reference manual: add visibility in glossary

michaellilltokiwa:
[ci skip]


maxteufel (approved):
Good, but maybe the identical section from the website should link here then.

PR #7493, opened 2026-07-31
http: compute `header_as_canonical_string` once only

michaellilltokiwa:


maxteufel (approved):

modules/http/src/http/Message.fz
```diff
@@ -124,9 +124,7 @@ public Message ref is
   # Host: www.example.com
   # User-Agent: Mozilla/5.0
   #
-  # NYI: PERFORMANCE: compute this only once
-  #
-  header_as_canonical_string =>
+  header_as_canonical_string := once mutate _ ()->
```
fridis: Using `mutate` effect here means that you lose the information what gets mutated, would be nice to have a specific mutate that shows that this is for the http message's canonical string. 

Even better would be if `once` would internally use such a local mutate and hide this from the user.

fridis (changes requested):
I guess this is to improve performance. Would be good to write this in the PR message, does this fix an important performance issue or is this just a minor improvement?

Try not to use plain `mutate`.

modules/http/src/http/Message.fz
michaellilltokiwa: changed to address this, `once` does not require a mutate anymore.

fridis (approved):
good.

PR #7494, opened 2026-07-31
base: date_time, fix milli_second bug

michaellilltokiwa:
date_time just divided nano_second by one million and returned this as the result but this was then not subtracted from nano_second.


maxteufel (commented):
nice catch! does a regression test make sense here?

maxteufel (approved):

PR #7495, opened 2026-07-31
parser: use universe as target for call to tuple

michaellilltokiwa:
Otherwise this might lead to problems if another feature named `tuple` is declared somewhere.


maxteufel (approved):

maxteufel:
I suggest adding a regression test here. The following example currently (i.e. without this change) throws an ambiguous targets error:

```
ex_uni_tuple unit =>

  tuple(A type...,
        values A...) : property.orderable, property.hashable is

      "blabla new tuple"


  x := (1, 2, 3, 4)
  say x
```

PR #7496, opened 2026-07-31
ast: Loop add explicit target `Universe` for calls to `i64`

michaellilltokiwa:


maxteufel (approved):
nice! probably needs a similar regression test like in https://github.com/tokiwa-software/fuzion/pull/7495#issuecomment-5144312561

PR #7497, opened 2026-07-31
base: remove parametrized `indices`

### LINKED ISSUE #7470: `abstract_array.indices(I type: property.interval)` is considered dangerous
This refers to PR #7460: I have some issues with this function: 

- it will fail for large arrays (>=2³¹ elements) without properly expressing this in a precondition
- `numeric.from_u32` is meant for small constants like 0, 1, 2, it requires time linear in the value of the integer
- it can have very surprising results, e.g., when called with `modulo_5` for a cyclic numeric type `modulo_5`. 

I suggest to remove this variant of `indices`.

What could be added instead is something like this 
```
  is_valid_index(I type : numeric, ix I)
  => 
    ix.fits_in_i64 && indices.contains ix.as_i64
```

michaellilltokiwa:
fixes #7470

### DIFF
--- a/modules/base/src/container/abstract_array.fz
+++ b/modules/base/src/container/abstract_array.fz
@@ -135,19 +135,10 @@ public abstract_array
       abstract_array.this
 
 
-  # a sequence of all valid indices to access this array. Useful e.g., for
-  # `for`-loops:
-  #
-  #     for i in arr.indices do
-  #       say arr[i]
-  #
-  public indices(I type: property.interval) interval I => I.zero..(I.from_u32 length.as_u32)-1
-
-
   # indices as i64 interval
   #
-  public indices interval i64 => indices i64
-
+  public indices interval i64 =>
+    (i64 0) .. length-1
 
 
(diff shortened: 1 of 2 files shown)


maxteufel (approved):

modules/base/src/container/abstract_array.fz
```diff
@@ -135,19 +135,10 @@ public abstract_array
...
-  # a sequence of all valid indices to access this array. Useful e.g., for
-  # `for`-loops:
-  #
-  #     for i in arr.indices do
-  #       say arr[i]
-  #
-  public indices(I type: property.interval) interval I => I.zero..(I.from_u32 length.as_u32)-1
-
-
   # indices as i64 interval
```
simonvonhackewitz: I'd keep the original comment
```suggestion
  # a sequence of all valid indices to access this array. Useful e.g., for
  # `for`-loops:
  #
  #     for i in arr.indices do
  #       say arr[i]
  #
```

Issue #6066, opened 2025-10-08
add native features to access local hostname(s)


maxteufel:
implement `gethostname` and friends

michaellilltokiwa:
what are the _friends_? _uname_?

PR #7514, opened 2026-08-06
base: add `os.hostname`

simonvonhackewitz:
first part of #6066

### DIFF
--- a/include/fz.h
+++ b/include/fz.h
@@ -484,6 +484,12 @@ int fzE_process_create(char * args[], size_t argsLen, char * env[], size_t envLe
 int64_t fzE_process_poll(int64_t p);
 
 /**
+ * get the hostname
+ * @return -1 error, length of hostname on success
+ */
+int fzE_hostname(char *buf, size_t nbytes);
+
+/**
  * open a new pipe, put read and write end into array passed here
  *
  * @param fds array to be filled with pipe's file descriptors
--- a/include/fz.h
+++ b/include/fz.h
@@ -485,7 +485,7 @@ int64_t fzE_process_poll(int64_t p);
 
 /**
  * get the hostname
- * @return -1 error, length of hostname on success
+ * @return -1 on error, length of hostname on success
  */
 int fzE_hostname(char *buf, size_t nbytes);
 
--- a/include/posix.c
+++ b/include/posix.c
@@ -835,10 +835,7 @@ int64_t fzE_process_poll(int64_t p){
 //
 int fzE_hostname(char *buf, size_t nbytes)
 {
-    if (buf == NULL || nbytes == 0) {
-        errno = EINVAL;
-        return -1;
-    }
+    assert (buf != NULL && nbytes != 0);
 
     if (gethostname(buf, nbytes) != 0) {
         return -1;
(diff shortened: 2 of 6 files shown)


include/posix.c
```diff
@@ -831,6 +831,26 @@ int64_t fzE_process_poll(int64_t p){
 }
 
 
+// returns -1 on error or length of the hostname otherwise
+//
+int fzE_hostname(char *buf, size_t nbytes)
+{
+    if (buf == NULL || nbytes == 0) {
+        errno = EINVAL;
```
maxteufel: i'm not sure if i understand the workings of `errno` well enough to say if this is fine or not; it seems the most critical part of this change.

maxteufel (approved):

include/posix.c
michaellilltokiwa: The pattern we currently use is to have an assert if runtime functions are used in the wrong way.

include/posix.c
```diff
@@ -831,6 +831,26 @@ int64_t fzE_process_poll(int64_t p){
...
+        return -1;
+    }
```
michaellilltokiwa:
```suggestion
    assert (buf != NULL && nbytes != 0);
```

michaellilltokiwa (changes requested):
Use assert instead of returning normally, see change suggestion.

include/fz.h
```diff
@@ -483,6 +483,12 @@ int fzE_process_create(char * args[], size_t argsLen, char * env[], size_t envLe
  */
 int64_t fzE_process_poll(int64_t p);
 
+/**
+ * get the hostname
+ * @return -1 error, length of hostname on success
```
michaellilltokiwa:
```suggestion
 * @return -1 on error, length of hostname on success
```

modules/base/src/os/hostname.fz
```diff
@@ -0,0 +1,33 @@
...
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion os.hostname
+#
+# -----------------------------------------------------------------------
+
+# get the hostname
+#
+public hostname outcome String =>
```
michaellilltokiwa: This should be a default effect.

michaellilltokiwa:
```
Hostname ref : effect is
  get outcome String =>
```

michaellilltokiwa (changes requested):

simonvonhackewitz:
test failure due to #7177

modules/base/src/os/Hostname.fz
```diff
@@ -0,0 +1,38 @@
...
+# -----------------------------------------------------------------------
+
+
+# effect to get the hostname
+#
+public Hostname ref : effect is
+
+  # get the hostname
+  #
+  public get outcome String =>
```
michaellilltokiwa: I think `get` should be `abstract`. 
And you need to implement it in `default_value`, see other default effects.

michaellilltokiwa (approved):
Very good now!

include/posix.c
maxteufel: why?

michaellilltokiwa: @maxteufel This is basically like a precondition. Shows wrong usage of this runtime function.

PR #7520, opened 2026-08-11
base: change Reader to `ref`-style from `Handler`-style

michaellilltokiwa:


bin/add_simple_test.fz
```diff
@@ -30,7 +30,7 @@ add_simple_test =>
   lm : mutate is
   lm ! ()->
 
-    io.stdin.reader lm ! ()->
+    _ := (io.buffered lm).Reader.instate (io.stdin.reader lm) ()->
       match (io.buffered lm).read_line
```
fridis, line 34: I am surprised that the original code worked, how did the called `(io.buffered lm).read_line` know that is had to use `io.stdin.reader.env` and not `(io.buffered lm).reader.env`?

fridis (commented):
Good. 

While reviewing, I failed to understand why the original code worked, so I add a corresponding question. 

Not for this PR, but for future improvements, it would be good to 

- reduce the boilerplate that we have  when using this
- have some mechanism that would permit several readers, say, I would like to write  a feature like `join(R1, R2 type : Reader, W type : Writer) that would read from `R1.env` and `R2.env` and write what it got to `W.env`. We could do this with different mutate types passed to `io.buffered`
- which leads me to the last point: the `ìo.buffered lm` is maybe a bit clumsy, but I cannot suggest an alternative right now. 

Maybe  a good solution would address all three points at the same time...

bin/add_simple_test.fz
michaellilltokiwa: `io.stdin.reader lm` returns `(buffered lm).reader`  and this is what gets instated

michaellilltokiwa:
I have been thinking about better API design for reader/writer as well. Obviously a buffered reader is stateful.
If we want to hold the state via means of the effect-system we need different types for different readers, this can be achieved by type pars as currently done with `buffered LM` or via inheritance. Or we do not model the state of the buffered reader/writer via the effect system at all, at least not directly.

bin/add_simple_test.fz
fridis: ok, I did not see that `io.stdin.reader` was a function.

fridis:
> I have been thinking about better API design for reader/writer as well. Obviously a buffered reader is stateful. If we want to hold the state via means of the effect-system we need different types for different readers, this can be achieved by type pars as currently done with `buffered LM` or via inheritance. Or we do not model the state of the buffered reader/writer via the effect system at all, at least not directly.

I think having a mutate type as a argument to `buffered` is good. What I would play around with is maybe moving `Reader` to `io.Reader`, maybe having a `io_buffer_mutate` with a default value as an alternative to a user define local mutate, using `R type : Reader` and `R.env` instead of `(io.bufferd lm).Reader.env` and instating the value type effects instead and similar things. I am not sure what will work out nicely, but I am full of hope that something will do.

fridis (approved):
good!

PR #7521, opened 2026-08-11
ast: raise error when unit result is discarded explictly

### LINKED ISSUE #7463: raise error when unit result is discarded explictly
```
  _ := some_code_resulting_in_unit
```

Rationales:
- if this is discarded explicitly programmer probably had some different behavior in mind.
- Also leads to more readable/cleaner code

simonvonhackewitz:
fix #7463

example:
```
❯ fz -e '_ := say "hello"' | cat

command line:1:1: error 1: A result of type 'unit' must not be ignored explicitly.
_ := say "hello"
^
Remove '_ :=' to fix this error.

one error.
```


src/dev/flang/ast/AstErrors.java
```diff
@@ -2675,6 +2675,13 @@ public static void anonymousFeatureMustNotInheritFromMultiple(SourcePosition ran
       i.stream().map(ic -> ic.pos().showInSource()).collect(Collectors.joining("\nand\n\n")));
   }
 
+  public static void unitResultExplicitlyIgnored(SourcePosition range)
+  {
+    error(range, "A " + st("unit") + " result must not be ignored explicitly.",
+      "If the result is expected to be " + st("unit") + " and you intend to ignore it, remove " + ss("_ :=") + "."
```
michaellilltokiwa: I would write something like: A result of type `unit` must not be ignored explicitly. Remove " + ss("_ :=")  to fix this error.

src/dev/flang/ast/AbstractAssign.java
```diff
@@ -247,6 +247,11 @@ void checkTypes(Resolution res, Context context)
             _value.checkAmbiguousAssignmentToChoice(frmlT);
           }
 
+        if (f.featureName().baseNameHuman().equals("_")
```
michaellilltokiwa:
```suggestion
        if (f.featureName().baseName().startsWith(FuzionConstants.UNDERSCORE_PREFIX)
```

michaellilltokiwa: Or even better not use string comparison at all...

michaellilltokiwa (commented):
Cool. I have two suggestions.

src/dev/flang/ast/AbstractAssign.java
simonvonhackewitz: But `startsWith` changes the behavior, 
e.g. `_x := 42` causes an unused field error, while
```
_x := 42
say _x
```
works without errors.


Is it possible to avoid string comparison?

michaellilltokiwa: `startsWith` code should work the same since underscore features start with a hashtag: `UNDERSCORE_PREFIX = INTERNAL_NAME_SYMBOL + "_";`

michaellilltokiwa: Yes it is possible to avoid string comparison, just add a method to Feature `isNamelessField` or similar that returns true if feature is a nameless field.

michaellilltokiwa (approved):
Nice

PR #7522, opened 2026-08-12
tokiwa: error early when closing of in/out fails

michaellilltokiwa:
I think this always indicates some misuse.

### DIFF
--- a/modules/tokiwa/src/tokiwa.fz
+++ b/modules/tokiwa/src/tokiwa.fz
@@ -148,12 +148,12 @@ public tokiwa is
                 (p.with_in unit lm ()->
                   (io.buffered lm).Writer.env.write (io.slurp stdin_file).or_panic .or_panic).or_panic
 
-        p.close_in.or_else (say "closing in of $p failed")
+        p.close_in.or_panic
 
         match p.wait (timeout.or_default time.duration.max)
           e error =>
-            p.close_out.or_else (say "closing out of $p failed")
-            p.close_err.or_else (say "closing err of $p failed")
+            p.close_out.or_panic
+            p.close_err.or_panic
             p.kill     .or_else (say "sending kill signal to $p failed")
             process_result "" "failed waiting for $cmd, error is: $e" 1
           ec u32 => process_result fout.get ferr.get ec)


modules/tokiwa/src/tokiwa.fz
```diff
@@ -148,12 +148,12 @@ public tokiwa is
                 (p.with_in unit lm ()->
                   (io.buffered lm).Writer.env.write (io.slurp stdin_file).or_panic .or_panic).or_panic
 
-        p.close_in.or_else (say "closing in of $p failed")
+        p.close_in.or_panic
```
simonvonhackewitz, line 151: I think keeping the error message might be helpful
```suggestion
        p.close_in.or_panic "closing in of $p failed"
```

modules/tokiwa/src/tokiwa.fz
```diff
@@ -148,12 +148,12 @@ public tokiwa is
...
 
         match p.wait (timeout.or_default time.duration.max)
           e error =>
-            p.close_out.or_else (say "closing out of $p failed")
-            p.close_err.or_else (say "closing err of $p failed")
+            p.close_out.or_panic
+            p.close_err.or_panic
```
simonvonhackewitz, line 156: This is already in the error case. Does it make sense to panic here?
If so, adding the error message might make sense here too.

modules/tokiwa/src/tokiwa.fz
```diff
@@ -148,12 +148,12 @@ public tokiwa is
...
             p.kill     .or_else (say "sending kill signal to $p failed")
```
simonvonhackewitz, line 157: Did you leave this out on purpose?

modules/tokiwa/src/tokiwa.fz
michaellilltokiwa: I think the _default_ message is better:
`error "closing $(desc_name std_in) of process $(pid)" fzE_last_error`

modules/tokiwa/src/tokiwa.fz
michaellilltokiwa: Yes, since I think sending the kill signal might reasonably fail, but might still be good to have this in the logs.

modules/tokiwa/src/tokiwa.fz
michaellilltokiwa: Hmm, good point

michaellilltokiwa: I think, even in the error case we should expect closing of out and err to succeed.

simonvonhackewitz (approved):

PR #7525, opened 2026-08-12
rt: add assertion `fzE_file_read` for case `EAGAIN`

michaellilltokiwa:
on `EAGAIN` we want the result to be greater than zero since means `end_of_file`.


include/posix.c
```diff
@@ -1058,6 +1058,7 @@ int32_t fzE_file_read(void * file, void * buf, int32_t size)
       while (fread_result == 0 && !feof((FILE*)file) && (ferror((FILE*)file) && errno == EAGAIN));  // if we got no data and no EOF, then repeat.
       if (!ferror((FILE*)file) || errno == EAGAIN)
         {
+          assert ( errno != EAGAIN || fread_result > 0 );
```
fridis: The assert failing means `errno == EAGAIN && fread_result == 0`. Together with the failed while-condition, this implies `feof((FILE*)file)`. So this is basically checking that we do not get `feof` together with`EAGAIN`. 

Are we sure that this cannot happen? And if it did, is it maybe better to continue anyways by returning `0` (which in your updated comment in fz.h is the result in case of end-of-file.

fridis: Maybe it would be better to do 

```suggestion
while (fread_result == 0 && !feof((FILE*)file) && (ferror((FILE*)file) && errno == EAGAIN));  // if we got no data and no EOF, then repeat.
if (!ferror((FILE*)file) || errno == EAGAIN)
{
          assert ( fread_result > 0 || feof((FILE*)file));
```
but this could as well be just a comment since it is implied by the failed while-condition.

fridis (commented):
Nice to have the eof-behavior documented well. 

I was struggling a bit with the conditions here. Maybe would be better to check in the assert just what is documented for the result: if zero is returned, we must be at eof.

fridis (approved):
good.

PR #7532, opened 2026-08-13
rt: add static_assertion `EWOULDBLOCK == EAGAIN`

michaellilltokiwa:
fzE_file_read just checks `EAGAIN` which is only sufficient if `EWOULDBLOCK == EAGAIN`


fridis (commented):
If I understand this right, `fread` on a non-blocking file would return `EWOULDBLOCK`/`EAGAIN`. But then, our code in `fzE_file_read` would spin actively until there is data available. 

This does not sound good, I think we need different handling for non-blocking reads, if somewhere we do not want blocking reads, I cannot imagine that we are fine with busy waiting instead...

maxteufel (approved):
We do not currently support non-blocking I/O so this is just a correctness fix for the current implementation. Fine with me.

PR #7535, opened 2026-08-13
rt: remove comment on flushing pipe

michaellilltokiwa:
I believe flushing a pipe is unnecessary


maxteufel (changes requested):
"I believe [it] is unnecessary" is never a good reason in C code. Do you have anything to back this up?

michaellilltokiwa:
@maxteufel If we close one end of the pipe, written data can still be read. This is the way pipes work for my understanding. So explicit flushing here is actually not what we want here.

maxteufel (approved):
I think that is the wrong intuition of how it works. But here is a good explanation: <https://stackoverflow.com/a/43188944>.

PR #7538, opened 2026-08-14
base: deduplicate pipewriter/reader code

michaellilltokiwa:


modules/base/src/os/pipe.fz
```diff
@@ -132,3 +105,36 @@ private:public pipe : effect is
         _ := fzE_pipe_close fd
       else
         unit # ignore
+
+
+module pipe_reader (LM type: mutate, desc i64, error_msg String) : (io.buffered LM).Reader is
```
fridis: please add some documention. Where is this to be used (only in `pipe`, also in `process`, or could there be other uses as well?, what is done to `error_msg`, what is `desc`?

modules/base/src/os/pipe.fz
```diff
@@ -132,3 +105,36 @@ private:public pipe : effect is
...
+    res := fzE_pipe_read desc arr.internal_array.data arr.count
+    if res = -1
+      error "$error_msg." fzE_last_error
+    else if res = 0
+      io.end_of_file
+    else
+      arr.slice 0 res
+
+
+module pipe_writer (LM type: mutate, desc i64, error_msg String) : (io.buffered LM).Writer is
```
fridis: dito.

fridis (changes requested):
Nice to remove the duplicated code. 

Just some requests to add documentation what this is intended for.

maxteufel (approved):

PR #7539, opened 2026-08-14
rt: simplify `fzE_file_read` by using `read` instead of `fread`

michaellilltokiwa:
Also fread may do buffering internally which is probably not what we want.


include/posix.c
```diff
@@ -1036,31 +1036,10 @@ int32_t fzE_file_read(void * file, void * buf, int32_t size)
...
-      while (fread_result == 0 && !feof((FILE*)file) && (ferror((FILE*)file) && errno == EAGAIN));  // if we got no data and no EOF, then repeat.
-      if (!ferror((FILE*)file) || errno == EAGAIN)
-        {
-          assert ( fread_result > 0 || feof((FILE*)file) );
-          result = fread_result;
-        }
+      do {
+        result = read(fileno(file), buf, (size_t)size);
+        assert( errno != EAGAIN );
+      } while (result < 0 && errno == EINTR);
```
fridis: nitpicking: stick with GNU style indentation...
```suggestion
      do 
       {
          result = read(fileno(file), buf, (size_t)size);
          assert( errno != EAGAIN );
       }
      while (result < 0 && errno == EINTR);
```

fridis (commented):
I am happy to see this code simplified. 

But I am also worried that we might break something: are there any files that we set to `non_blocking`? How does this change affect interactive i/o? I fear we might run into new problems then. 

Can you check if the communication between `feeze` and the `feeze_recorder` still works after your change?

include/posix.c
simonvonhackewitz: I think the indentation in the suggestion needs to be increased by another space 😄

michaellilltokiwa: @fridis feeze still seems to work. At least I managed to record and display something. But I'm not sure what the problem was?

@fridis  To your point of non-blocking IO. We don't support it yet and also did not support it before. stdin descriptor is actually set to non-blocking in `fzE_init`, this is the reason for the `poll` code before the read. Any non-blocking IO is currently `polled` for and thus waited until there is actual data available.

maxteufel:
return of #7177 in the test run of 6b10eb0

fridis (commented):
Still having the same concern since I originally changed this since `feeze`/`feeze_recorder` communication failed: 

> Can you check if the communication between feeze and the feeze_recorder still works after your change?

michaellilltokiwa:
@fridis In my short test, feeze still works.

fridis (approved):

PR #7540, opened 2026-08-14
base: various code simplifications

michaellilltokiwa:


modules/base/src/container/ps_map.fz
```diff
@@ -242,16 +242,9 @@ O(n log n).
...
-  #
-  public min option PK =>
-
-    min0 (m PK, l i64, r i64) PK =>
-      lk, _ := data[l]
-      if (m ≤ lk) m else lk
-
-    min (m PK, at i64, sz i64, tail i64) PK =>
-      if sz = i64.zero
+  public min_max(F type : Function PK PK i64 i64, minmax F) option PK =>
```
simonvonhackewitz: This feature needs a comment, if it's intentionally public. Even if not a comment would help.

simonvonhackewitz (approved):

simonvonhackewitz (changes requested):
meant to request changes

modules/base/src/container/ps_map.fz
michaellilltokiwa: I forgot to remove the visibility modifier. `min_max` was supposed to be private.

simonvonhackewitz (approved):
`min_max` is a good deduplication, which is easy to understand when comparing against the old code.

But I find it hard to understand what actually happens in either the old `min`/`max` features or the new `min_max`, at least without looking at any other code in `ps_map`.

PR #7548, opened 2026-08-17
fe: raise error if visibility is specified for type parameter

michaellilltokiwa:
This PR as a suggestion, since i currently don't see the point of controlling the visibility of type parameters. They can IMHO always be public since they are also marked as fixed and are never inherited.


fridis (approved):
looks good to me.

PR #7551, opened 2026-08-18
base: process, more detailed error message

michaellilltokiwa:
example:
`error(code=9): reading from stdout(desc=7) of process make int --environment-overrides --directory=build/tests/reg_issue7341 (PID: 39388).`


maxteufel (approved):
Okay. This might need adaptation in the future if we let developers set a process title.

PR #7556, opened 2026-08-18
ast: revert loop result propagation via setOuter

### LINKED ISSUE #7088: Unexpected propagation of expected type from effect instatement into nested loop

michaellilltokiwa:
We can't do it like this as can be seen in `aoc/2023/18/part2.fz`. In `aoc/2023/18/part2.fz` an inner loop is assigned to a field but gets propagated the outers result type which is wrong.

fixes #7088


maxteufel (approved):

simonvonhackewitz:
Does this fix #7088?

michaellilltokiwa:
@simonvonhackewitz  yes

PR #7558, opened 2026-08-18
rt: set `CLOEXEC` in `fzE_pipe_create`

michaellilltokiwa:
we don't want these to be inherited by child processes

### DIFF
--- a/include/posix.c
+++ b/include/posix.c
@@ -855,17 +855,27 @@ int fzE_pipe_create(int64_t * fds)
 {
   int pipefd[2];
 
-  if (pipe(pipefd) == -1)
+  // make sure no fork is done while we open pipes
+  fzE_lock();
+
+  int res = pipe(pipefd);
+
+  if (res == 0)
   {
-    return errno;
-  }
-  else
-  {
+    int readCloseExec = fcntl(pipefd[0], F_SETFD, FD_CLOEXEC);
+    int writeCloseExec = fcntl(pipefd[1], F_SETFD, FD_CLOEXEC);
+    assert(readCloseExec == 0);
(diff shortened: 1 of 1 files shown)


maxteufel (changes requested):
Use `pipe2` with flags instead of `pipe` and `fcntl` and lock.

michaellilltokiwa:
@maxteufel pipe2 is linux only AFAIK

maxteufel:
`pipe2` is `POSIX.1-2024`, according to the manual

michaellilltokiwa:
@maxteufel Ah okay, but we can currently use only up to POSIX 2008: `#define _POSIX_C_SOURCE 200809L`

maxteufel:
I had
```
$ cat ../ex_pipe.fz
max_test_time := time.duration.s 10

envir_vars := envir.Vars.env.map_of [ "PATH", "LANG" ]

Sequence.prefix ! outcome tokiwa.process_result =>
  seq := map x->x.as_string
  os.process.start seq.first.or_panic (seq.drop 1) envir_vars
            .bind (p -> tokiwa.record_process p Sequence.this.as_string max_test_time nil)

ex_pipe =>
  lm : mutate is
  _ := lm ! ()->
    os.pipe.for_writing _ lm (()->
        (io.buffered lm).writer.env.write "hello, world!\n" .or_panic
        (io.buffered lm).writer.env.flush.or_panic
        say "written") (r->
        x := !["/run/current-system/sw/bin/cat", "/dev/fd/$(r)"]
        say x)

_ := ex_pipe
```
(maybe old syntax pre-buffered reader API change) as a test example for pipes. I fear this would no longer work with `CLOEXEC`.

PR #7566, opened 2026-08-18
base: add effect `Runtime_Failure`

michaellilltokiwa:
The idea is the get some message about any failures happening at runtime. Currently some failures are silently ignored.

### DIFF
--- a/modules/base/src/Runtime_Failure.fz
+++ b/modules/base/src/Runtime_Failure.fz
@@ -1,0 +1,37 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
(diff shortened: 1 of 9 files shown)


maxteufel (approved):

modules/base/src/Runtime_Failure.fz
```diff
@@ -0,0 +1,37 @@
...
+#
+#  Source code of Fuzion standard library feature Runtime_Failure
+#
+# -----------------------------------------------------------------------
+
+
+# effect to register any failures happening at runtime.
+#
+# By default this just prints the error to stderr
+#
```
fridis, line 28: I think this needs some more documentation. As I understand, this is to report errors that are essentially ignored since they are unexpected or the API does not permit reporting them. Would be good to give an example how to use this or how to suppress the error output or to turn it into a fault. 

Might also be good to find a better name such as `Ignored_Failure` or similar.

PR #7578, opened 2026-08-20
rt: add `fzE_process_close`

michaellilltokiwa:
On windows the process HANDLE needs to be freed.
We need to do this independently from `fzE_process_poll`


michaellilltokiwa:
@tokiwa-software/developers failing tests here are IMHO okay/expected

maxteufel (approved):

PR #7585, opened 2026-08-20
base: documentation fixes found by AI

michaellilltokiwa:


modules/base/src/float.fz
```diff
@@ -181,14 +181,14 @@ public float : numeric is
...
-  # hyperbolicus
+  # hyperbolic
 
   public sinh float.this => abstract
   public cosh float.this => abstract
   public tanh float.this => abstract
 
 
-  # area hyperbolicus
+  # area hyperbolic
```
maxteufel: if we're going for Latin to English we should also change "area" to "inverse"

modules/base/src/os/process.fz
```diff
@@ -77,7 +77,7 @@ module:public process(public pid i64, module std_in, std_out, std_err i64, name
     (io.buffered LM).Writer.instate (std_in_writer LM) f
 
 
-  # pipe this processes output to new process
+  # pipe this process's output to new process
```
maxteufel: process's is always wrong. one of processes or process' or processes' is correct.

maxteufel (changes requested):

examples/jitter/jitter_threaded.fz
simonvonhackewitz:
```suggestion
# jitter_threaded -- demo measuring periodic task jitter using time.histogram
```

modules/base/src/state.fz
```diff
@@ -61,7 +61,7 @@
 #
 public state(
 
-  # types of contained and the state value
+  # types of the contained value and the state
```
simonvonhackewitz: Maybe splitting this up with two separate comments would be better?

modules/base/src/state.fz
```diff
@@ -111,10 +111,7 @@ is
     state val new mode.next
 
 
-  # converts option to a string
-  #
-  # returns the result of $T for an option containing an instance
-  # of T, alternatively returns $nil for an option that is nil.
+  # converts state to a string
```
simonvonhackewitz: Maybe *convert* is not the best word? How about "create a string representation"

modules/base/src/state.fz
```diff
@@ -61,8 +61,11 @@
 #
 public state(
 
-  # types of contained and the state value
-  T, S type,
+  # type of the contained value
+  T,
```
simonvonhackewitz:
```suggestion
  T type,
```

simonvonhackewitz (approved):

maxteufel (approved):

PR #7593, opened 2026-08-21
base: add least-recently-used cache `container.LRU_Cache`

### LINKED ISSUE #7273: LRU Cache
We have a need for this in fzweb but also LRU is commonly needed. Would be good to have this.

simonvonhackewitz:
fix #7273


simonvonhackewitz:
Any thoughts on the NYIs?

modules/base/src/container/LRU_Cache.fz
```diff
@@ -0,0 +1,313 @@
...
+
+
+  # remove value for given key from the cache,
+  # returns the value if the key existed
+  #
+  public remove(key K) option V => # NYI: maybe call this invalidate?
+    LM.env.exclusive ()->
+      map.get key .bind i->{
+        res := values[i].or_panic
+        # values[i] := nil # NYI: necessary to overwrite removed entries, e.g. for privacy?
```
michaellilltokiwa: I think this probably does not hurt but in theory it should not be possible to access these anymore, correct?

modules/base/src/container/LRU_Cache.fz
```diff
@@ -0,0 +1,313 @@
...
+    LM.env.exclusive ()->
+      map.get key .bind i->(
+        move_first i
+        values[i].or_panic)
+
+
+  # remove value for given key from the cache,
+  # returns the value if the key existed
+  #
+  public remove(key K) option V => # NYI: maybe call this invalidate?
```
michaellilltokiwa: `invalidate` seems the better name, but I'm also okay with remove.

modules/base/src/container/LRU_Cache.fz
```diff
@@ -0,0 +1,313 @@
...
+    pre debug : 0 <= i < capacity
+  =>
+    pred := prev i
+    succ := next i
+
+    set_next pred succ
+    set_prev succ pred
+    if i = mru.get then mru <- succ
+    if i = lru.get then lru <- pred
+    nxtprv[i] := (i64 -1, i64 -1) # NYI: OPTIMIZATION: not really necessary, but might help to catch errors early
```
michaellilltokiwa: I don't understand this `NYI`, how is this an optimization and why is this an `i64` then and not e.g. `option u64`

modules/base/src/container/LRU_Cache.fz
```diff
@@ -0,0 +1,313 @@
...
+
+
+  # create an empty LRU cache with the given capacity
+  #
+  public type.new(I type : integer, capacity I) container.LRU_Cache LM K V
+    pre safety : 0 < capacity
+  =>
+    # NYI: OPTIMIZATION: should Mutable_Hash_Map provide an option to set minimum size to avoid reallocation when filling the cache?
+    m  := Mutable_Hash_Map LM K i64 .empty
+    ks := LM.env.new_array (option K) i64 capacity.as_i64 (_->nil) # NYI: array index type required here for lambda type inference
```
michaellilltokiwa: You can remove the `NYI` here. It is okay that type has to be specified here. Inference in this case is very hard.

michaellilltokiwa (changes requested):
Great stuff, one thing that is a blocker for me, the time running the actual test: 
JVM: `96s build/tests/lru_cache`
Should be less than 10sec, maybe with an option to run the long-form of the test.

modules/base/src/container/LRU_Cache.fz
simonvonhackewitz: yes

modules/base/src/container/LRU_Cache.fz
simonvonhackewitz: It is an additional array access, that is not strictly necessary. When no other element points to this one, it does not matter where the pointers of this one point to.

It is an `i64` because that is the default array index/length type, but also because it is convenient and avoids boxing.

michaellilltokiwa:
```suggestion
    if debug
      nxtprv[i] := (i64 -1, i64 -1)
```

michaellilltokiwa (approved):

PR #7597, opened 2026-08-24
base: allow index operations on `array2`/`3` without type parameter

### LINKED ISSUE #7543: index operations of multi-dimensional arrays should not require type parameter
I have just seen this change in #7537 ([reg_issue2380.fz](https://github.com/tokiwa-software/fuzion/pull/7537/changes#diff-212aefe4328ff07ee1aea490484d555e564f6f404c5ccd35c6ddb321c74387b3)):
```
-          c := area[y,x]
+          c := area[_,y,x]
```
Requiring a type parameter `_` here is IMHO not acceptable. 
I see two possible solutions

- hard-coding these operations to use a specific type like `i64` or `i32 | i64 | u32 | u64`
- make `array2` not inherit from `array`, but instead declare an inner field `base : array` and remove `index [] (one argument)`. Then, calls accessing the one-dimensional array would need to be written `array2.base[i]`, while accesses using two indices can still use `array2[x,y]` without a type parameter. 

I tend to prefer the second solution since that give a somewhat clearer distinction between an `array2` and the underlying one-dimensional array. 

The same holds for `array3`.

simonvonhackewitz:
Change `array2`/`3` to not inherit from `array`, instead define an inner field `base` of type `array`.

fix #7543

### DIFF
--- a/modules/base/src/array2.fz
+++ b/modules/base/src/array2.fz
@@ -33,7 +33,7 @@
 private:public array2(
        T type,
        public length0, length1 i64,
-       public base array T,
+       module base array T,
        _ unit,
        _ unit) : property.hashable
   pre
--- a/modules/base/src/array2.fz
+++ b/modules/base/src/array2.fz
@@ -33,17 +33,14 @@
 private:public array2(
        T type,
        public length0, length1 i64,
-       module redef internal_array fuzion.sys.internal_array T,
+       public base array T,
        _ unit,
-       _ unit)
- : array T internal_array unit unit unit unit
+       _ unit) : property.hashable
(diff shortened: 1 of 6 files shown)


modules/base/src/array2.fz
```diff
@@ -33,18 +33,15 @@
 private:public array2(
        T type,
        public length0, length1 i64,
-       module redef internal_array fuzion.sys.internal_array T,
+       public base array T,
```
michaellilltokiwa:
```suggestion
       module base array T,
```

michaellilltokiwa (approved):

PR #7598, opened 2026-08-24
win: fix mod_sqlite test

michaellilltokiwa:
Tested on my machine, seems to work:

image

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -61,12 +61,11 @@ jobs:
           install: >-
             make
             git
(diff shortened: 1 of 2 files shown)


simonvonhackewitz:
Why did the Windows test not run in this PR?

michaellilltokiwa:
@simonvonhackewitz They did, seems to be a GitHub problem. Click checks and go to windows tests from there. But it is of no use anyway since windows is broken currently.

simonvonhackewitz (approved):

Issue #7423, opened 2026-07-21
add `interval.reverse` and `interval.last`


michaellilltokiwa:
```
for i in 1..1000 do
  _ := codepoint codepoint.utf8_encoded_in_four_bytes.as_list.last.or_panic
```
takes 37s     on my machine with jvm be 
takes >2min on my machine with c     be

fridis:
This is not that surprising since `codepoint.utf8_encoded_in_four_bytes` contains 65536 codepoints, so we have a total of 65 million codepoints to walk thropugh, 37s means that each takes about 500ns or >2µs for the C backend. 

What can be improved is that omitting `.as_list` does not make this any faster. `interval.last` could probably be implemented more efficiently.

michaellilltokiwa:
We may want `interval.reverse`?

fridis:
> We may want `interval.reverse`?

both, `interval.reverse` and `interval.last`. And maybe more.

michaellilltokiwa:
@simonvonhackewitz Might be a good first issue for Arthur.

PR #7600, opened 2026-08-24
interval: add O(1) first and last (#7423)

arturdmi:
Part of task #7423.

The `first` and `last` methods were inherited from `Sequence` and used `as_list`, which resulted in memory allocation for every element. Since the range represents an arithmetic progression, both values ​​can be calculated directly:

first = from
last  = from + ((through - from) / step) * step

The division operation truncates the fractional part, yielding the last element actually included in the range. For example, with `1..10:4`, the result is 9 rather than 10, as the sequence of elements is 1, 5, 9.

I used the same conditional cases as in the `size` method: the `from == through` case, the empty range case, and the general case.

JVM benchmark results (1000 iterations): 78.7 s → 2.9 s (empty program: 2.7 s).
The code example demonstrating the issue in the original ticket uses `.as_list.last` and is unaffected by these changes, as `as_list` constructs the list before `last` is called.

---

On `reverse`: the obvious version is `interval last from -step`, but `T` can be
unsigned and `interval u32` is exactly the case in the issue. As far as I can
tell there's no way to ask whether negation is meaningful for a given `T` —
`prefix -!` is `=> true` in `numeric.fz` and never overridden, `prefix -?` never
returns `nil`, and `u32`/`i32` only override the wrapping `-°`.

So should `reverse` return a lazy `Sequence` instead of an `interval`, or would
you rather have a proper predicate in `numeric` for this? Happy to do either.


modules/base/src/interval.fz
```diff
@@ -97,6 +97,45 @@ is
         T.from_u32 as_list.count.as_u32
 
 
+  # get the first element of this interval
+  #
+  # Performance: O(1)
+  #
+  public redef first option T
```
michaellilltokiwa: I think this is `O(1)` even without this `redef`?

tests/interval_last/test_interval_first_last.fz
```diff
@@ -17,34 +17,43 @@
...
+
+# negative step, through is not an element: 10,6,2
+say (i32 10 .. 1 : -4).last
+
+# empty intervals, see the docstring of interval
+say (interval 2 1 1).last
+say (interval 1 2 -1).last
+
+# large interval, must be O(1)
+say (u32 1 .. 1000000).last
```
michaellilltokiwa:
```suggestion
check (u32 1 .. u32.max).last = u32.max
```

tests/interval_last/test_interval_first_last.fz
```diff
@@ -17,34 +17,43 @@
...
-
-        if len < 0
-          error "hostname failed" fzE_last_error
-        else
-          String.from_bytes (arr.slice 0 len)
+
+# --- last ---
+
+# through is an element of the interval
+say (u32 1 .. 10).last
```
michaellilltokiwa: I would suggest to rather do it like this than via `say`. 
```suggestion
check (u32 1 .. 10).last = 10
```

michaellilltokiwa:
Good, the PR description seems unredacted AI though. I don't understand all of it, e.g.:

> Same three-case structure as size.

michaellilltokiwa:
I think, if possible `reverse` should return an `interval`.

modules/base/src/interval.fz
```diff
@@ -97,6 +97,61 @@ is
...
+  public redef last option T
+  =>
+    match through
+      thru T =>
+        if from = thru then
+          from
+        else if step.sign = (from ⋄ thru).sign then
+          nil
+        else
+          from + (((thru - from) / step) * step)
```
simonvonhackewitz: This might be slightly easier to understand
```suggestion
          thru - (thru - from) % step
```

modules/base/src/interval.fz
```diff
@@ -97,6 +97,61 @@ is
...
+  #
+  # Performance: O(1)
+  #
+  public redef first option T
+  =>
+    match through
+      thru T =>
+        if from = thru then
+          from
+        else if step.sign = (from ⋄ thru).sign then
```
simonvonhackewitz: This could be simplified, since two branches return the same result
```suggestion
        if from != thru && step.sign = (from ⋄ thru).sign then
```

simonvonhackewitz:
> I think, if possible `reverse` should return an `interval`.

Maybe it is a design flaw that negative steps are not possible with unsigned integers? Subtraction would be possible as long as the result does not become negative.

arturdmi:
Suggestion: treat the direction as a separate field and keep the step always positive. This would make the interval `u32` reversible. However, this change would affect many other parts of the code.
All `u32` intervals in the `base` module—specifically in `codepoint.fz`—are ascending. The reverse unsigned interval is currently not used or tested anywhere.That's why I don't know if we need this.

arturdmi:
There is already an "interval" folder in the tests. Should i add the tests for the new features there as well?

modules/base/src/interval.fz
```diff
@@ -97,6 +97,61 @@ is
...
+  # The i-th element is `last - i*step`, which works for positive and
+  # negative `step` without requiring negation, so it also works for
+  # unsigned `T`.
+  #
+  public redef reverse Sequence T
+  =>
+    match last
+      l T =>
+        n := size.as_i32
+        (0 .. n-1).map (i -> l - (T.from_u32 i.as_u32)*step)
```
fridis: This does not work for, e.g., `interval (i64 1000000000000 1000000000010) .reverse`. 

Try something like this

```suggestion
        interval l first -step
```

fridis (changes requested):
Looks pretty good. 
Below my suggestion to avoid depending on `u32`/`i32`

modules/base/src/interval.fz
```diff
@@ -97,6 +97,63 @@ is
...
+  #
+  # Performance: O(1)
+  #
+  # The i-th element is `last - i*step`, which works for positive and
+  # negative `step` without requiring negation, so it also works for
+  # unsigned `T`.
+  #
+  public  reverse_interval interval T 
+  pre
+    safety: T.zero -! step
```
fridis: Maybe a comment would be good here: 

```suggestion
    safety: T.zero -! step   # We must be able to negate `step`, so `T` must be signed
```

fridis (changes requested):
very nice now!

Just a request for a comment since `infix -!` is probably not an obvious operator known by everyone.

modules/base/src/interval.fz
```diff
@@ -97,6 +97,63 @@ is
...
+
+  # reverse the order of the elements in this interval
+  #
+  # Performance: O(1)
+  #
+  # The i-th element is `last - i*step`, which works for positive and
+  # negative `step` without requiring negation, so it also works for
+  # unsigned `T`.
+  #
+  public  reverse_interval interval T
```
fridis:
```suggestion
  public reverse_interval interval T 
```

fridis:
> I think, if possible `reverse` should return an `interval`.

I think `Sequence T` is fine, we might think about adding `reverse` to `Sequence` as well, so then the result should be `Seqeuence T`.

fridis:
> Suggestion: treat the direction as a separate field and keep the step always positive. This would make the interval `u32` reversible. However, this change would affect many other parts of the code. All `u32` intervals in the `base` module—specifically in `codepoint.fz`—are ascending. The reverse unsigned interval is currently not used or tested anywhere.That's why I don't know if we need this.

I would not add an extra field and tests of that field since that would add overhead to all intervals (and they tend to be used often in performance critical code).

fridis:
> There is already an "interval" folder in the tests. Should i add the tests for the new features there as well?

This is up to you. I would prefer adding the new tests to `tests/interval` if possible.

arturdmi:
> > I think, if possible `reverse` should return an `interval`.
> 
> I think `Sequence T` is fine, we might think about adding `reverse` to `Sequence` as well, so then the result should be `Seqeuence T`.

`Sequence.reverse` already exists (Sequence.fz:320, `=> as_list.reverse_list`),
so I want to make sure I understood you correctly — did you mean rewriting
that one, or adding `reverse` somewhere else?

modules/base/src/interval.fz
```diff
@@ -97,6 +97,63 @@ is
...
+  # Performance: O(1)
+  #
+  # The i-th element is `last - i*step`, which works for positive and
+  # negative `step` without requiring negation, so it also works for
+  # unsigned `T`.
+  #
+  public reverse_interval interval T 
+  pre
+    safety: T.zero -! step # We must be able to negate `step`, so `T` must be signed
+    debug: through.exists && first.exists
```
michaellilltokiwa: First does not need to exist, I think? The reverse of an empty interval is also just an empty interval

modules/base/src/interval.fz
```diff
@@ -97,6 +97,63 @@ is
...
+  =>
+    match last
+      l T =>
+        interval l first -step
+      nil =>
+        interval T.zero T.zero T.zero
```
michaellilltokiwa:
```suggestion
        panic "reversing an unbounded interval is not possible"
```

maxteufel:
`Sequence.reverse` gets an automatic implementation due to `interval` being a `Sequence`. However, I did not see a way to make this O(1), since preconditions are _contravariant_, we cannot _add_ the restriction of only signed integers here. That's why I suggested to @arturdmi to make this a separate feature.

modules/base/src/interval.fz
```diff
@@ -97,6 +109,42 @@ is
...
+  # Performance: O(1)
+  #
+  public reverse_interval interval T
+  pre
+    safety: T.zero -! step
+    debug: through.exists && first.exists
+  =>
+    match last
+      l T =>
+        interval l first -step
```
maxteufel, line 143: you removed `first` again, so how is this supposed to work?

michaellilltokiwa: @maxteufel only the redefinition was removed, `first` is inherited anyway.

arturdmi: first is not defined in interval.fz — it comes from Sequence.first (Sequence.fz:189), which is as_list.first, and list.first (list.fz:182) just returns the head.

maxteufel: Yes; however, `as_list` is an O(n) complex operation. This means that the complexity of `reverse_interval` also goes up to O(n) from O(1) before. Then we can just use `Sequence.reverse` instead.

arturdmi:
`as_list` treated `interval x x 0` as empty.

Now `from = thru` is excluded from the check and the recursion stops for a zero step.
Refs #7602
```
➜  fuzion git:(issue/7423) cat > /tmp/z.fz <<'EOF'
i := interval (i32 5) (i32 5) (i32 0)
say i.size
say i.count
say i.as_list
say i.is_empty
say (i.contains 5)
EOF
fz /tmp/z.fz

Output:
1
1
[5]
false
true
```

modules/base/src/interval.fz
```diff
@@ -24,58 +24,78 @@
...
+    debug: ((step_factor.sign = 0): through=from) # `step_factor` cannot be zero unless `from == through`, because a zero step would never reach `through`.
     safety: (step *? step_factor).exists
   =>
     interval from through step*step_factor
 
 
-  # list representation of values in this interval
+  # the values of this interval as a list, in order
+  #
+  # Returns an empty list if `step` moves in the opposite direction of `through`.
```
maxteufel, line 81: why not just return the interval as a list but in reverse?

arturdmi: Returning the elements in reverse would make as_list.count ≠ size and include elements contains rejects  same as in #7602.

arturdmi: `interval 0 -1 1 ` has to stay empty, not become [-1, 0].
Descending order is already expressible as interval 2 1 -1 or .reverse

maxteufel (approved):

fridis (approved):
good.

PR #7603, opened 2026-08-25
base: allow access to one dimensional base array in `array2`/`3`

simonvonhackewitz:
follow up to #7597

### DIFF
--- a/modules/base/src/array2.fz
+++ b/modules/base/src/array2.fz
@@ -36,6 +36,15 @@ private:public array2(
 
       # the underlying one dimensional array
       #
+      # multidimensional indices are mapped to one dimensional indices as follows:
+      #
+      # [x,y] -> [x * length1 + y]
+      #
+      # resulting in the order:
+      #
+      # arr[0,0], arr[0,1], arr[0,2], ..., arr[1,0], arr[1,1], arr[1,2], ...,
+      # arr[2,0], arr[2,1], arr[2,2], ...
+      #
       public base array T,
       _ unit,
       _ unit) : property.hashable
--- a/modules/base/src/array2.fz
+++ b/modules/base/src/array2.fz
@@ -31,11 +31,14 @@
 # two index parameters.
 #
 private:public array2(
-       T type,
-       public length0, length1 i64,
-       module base array T,
-       _ unit,
-       _ unit) : property.hashable
+      T type,
+      public length0, length1 i64,
+
+      # the underlying one dimensional array
+      #
+      public base array T,
+      _ unit,
+      _ unit) : property.hashable
   pre
     safety: length0 ≥ 0
     safety: length1 ≥ 0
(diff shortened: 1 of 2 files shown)


simonvonhackewitz:
I'm open to suggestions for a better name.

modules/base/src/array2.fz
```diff
@@ -129,6 +129,11 @@ is
     array2.this.base[i1.as_i64 * length1 + i2.as_i64]
 
 
+  # returns this two dimensional array as a one dimensional one by flattening it
+  #
```
fridis: This suggests that this has performance `O(length0*length1)`, while it has `O(1)`. Better say

```suggestion
  # returns the underlying one dimensional array 
  #
```

modules/base/src/array3.fz
```diff
@@ -133,6 +133,11 @@ is
     array3.this.base[(i0.as_i64 * length1 + i1.as_i64) * length2 + i2.as_i64]
 
 
+  # returns this three dimensional array as a one dimensional one by flattening it
```
fridis: As above, should not suggest that this is expensive.

modules/base/src/array2.fz
```diff
@@ -129,6 +129,11 @@ is
...
+  public as_flat array T => base
```
fridis: Maybe `base_array` would be a better name? Or do not add any feature but make `base` public?

fridis (changes requested):
I do not think that we really need to hide the underlying array implementation, we could just make `base` public and add the documentation to that field.

modules/base/src/array2.fz
```diff
@@ -31,11 +31,14 @@
...
-       T type,
-       public length0, length1 i64,
-       module base array T,
-       _ unit,
-       _ unit) : property.hashable
+      T type,
+      public length0, length1 i64,
+
+      # the underlying one dimensional array
+      #
```
fridis, line 38: I think it would be good to explain how the indices are mapped, i.e., does 

`base[0]`,  `base[1]`,  `base[2]`, ..., `base[length0-1]`, `base[length0]`, ...

correspond to

`array2.this[0,0]`, `array2.this[1,0]`, `array2.this[2,0]`, ..., `array2.this[length0-1,0]`, `array2.this[0, 1]`, ...

or does  

`base[0]`,  `base[1]`,  `base[2]`, ..., `base[length1-1]`, `base[length1]`, ...

correspond to

`array2.this[0,0]`, `array2.this[0,1]`, `array2.this[0,2]`, ..., `array2.this[0,length1-1]`, `array2.this[1,0]`, ...

?

modules/base/src/array3.fz
```diff
@@ -31,11 +31,14 @@
...
-       public length0, length1, length2 i64,
-       module base array T,
-       _ unit,
-       _ unit)
+      T type,
+      public length0, length1, length2 i64,
+
+      # the underlying one dimensional array
+      #
+      public base array T,
```
fridis, line 49: A similar comment on the mapping of indices as above would be good.

fridis (changes requested):
I like this simpler solution!

Only would like to see some more documentation on how `base` maps to the 2-/3-dimensional array, see details below.

fridis (approved):
nice!

PR #7605, opened 2026-08-26
be/int: disable 7572, too slow

michaellilltokiwa:
[ci skip]


tests/reg_issue7572/skip_int
```diff
@@ -0,0 +1 @@
+NYI: PERFORMANCE: too slow
```
maxteufel: newline missing

maxteufel (changes requested):

tests/reg_issue7572/skip_int
michaellilltokiwa:
```suggestion
NYI: PERFORMANCE: too slow

```

simonvonhackewitz (approved):

maxteufel (approved):

PR #7608, opened 2026-08-26
base: `read_fully`, `read_lines`, `read_delimiter_full` return `outcome`

michaellilltokiwa:
instead of panicking


maxteufel:
I am not entirely a fan of this; handling `outcome` requires a lot of boilerplate code. You can just choose to catch a panic on an outer level instead.

michaellilltokiwa:
@maxteufel You can always do `or_panic` on an `outcome`, not that much boilerplate, IMHO. On the other hand catching a `panic` does require some boilerplate.

@tokiwa-software/developers What is your opinion on this particular API design?

simonvonhackewitz:
I think outcome clearly indicates that this might fail. The panic you only notice if you do the effect analysis or check the source code.

michaellilltokiwa:
@tokiwa-software/developers please review (again)

maxteufel (approved):

PR #7610, opened 2026-08-26
net/ip_address: implement `property.orderable`

maxteufel:
This is necessary for use in a `container.Set`.

### DIFF
--- a/modules/base/src/net/ip_address.fz
+++ b/modules/base/src/net/ip_address.fz
@@ -27,7 +27,7 @@
 # in the case of an IPv4 address, this is the canonical mapping of that
 # address to an IPv6 address.
 #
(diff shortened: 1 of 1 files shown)


modules/base/src/net/ip_address.fz
```diff
@@ -27,7 +27,7 @@
 # in the case of an IPv4 address, this is the canonical mapping of that
 # address to an IPv6 address.
 #
-public ip_address (val u128) : property.equatable is
+public ip_address (val u128) : property.equatable, property.orderable is
```
fridis: `property.equatable` is redundant: 

```suggestion
public ip_address (val u128) : property.orderable is
```

fridis (approved):

PR #7616, opened 2026-08-26
clang: set `C_INCLUDE_PATH` when generating `clang/src/native.fz`

michaellilltokiwa:


maxteufel (approved):

maxteufel:
what specifically is this for?

michaellilltokiwa:
This is for relative includes in some header files like `#include <clang/some_file.h>`, The include folder needs to be correct for clang to find them.

PR #7624, opened 2026-08-27
base: add `switch.otherwise`

### LINKED ISSUE #7609: Add `switch.otherwise` or similar feature to handle error case
Saw the following code in #7576 
```
                  match (io.buffered lm).Writer.env.write s
                    unit =>
                    e error => Runtime_Failure.env.register_failure e
```
Would be nice to have a variant of `or_else` in `switch` such that we can do 
```
                  (io.buffered lm).Writer.env.write s
                    .otherwise Runtime_Failure.env.register_failure
```
or similar. 

CAVEAT: There are already many `or`, `or_else` and similar features in `switch`. Adding one more requires some thought on consistent and intuitive naming!

arturdmi:
Closes #7609.

### DIFF
--- a/modules/base/src/switch.fz
+++ b/modules/base/src/switch.fz
@@ -103,11 +103,25 @@ is
                 |   B => default
 
 
-
+                 
+  # value of a switch, or, if this switch contains a `B`, the
+  # result of `f` applied to that `B`.
+  #
+  # in contrast to `or_else`, `f` receives the `B` value, so this can be
+  # used to handle a failure without a `match`, e.g. to log an error
+  # that cannot be propagated:
+  #
+  # example:
+  #
+  #     (fuzion.sys.fileio.close fd)
+  #       .otherwise Ignored_Failure.env.register_failure
+  #
   public otherwise(f B -> A) A =>
     match switch.this
       a A => a
       b B => f b
+
+
   # converts switch into a list of either a single element in case
   # switch.this.exists or `nil` otherwise
   #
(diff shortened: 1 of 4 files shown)


modules/base/src/switch.fz
```diff
@@ -104,6 +104,10 @@ is
 
 
 
+  public otherwise(f B -> A) A =>
```
simonvonhackewitz, line 119: Public features always need a comment

modules/base/src/switch.fz
```diff
@@ -104,6 +104,10 @@ is
...
+    match switch.this
+      a A => a
+      b B => f b
```
simonvonhackewitz, line 122: add some newlines here for readability

```suggestion
      b B => f b
      
      
```

tests/lib_switch/lib_switch.fz
```diff
@@ -30,3 +30,16 @@ lib_switch =>
   say (b.and_then x->x+5)
   say (a.or ()->5)
   say (b.or ()->5)
+
+  say (a.otherwise _->5)
```
simonvonhackewitz, line 34: When making changes to a test that affect its output, you need to re-record the expected output/error; otherwise, the test will fail. There is a Makefile target for this.

simonvonhackewitz (changes requested):

michaellilltokiwa (approved):
Perfect!

simonvonhackewitz (approved):

PR #7633, opened 2026-08-28
fix #7626: use position of value expression for loop index var actuals

### LINKED ISSUE #7626: Type inference failure for loop index var shows wrong positions of expressions
Try this, which shows three times the same source code position:
```
 > FUZION_DISABLE_ANSI_ESCAPES=true ./build/bin/fz -e 'for i := 3, true do' 

command line:1:5: error 1: Type inference from actual arguments failed due to incompatible types of actual arguments
for i := 3, true do
----^
For the formal argument 'loop.i' the following incompatible actual arguments where found for type inference:
actual is value of type 'i32' at command line:1:5:
for i := 3, true do
----^
actual is value of type 'bool' at command line:1:5:
for i := 3, true do
----^

one error.
```
It should better look like this:
```
 > FUZION_DISABLE_ANSI_ESCAPES=true ./build/bin/fz -e 'for i := 3, true do' 

command line:1:5: error 1: Type inference from actual arguments failed due to incompatible types of actual arguments
for i := 3, true do
----^
For the formal argument 'loop.i' the following incompatible actual arguments where found for type inference:
actual is value of type 'i32' at command line:1:10:
for i := 3, true do
---------^
actual is value of type 'bool' at command line:1:13:
for i := 3, true do
------------^^^^

one error.
```

arturdmi:
Fixes #7626.

Loop index vars are passed to the tail recursive loop feature as actual
arguments. These actuals were built with the index var's position, so
every expression in the error message pointed at the same place. They
now take position and source range from the value expression.

For `for x in s` the value has no position in the source code after
addIterators, so the old position is kept.

reg_issue2347 expected_err updated: error 2 now points at
`s.filter x->d.bla` instead of at `s`.

Developed with the assistance of AI tools. All local tests passed.

### DIFF
--- a/src/dev/flang/ast/Loop.java
+++ b/src/dev/flang/ast/Loop.java
@@ -610,6 +610,39 @@ public class Loop extends ANY
   }
 
 
+   /**
+   * Create an actual argument that passes the value of an index var to the tail
+   * recursive loop feature.
+   *
+   * The source position and range are taken from the expression that defines
+   * the value such that error messages, e.g., for failed type inference, refer
+   * to that expression and not to the index variable's name.
+   *
+   * @param f an index var (for the initial value) or the corresponding feature
+   * from _nextValues (for the value used in the next iteration)
+   *
+   * @param prefix ITER_ARG_PREFIX_INIT or ITER_ARG_PREFIX_NEXT
+   *
+   * @param name the base name of the index var
+   *
+   * @param fallback the position to be used in case the value of f does not
+   * correspond to an expression in the source code, e.g., for an iterating
+   * index var ('for x in s') whose Impl was replaced in addIterators.
+   */
+  private Call iterArgActual(Feature f, String prefix, String name, SourcePosition fallback)
+  {
+    var e = f.impl().expr();
+    var useE = e != null && !e.pos().isBuiltIn();
+    var result = new Call(useE ? e.pos() : fallback, prefix + name);
+    // NYI: CLEANUP: set source range/pos in constructor call
+    if (useE && e.sourceRange() instanceof SourceRange r)
+      {
+        result.setSourceRange(r);
+      }
+    return result;
+  }
+
+
   /**
    * Helper routine to determine the formal and actual arguments to be passed to the tail recursive loop
    *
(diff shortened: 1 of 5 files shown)


src/dev/flang/ast/Loop.java
```diff
@@ -610,6 +610,38 @@ private Expr callLoopElse(int elseNum)
...
+   */
+  private Call iterArgActual(Feature f, String prefix, String name, SourcePosition fallback)
+  {
+    var e = f.impl().expr();
+    var useE = e != null && !e.pos().isBuiltIn();
+    var result = new Call(useE ? e.pos() : fallback, prefix + name); //NYI: CLEANUP
+    if (useE && e.sourceRange() instanceof SourceRange r)
+      {
+        result.setSourceRange(r);
+      }
```
simonvonhackewitz: How about this?

```suggestion
    var pos = !useE                                    ? fallback :
              e.sourceRange() instanceof SourceRange r ? r
                                                       : e.pos()
    var result = new Call(pos, prefix + name); //NYI: CLEANUP
```

src/dev/flang/ast/Loop.java
```diff
@@ -610,6 +610,38 @@ private Expr callLoopElse(int elseNum)
...
+   * @param fallback the position to be used in case the value of f does not
+   * correspond to an expression in the source code, e.g., for an iterating
+   * index var ('for x in s') whose Impl was replaced in addIterators.
+   */
+  private Call iterArgActual(Feature f, String prefix, String name, SourcePosition fallback)
+  {
+    var e = f.impl().expr();
+    var useE = e != null && !e.pos().isBuiltIn();
+    var result = new Call(useE ? e.pos() : fallback, prefix + name); //NYI: CLEANUP
+    if (useE && e.sourceRange() instanceof SourceRange r)
```
michaellilltokiwa:
```suggestion
    var result = new Call(useE ? e.pos() : fallback, prefix + name);
    // NYI: CLEANUP: set source range/pos in constructor call
    if (useE && e.sourceRange() instanceof SourceRange r)
```

michaellilltokiwa (approved):

PR #7635, opened 2026-08-28
base: change `switch.postfix !` to return A or `panic`

michaellilltokiwa:


modules/base/src/switch.fz
```diff
@@ -54,7 +54,7 @@ is
 
   # shorthand postfix operator for '!exists'
```
simonvonhackewitz:
```suggestion
  # shorthand postfix operator for 'or_panic'
```

simonvonhackewitz (approved):

PR #7644, opened 2026-08-28
base: implement `html.decode`

michaellilltokiwa:
This was mostly done by opencode BigPickle with minor manual corrections


modules/base/src/encodings/html.fz
```diff
@@ -102,7 +102,56 @@ public html is
       repl map_of_escapings
 
 
+  # decode html encoded string back to original
```
simonvonhackewitz, line 105: Does this follow any particular standard? How does it handle broken or non-standard input, e.g., `&`, `&#39` or `&39393939;`.

modules/base/src/encodings/html.fz
```diff
@@ -102,7 +102,56 @@ public html is
...
+    # map entity name (without the surrounding '&' and ';') to the byte
+    # it encodes; nil if the entity is unknown
+    #
+    decode_entity(e String) option u8 =>
+      if e = "amp"   then "&".val.as_u8
+      else if e = "lt"  then "<".val.as_u8
+      else if e = "gt"  then ">".val.as_u8
+      else if e = "quot" then "\"".val.as_u8
+      else if e = "#39"  then "'".val.as_u8
+      else if e = "#x2F" || e = "#x2f" then "/".val.as_u8
```
simonvonhackewitz: These are just hardcoded, from the feature comment I would have expected `&#65;` to decode to `A`

michaellilltokiwa: I did not really notice the comment and thought it is fine when it can reverse our encode but I fixed the slop now.

maxteufel (approved):
nice

simonvonhackewitz:
It works for combined codepoints
```
❯ fz -e 'say (encodings.html.decode "&#9881;&#65039; \n&#9881;&#65038; \n&#128104;&#8205;&#128187; \n&#1114111;")'
⚙️ 
⚙︎ 
👨‍💻 
􏿿
```

But causes a precondition failure when exceeding the valid codepoint range
```
❯ fz -e 'say (encodings.html.decode "&#1114112;")' | cat                                                                     

error 1: FATAL FAULT `precondition`: debug: codepoint.range.contains val
Call stack:
fuzion.sys.fatal_fault#2 at {base.fum}/fuzion/sys.fz:55
[..]
fuzion.runtime.precondition_fault#1 at {base.fum}/fuzion/runtime/pre_fault.fz:56
pre codepoint at {base.fum}/codepoint.fz:30
codepoint at {base.fum}/codepoint.fz:29
encodings.html.decode#1.decode_entity#1.(λcodepoint).call#1 at {base.fum}/encodings/html.fz:-1
(outcome u32).bind#2 codepoint at {base.fum}/outcome.fz:111
encodings.html.decode#1.decode_entity#1 at {base.fum}/encodings/html.fz:134
encodings.html.decode#1.dec#2 at {base.fum}/encodings/html.fz:1
encodings.html.decode#1 at {base.fum}/encodings/html.fz:156
universe at --builtin--:1

one error.
```

For our needs this is probably enough at the moment, but maybe add a NYI comment about checking compliance with a standard.
From a quick search the [WHATWG HTML Living Standard](https://html.spec.whatwg.org/multipage/) seems to be the current standard defining these.

Also, it might depend on what it is used for, e.g., browsers tend to be as tolerant as possible, I think, so they might render `&#65` (missing semicolon) even though this could be considered broken encoding.

simonvonhackewitz:
nice :+1:

PR #7655, opened 2026-08-31
base: change `Type.as_string` to omit `Type of...`

### LINKED ISSUE #6865: two APIs for Type.name,  one for a human readable type that essentially maps to AbstractType.toString(true) and one that shows the internal name like AbstractType.toString(false).
Toke this code 
```
> ./build/bin/fz -e "say (type_of 42)"
Type of 'i32'
```
would maybe be nicer if this just produced
```
> ./build/bin/fz -e "say (type_of 42)"
i32
```
It might make sense to provide a means to get the internal type name as well, i.e. to have two APIs, one for a human readable type that essentially maps to `AbstractType.toString(true)` and one that shows the internal name like `AbstractType.toString(false)`.

michaellilltokiwa:
fixes #6865


simonvonhackewitz (approved):
What about the second part of the issue?

PR #7657, opened 2026-08-31
ast: fix#7415: adjust type of constraints in a call when infering type pars

### LINKED ISSUE #7415: Type propagate from actual type parameter of outer feature to inferred lambda type does not work
Try this:
```
x(T type) is
  y(f T->T) =>
  z(F type : T->T, f F) =>

x i32 .y +3   # works
x i32 .z +3   # fails
```
it works fine for the call to `y`, but fails to properly propagate the actual type `i32` for `T` when calling `z`:
```
 > FUZION_DISABLE_ANSI_ESCAPES=true ./build/bin/fz type_infer_test.fz

/home/fridi/fuzion/fuzion/type_infer_test.fz:6:10: error 1: Could not find called feature
x i32 .z +3   # fails
---------^
Feature not found: 'infix +' (one argument)
Target feature: 'Any'
In call: '+'


/home/fridi/fuzion/fuzion/type_infer_test.fz:6:8: error 2: Incompatible type parameter
x i32 .z +3   # fails
-------^
formal type parameter 'F : Unary T T' with constraint 'Unary i32 i32'
actual type parameter '(λ+3)'

2 errors.
```
This is particularly annoying for calls to, e.g. `concur.atomic.update`.

fridis:
The main effect this has is that now we can use type parameters for function
arguments even if the types of the funciton arguments are type parameters as in

    test(T type)
    is

      bla(F type : T->T, f F)
      =>
        _ := f()

Note that this depends on #7654 that should be merged first. 

This fixes #7415 and it repairs #6921, which was closed but the test should not fail with an error but work, which it does now.


michaellilltokiwa (approved):

maxteufel:
"The main effect this has is that now we can use type parameters for function
arguments even if the types of the funciton arguments are type parameters"

I don't really understand this sentence.

fridis:
> "The main effect this has is that now we can use type parameters for function arguments even if the types of the function arguments are type parameters"
> 
> I don't really understand this sentence.

This was my description of `x.z` in this example 
```
  x(T type) is
    y(f T->T) =>
    z(F type : T->T, f F) =>

  x i64 .y +3   # works
  x i64 .z +3   # fails
```
that used to fail due to wrong type inference and it works now.

PR #7667, opened 2026-09-01
fe: raise error when effect feature is called directly not via `env`

### LINKED ISSUE #4119: For effect operation `o`, require call via `effect_type.env.o`
This is related to #2626 and may help programmers spot obvious mistakes.

michaellilltokiwa:
fixes #4119


fridis (approved):
Nice. 

This has much less dramatic consequences than I had expected!

PR #7672, opened 2026-09-01
fe: fix require-condition in `isLambdaTarget`

michaellilltokiwa:
If the constraint is a parametric type again this will lead to problems. Changed selfOrConstraint to recurse in this case.
```
error 1: java.lang.Error: require-condition2 failed: AbstractType.java:237 "(!(this instanceof UnresolvedType), !isParametricType());"
        at dev.flang.util.ANY.require(ANY.java:156)
        at dev.flang.ast.AbstractType.feature(AbstractType.java:237)
        at dev.flang.ast.AbstractType.isLambdaTarget(AbstractType.java:1551)
        at dev.flang.ast.AbstractType.isLambdaTargetButNotLazy(AbstractType.java:1537)
```

### DIFF
--- a/src/dev/flang/ast/AbstractType.java
+++ b/src/dev/flang/ast/AbstractType.java
@@ -2808,7 +2808,7 @@ there is no common super type of the two types (Types.t_ERROR)
    */
   public AbstractType selfOrConstraint()
   {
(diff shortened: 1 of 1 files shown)


maxteufel (approved):

fridis:
Would be good to have an issue, example and a regression test!

PR #7673, opened 2026-09-01
interval:  fix `contains`, add O(1) `nth`, `index_of`, `size_option`, `equality`

### LINKED ISSUE #7663: Bug: Interval .contains() returns false for a valid element
It should return true (interval = {-3 -1 1 3})

```
➜  interval git:(interval) ✗ cat > /tmp/z.fz <<'EOF'              
say ((-3..3:2).contains 1)  
EOF
fz /tmp/z.fz
false
```

arturdmi:
**Fix `contains`.** It compared signed remainders (`e % step = from % step`),
which differ in sign across zero, so `(-3..3:2).contains 1` was `false`
although `1` is an element. Now `(e - from) % step = 0`.

**Add `size_option`.** `Set.finite` and `Set.type.equality` are built on it
(`container/Set.fz:36,182`) and its default is `nil`, so `(1..3) = (1..3)`
panicked.

**`nth` and `index_of` in O(1)** instead of walking `as_list`. On
`interval 0 10_000_000 1`, `nth 1_000_000`: 416 s → 6 ms.

**`type.equality`** compares elements, not fields: `interval 1 8 3 =
interval 1 7 3`. Deliberate: `1..10 ≠ 10..1:-1` — ordered sequence, not
unordered set. Say if the `Set` reading should win.

**Tests.** `test_interval.fz` gains columns for the new features and an
`inv` check that `nth` and `index_of` are inverse — holds for all 29
intervals. Existing expected output changes where `contains` was wrong.

[fixes #7663]


modules/base/src/interval.fz
```diff
@@ -24,64 +24,91 @@
 # -----------------------------------------------------------------------
 
 
-# An interval `from..through` that includes `from`, `from+step`, `from+step+step`,
-# etc. up to and including `through`.
+# A `from..through` interval with a given `step`.
 #
-# In case step is positive (negative), the first value larger (smaller) than
-# through will not be included.
+# Generates a sequence by repeatedly adding `step` to `from`.
```
michaellilltokiwa: I would probably not speak of Sequence here, interval inherits from `Set` which inherits `Sequence` but still...

maxteufel:
Title of this pull request should be "interval: …" rather than "base: …" (more precise). Also this seems to incorporate the changes from #7600 and depends on that. That makes it a bit more tedious to review since I need to filter out all the changes I've already reviewed.

modules/base/src/interval.fz
```diff
@@ -101,7 +105,10 @@ is
 
   # get the number of elements in this interval
   #
-  # Performance O(1) in case  through.exists, otherwise O(result)
+  # Performance O(1).
+  #
+  # NOTE: diverges if `through` is `nil` —
+  # use `size_option`, which returns `nil` for unbounded intervals.
```
maxteufel: i would not use the word "diverges" here, rather say "enters an infinite loop"

modules/base/src/interval.fz
```diff
@@ -101,7 +105,10 @@ is
...
   #
   public size T
   =>
```
maxteufel, line 117: maybe add a precondition for finiteness

modules/base/src/interval.fz
```diff
@@ -153,6 +162,72 @@ is
...
+    # NYI: for `T` wider than `i64` the result may overflow.
+    through.exists ? size.as_i64 : nil
+
+
+  # index of `e` in this interval, `nil` if not contained
+  #
+  # Inverse of `nth`: for a valid index `i`, `index_of` of the `i`-th
+  # element returns `i`.
+  #
+  # Performance: O(1), unlike the inherited O(n) search.
```
maxteufel: it is critical here that `contains` is also O(1), which is only a given due to redefinition in this file. maybe add a comment.

modules/base/src/interval.fz
```diff
@@ -26,15 +26,19 @@
 
 # A `from..through` interval with a given `step`.
 #
-# Generates a sequence by repeatedly adding `step` to `from`.
+# Generates an interval by repeatedly adding `step` to `from`.
```
fridis: an interval is also a `Sequence T`, so I would leave this in or write something like

```suggestion
# This is a `Set T` and a `Sequence T` created by repeatedly adding `step` to `from`.
```

modules/base/src/interval.fz
```diff
@@ -104,9 +108,14 @@ is
 
   # get the number of elements in this interval
   #
-  # Performance O(1) in case  through.exists, otherwise O(result)
+  # Complexity: O(1).
+  #
+  # NOTE: enters an infinite loop if `through` is `nil` —
+  # use `size_option`, which returns `nil` for unbounded intervals.
```
fridis: This comment is not needed. The precondition requires that that `through` is not `nil`, we do not need to describe what happens if the caller does not satisfy the precondition.

modules/base/src/interval.fz
```diff
@@ -117,11 +126,13 @@ is
         else
           (thru - from) / step + T.one
       nil =>
+        # unreachable under the precondition
         T.from_u32 as_list.count.as_u32
```
fridis: Since the preconditions ensures the `through` exists, we can panic if it does not:

```suggestion
    thru := through.or_panic
    if from = thru then
      1
    else if step.sign = (from ⋄ thru).sign then
      0
    else
      (thru - from) / step + 1
```

modules/base/src/interval.fz
```diff
@@ -156,9 +168,77 @@ is
...
+    # NYI: for `T` wider than `i64` the result may overflow.
+    through.exists ? size.as_i64 : nil
+
+
+  # index of `e` in this interval, `nil` if not contained
+  #
+  # Inverse of `nth`: for a valid index `i`, `index_of` of the `i`-th
+  # element returns `i`.
+  #
+  # Complexity: O(1), unlike the inherited O(n) search.This depends on
```
fridis:
```suggestion
  # Complexity: O(1), unlike the inherited O(n) search. This depends on
```

modules/base/src/interval.fz
```diff
@@ -169,16 +249,16 @@ is
       l T =>
         interval l first -step
       nil =>
-        interval T.one T.zero T.one # empty interval
+        interval T.one T.zero T.one # any empty interval; from/step are irrelevant
```
fridis: Does this work?
```suggestion
        interval T 1 0 1  # any empty interval; from/step are irrelevant
```

modules/base/src/interval.fz
```diff
@@ -169,16 +249,16 @@ is
...
   # does this interval contain the given value?
   #
-  # `e` must lie within the bounds and be reachable from `from` by whole
+  # `e` must lie within the bounds and be reachable from `from` by whole steps
   #
   public redef contains(e T) bool =>
-    if      step.sign > 0 then from <= e <= through.or_else e && (e %  step = from %  step)
-    else if step.sign < 0 then from >= e >= through.or_else e && (e % -step = from % -step)
+    if      step.sign > 0 then from <= e <= through.or_else e && (e - from) %  step = T.zero
+    else if step.sign < 0 then from >= e >= through.or_else e && (e - from) % -step = T.zero
```
fridis: Does this work?
```suggestion
    if      step.sign > 0 then from <= e <= through.or_else e && (e - from) %  step = 0
    else if step.sign < 0 then from >= e >= through.or_else e && (e - from) % -step = 0
```

modules/base/src/interval.fz
```diff
@@ -202,4 +282,4 @@ is
   # count diverge.
   #
   public redef finite trit =>
-    if through.exists then trit.yes else trit.no
+    through.exists ? trit.yes : trit.no
```
fridis:
```suggestion
    through.exists.as_trit
```

tests/interval/test_interval.fz
```diff
@@ -27,35 +27,64 @@
 test_interval is
 
   test(x interval i64) =>
-    say "$x$(" "*(10 - ($x).codepoint_count))$(
+    inv := (0..3) ∀ (n -> match x.nth n
+                            e i64 => (x.index_of e ? i i64 => i = n
+                                                   | nil   => false)
```
fridis:
```suggestion
                            e i64 => x.index_of e >>? =n
```

tests/interval/test_interval.fz
```diff
@@ -27,35 +27,64 @@
...
+  test_eq (1..5)           (interval 1 5 2)
+  test_eq (-3..3:2)        (-3..4:2)
+  test_eq (1..10)          (1..10)
+  test_eq (1..10)          (10..1:(-1))
+
+  say "interval  |  contains -4..4            |size | frs lst | index_of -3..3        | nth 0..3    | inv  | reverse"
+  say "----------+----------------------------+-----+---------+-----------------------+-------------+------+---------"
   test -3..3
-  test -3..(-3)
+  test -10..(10):3
```
fridis: Maybe keep the original test as well:
```suggestion
  test -3..(-3)
  test -10..(10):3
```

tests/interval/test_interval.fz
```diff
@@ -27,35 +27,64 @@
...
   test 0..0
-  test 3..3
+  test 3..10
```
fridis, line 87: Maybe keep the original test as well:

```suggestion
  test 3..3
  test 3..10
```

fridis (changes requested):
Nice change. 

Lots of minor, picky comments from me with some minor code improvements and suggestions.

modules/base/src/interval.fz
arturdmi:
```
build/modules/base/src/interval.fz:246:9: error 1: Incompatible types in assignment
        interval 1 0 1 # any empty interval; from/step are irrelevant

assignment to field : 'interval.reverse_interval.result'
expected formal type: 'interval T'
actual type found   : 'interval i64'
assignable to       : 'interval i64'
for value assigned  : 'interval 1 0 1'
```

fridis: Use `interval T 1 0 1`

arturdmi: It also throws an error.
```
build/modules/base/src/interval.fz:246:22: error 1: Incompatible types when passing argument in a call
        interval T 1 0 1

Actual type for argument #2 'through' does not match expected type.
In call to          : 'interval'
expected formal type: 'option T'
actual type found   : 'i64'
assignable to       : 'i64'
for value assigned  : '0'
```

fridis (approved):
good.

PR #7677, opened 2026-09-02
add editorconfig

michaellilltokiwa:
[skip ci]

Seems to be quite common these days

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,7 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
+tab_width = 8
 
 [Makefile]
 indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# https://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


maxteufel (approved):
I find this good, and I'm glad someone else suggests it too now. I believe I tried to suggest this two years but was questioned because it added stuff to the repository root.

fridis (approved):
maybe we should add `tab_width = 8`?

fridis:
> I find this good, and I'm glad someone else suggests it too now. I believe I tried to suggest this two years but was questioned because it added stuff to the repository root.

I think Michael suggested this back in 2021 and never got a reply from me. My bad...

.editorconfig
```diff
@@ -0,0 +1,16 @@
+# https://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 2
```
michaellilltokiwa, line 10:
```suggestion
indent_size = 2
tab_width = 8
```

maxteufel:
I believe we could also set a soft line length limit.

fridis:
> I believe we could also set a soft line length limit.

Not sure how this is defined. I see 79 as the limit for header comments, the rest of the code should stay within 120 but there must be no hard limit for special cases.

PR #7678, opened 2026-09-02
add github issue templates

michaellilltokiwa:
[ci skip]

@tokiwa-software/developers Do you think these are helpful?


maxteufel:
I already find it slightly annoying with the pull request template to always remove that, so I don't really like this. People who contribute to Fuzion are likely familiar with Git(Hub) and bug reporting in the first place so I don't think this will help a lot.

simonvonhackewitz:
I think we don't need it at the moment, as we are not getting lots of bug reports lacking important information from external people. ~~But if we add it I think we should have an option for a blank issue.~~ I should have read it more attentively 🙈

PR #7690, opened 2026-09-02
base: switch, implement generic versions of `type.equality`, `type.lteq`, `type.hash_code`

michaellilltokiwa:
For this to work I added a workaround to AbstractType.compareTo that cotypes parametric types and corresponding parametric type are considered the same feature.


maxteufel (approved):
neat

PR #7693, opened 2026-09-02
ast: warn for unused fields that get placed in internal features by syntax sugar 

### LINKED ISSUE #7590: unused field in loop body not reported — causing confusing code in case of field shadowing
```
ex =>
  done := false
  while !done
    say "loop iter"
    done := true

  say done
```
results in an infinite loop, as it should. However to people unfamiliar with the language, this might be surprising. Can we detect this dangerous shadowing and throw a warning?

simonvonhackewitz:
So, for a feature that is not internal, i.e., one written directly in the code, it should not matter whether syntax sugar places it in an internal feature when checking for its usage.

Seems a bit to simple though...

fix #7590


michaellilltokiwa (approved):
The rule `!this.outer().featureName().isInternal()` did not make much sense.

PR #7571, opened 2026-08-19
change default integer type from `i32` to `i64`

michaellilltokiwa:
fixes #880

PR #7694, opened 2026-09-03
be/interpreter: support racy accesses for `i64` and `u64`

### LINKED ISSUE #7688: git bisect to find bad commit for issue in interpreter, test/atomic currently broken
Interpreter is currently broken for tests/atomic
https://github.com/tokiwa-software/fuzion/actions/runs/33577102269/job/100083373313

A git bisect is needed to find the offending commit, to test run:
```
  make no-java && make int -C ./build/tests/atomic
```

arturdmi:
[fixes #7688]
Since #7571 changed the default integer type, test 1 2 3 4 in
test_atomic.fz is now concur.atomic i64 instead of i32


maxteufel:
are we certain that the JVM implements atomicity for both these integer types?

michaellilltokiwa:
We should just use `clazzIsBuiltInPrimitive` instead of this large or. You can change the code in `Intrinsix.java` as well.

arturdmi:
Agreed on replacing the or-chain. One issue though:` clazzIsBuiltInPrimitive is isScalar || bool`, and `isScalar` includes i64, u64 and f64. In the interpreter those are stored as two separate int slots (Instance lines 202-203), so a racy read can observe a half-written value. 

Same problem for the jvm backend, which is what asked above: the value is a plain field, `ACC_VOLATILE` is defined in `ClassFileConstants` but never used, and [JLS §17.7](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html) does not guarantee atomicity for non-volatile long/double.

So should the check be `clazzIsRef || (clazzIsBuiltInPrimitive && size <= 4)`, or do you want to make the jvm backend emit ACC_VOLATILE for the atomic value field and change the interpreter representation? The second is correct but a much bigger change.

arturdmi:
`racy_accesses_supported` lists c_i64/c_u64 but not c_f64. Should c_f64 be added as part of this, or is it excluded for another reason?

fridis:
These operations are not atomic, the reason is code, i.e., in `src/dev/flang/be/interpreter/u64Value.java`:
```
  /**
   * Store this value in a field
   *
   * @param slot the slot that addresses the field this should be stored in.
   *
   * @param size the size of the data to be stored
   */
  void storeNonRef(LValue slot, int size)
  {
    if (PRECONDITIONS) require
      (size == 2);

    slot.container.nonrefs[slot.offset    ] = (int) _val;
    slot.container.nonrefs[slot.offset + 1] = (int) (_val >> 32);
  }
```
Here, the value is split up in two halves that are written independently. This is the case for all 64-bit values. 

So this cannot be fixed easily.

michaellilltokiwa (commented):
`*.expected_err_int` is missing. Can you check this in as well?

src/dev/flang/be/interpreter/Intrinsics.java
```diff
@@ -235,9 +235,11 @@ public static Callable call(Executor executor, int site, int innerClazz)
              (t == executor.fuir().clazz(SpecialClazzes.c_i8  )) ||
              (t == executor.fuir().clazz(SpecialClazzes.c_i16 )) ||
              (t == executor.fuir().clazz(SpecialClazzes.c_i32 )) ||
+             (t == executor.fuir().clazz(SpecialClazzes.c_i64 )) ||
              (t == executor.fuir().clazz(SpecialClazzes.c_u8  )) ||
              (t == executor.fuir().clazz(SpecialClazzes.c_u16 )) ||
              (t == executor.fuir().clazz(SpecialClazzes.c_u32 )) ||
+             (t == executor.fuir().clazz(SpecialClazzes.c_u64 )) ||
              (t == executor.fuir().clazz(SpecialClazzes.c_f32 )) ||
```
michaellilltokiwa, line 243:
```suggestion
(t == executor.fuir().clazz(SpecialClazzes.c_u8  )) ||
(t == executor.fuir().clazz(SpecialClazzes.c_u16 )) ||
(t == executor.fuir().clazz(SpecialClazzes.c_u32 )) ||
(t == executor.fuir().clazz(SpecialClazzes.c_f32 )) ||
```

maxteufel:
Closing this because it's not possible as per @fridis comment. Please open a new pull request for whatever rerecording needs to be done.

Issue #3670, opened 2024-08-29
Warn if atomic field is created on value type, or should atomic be `ref`?


michaellilltokiwa:
```
some_feat is
  x := atomic i32 3
```
If we use `some_feat` and pass it around, copies are made of `x`, which I think is a `racy_read` and also probably not what is intended.

fridis:
Maybe instead of forbidding atomics in a value type, we could have a `property.not_cloneable` (or `not_clonable`?), let `atomic` inherit from this property and flag an error if a value that is `not_cloneable` is passed as a value?

fridis:
Another possibility that might be easier is to make `atomic` a `ref`. I find this a little unsatisfactory since it adds a level of indirection in many cases where this is not needed.

fridis:
Moved this discussion to #7703 that addresses the Fuzion memory model with respect to instance initialization and cloning in a more general way.

PR #7698, opened 2026-09-03
base: remove atomic field in value effect pipe

### LINKED ISSUE #7561: pipe.used field probably does not work as intended
```
private:public pipe : effect is
...
  # has this pipe been used already?
  #
  public used concur.atomic bool := concur.atomic bool .new false
```

Since pipe is a value, pipe will get copied around and with it the atomic field.

michaellilltokiwa:
fixes #7561

see also #3670

### DIFF
--- a/modules/base/src/os/pipe.fz
+++ b/modules/base/src/os/pipe.fz
@@ -37,21 +37,13 @@ private:public pipe : effect is
   status := fzE_pipe_create pipefd.internal_array.data
 
 
-  # has this pipe been used already?
-  #
-  public used concur.atomic bool := concur.atomic bool .new false
-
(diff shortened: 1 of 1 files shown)


arturdmi (approved):

simonvonhackewitz (approved):

simonvonhackewitz:
Is it necessary to ensure in some other way that the pipe is used only once?

fridis (approved):
I have added a request to #7561 to add an example.

maxteufel:
I think the proper course of action here would have been to make pipe a ref effect? It is quite important we do not use a previously used pipe again.

PR #7700, opened 2026-09-03
parser: fix parsing of effect annotations in feature signatures

### LINKED ISSUE #7510: inheritance and `!` for effects conflict when defining a constructor
```
build/modules/base/src/concur/task_pool.fz:30:24: error 1: Syntax error: expected 'is', or '=>' in routine declaration, found operator '!'
) : realtime.Scheduler ! M

While parsing: implRout, parse stack: implRout, routOrField, feature, expr, exprs, block (twice), unit
```

simonvonhackewitz:
Prevent `effects()` from treating the `:` in an inherits call as a free type and consuming it.

Mostly implemented by AI using OpenCode with Big Pickle, with some manual intervention required.


maxteufel (approved):

michaellilltokiwa (changes requested):
I think this needs to be adjusted as well:
```
  boolean skipEffects()
  {
    return skip('!') && skipSimpleTypeList();
  }
```

src/dev/flang/parser/Parser.java
```diff
@@ -1549,7 +1569,7 @@ boolean skipTypeList()
    * @return true iff the next token(s) form typeList, otherwise no typeList was
    * found and the parser/lexer is at an undefined position.
    */
-  boolean skipSimpleTypeList()
+  boolean skipOneTypeList()
```
maxteufel, line 1572: i believe `skipOneTypeList` also needs to `skipOnetype`.

maxteufel (changes requested):

src/dev/flang/parser/Parser.java
simonvonhackewitz: That does not work. For code like this
```
lm ! ()->
  res := (container.mutable_tree_map lm K B).empty
```
it falsely produces syntax errors
```
build/modules/base/src/Sequence.fz:1222:11: error 1: Syntax error: expected 'is', or '=>' in routine declaration, found operator ':='
      res := (container.mutable_tree_map lm K B).empty
----------^
While parsing: implRout, parse stack: implRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), implRout, routOrField, feature, expr, exprs, block (twice), unit
```

I think @michaellilltokiwa said that there are some inconsistencies between the skip and the normal methods, but that should probably be fixed in a separate PR.

maxteufel: then at least add an NYI comment here

simonvonhackewitz: > That does not work.

It actually works when setting `isFunctionReturnType` to `true` in `skipOneType(true, true)`

maxteufel (approved):

michaellilltokiwa (approved):

PR #7701, opened 2026-09-03
base: cleanup in `switch` features

michaellilltokiwa:
remove `postfix ??` and removed `exists` in favour of `ok`

### DIFF
--- a/modules/lock_free/src/lock_free/Map.fz
+++ b/modules/lock_free/src/lock_free/Map.fz
@@ -506,7 +506,7 @@ is
   # Main_Node -> container_node -> Indirection_Node -> dual x y
   dual(x, y singleton_node CTK CTV, lev u32, gen uid) Main_Node CTK CTV =>
     d choice (container_node CTK CTV) (tomb_node CTK CTV) (list_node CTK CTV) =>
-      # NYI: UNDER DEVELOPMENT: why 35??
+      # NYI: UNDER DEVELOPMENT: why 35.ok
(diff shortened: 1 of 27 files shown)


modules/lock_free/src/lock_free/Map.fz
```diff
@@ -506,7 +506,7 @@ is
   # Main_Node -> container_node -> Indirection_Node -> dual x y
   dual(x, y singleton_node CTK CTV, lev u32, gen uid) Main_Node CTK CTV =>
     d choice (container_node CTK CTV) (tomb_node CTK CTV) (list_node CTK CTV) =>
-      # NYI: UNDER DEVELOPMENT: why 35??
+      # NYI: UNDER DEVELOPMENT: why 35.ok
```
simonvonhackewitz: I think this is just a question
```suggestion
      # NYI: UNDER DEVELOPMENT: why 35?
```

simonvonhackewitz (approved):

fridis (commented):
I kind of like the `postfix ??`, but I would like this to be `postfix ?`, which the parser currently does not support yet.

Then we could use `o?` to test and `o!` to check/panic (which IMHO should be used more often than `.or_panic` since it is more convenient). I also like the correspondence with the `infix >>?`, that should maybe be moved from `option` to `switch` and allows nice things like `str.parse_i32 >>? <=10`.

simonvonhackewitz (approved):

PR #7711, opened 2026-09-07
base: Mutable_Map, implement default `as_map`

michaellilltokiwa:


modules/base/src/container/Mutable_Map.fz
```diff
@@ -111,9 +111,13 @@ public Mutable_Map(K type : property.equatable, V type) ref is
...
   # create an immutable copy of this mutable map
   #
-  # NYI: maybe add default implementation when there is a naive implementation only requiring property.equatable
-  #
-  public as_map container.Map K V => abstract
+  public as_map container.Map K V =>
+    if K : property.orderable
+      container.ps_map K V .new keys values
+    else if K : property.hashable
+      container.hash_map keys values
```
simonvonhackewitz, line 118: A bit inconsistent that one requires `.new` while the other one doesn't.

Should we make the API more uniform?

simonvonhackewitz (approved):

modules/base/src/container/Mutable_Map.fz
michaellilltokiwa: @simonvonhackewitz Yes, I'm working on this currently.

Issue #7712, opened 2026-09-07
Redefining a function with precondition as argument field results in fz crash


fridis:
Take this example 
```
a is
  ok => true
  val i32
  pre
    ok
  =>
    -1

b(redef val i32) : a is
say <| b 42 .val

c : a is
  redef val i32 := 67

say <| c.val
```
it results in a crash
```
 > ./build/bin/fz -modules=base test_redef.fz

error 1: java.lang.NullPointerException: Cannot invoke "dev.flang.util.List.size()" because the return value of "dev.flang.ast.Feature.arguments()" is null
	at dev.flang.ast.Feature.<init>(Feature.java:808)
	at dev.flang.ast.Feature.<init>(Feature.java:752)
	at dev.flang.ast.Contract$2.<init>(Contract.java:646)
	at dev.flang.ast.Contract.addPreFeature(Contract.java:639)
	at dev.flang.ast.Feature.internalResolveTypes(Feature.java:1585)
	at dev.flang.ast.Resolution.resolveTypes(Resolution.java:531)
	at dev.flang.ast.Feature.resultTypeIfPresentUrgent(Feature.java:2487)
[..]

*** fatal errors encountered, stopping.
one error.
```
After removing the code related to `b`, the crash is gone, but we run into the restriction of #3092:

PR #7713, opened 2026-09-07
base: enhancesments of `container.Doubly_Linked_List`

fridis:
The original code was basically good, but I found some points that I think would benefit from being improved:

- sentinel no longer contains a value, so value no longer needs to be an option and `.or_panic` can be avoided.

- `next`/`prev` now contain `L_Node`, not `option L_Node`, so we do not need all the `.or_panic`calls. Instead of `nil`, `L_Node.this` is used.

- new feature `V_Node` inherits from `L_Node` and adds a value

- `val` renamed as `payload` since `val` is a field in `mutate.new` such that automatic unwrapping does not work with this name.

- removed all calls to `.or_panic`, they are no longer needed.

- fixed postcondition of `is_empty` since `!a^b` equals `a=b`

- removed repeated checking in loops of the form

      for
      until a || b then
        if a then
          x
        else
          y

  using

      for
      while !a
      until b then
        y
      else
        x

- in `as_string`, use `as_list.as_string`

- renamed `count` as `length`. The idea is that `count` may actually need to count the elements, hence `Sequence.count` is in `O(count)`, while `length` is in `O(1)`.

Feature `payload` could be an argument field if #7712 was fixed, for now `NYI` comments mention this.

### DIFF
--- a/modules/base/src/container/Doubly_Linked_List.fz
+++ b/modules/base/src/container/Doubly_Linked_List.fz
@@ -245,7 +245,7 @@ private:public Doubly_Linked_List(
     LM.env.exclusive ()->
       for e := sentinel.next, e.next
       while !e.is_sentinel
-      until id T e.payload = elem then
+      until e.payload = elem then
         true
       else
         false
--- a/modules/base/src/container/Doubly_Linked_List.fz
+++ b/modules/base/src/container/Doubly_Linked_List.fz
@@ -75,7 +75,8 @@ V_Node(# mutate effect to be used manipulate pointers and stored element
        # NYI: BUG: #7712: Should be possible to `redef payload` right here
        #
        payload0 T
-       ) : L_Node LM T
+       ) ref
+       : L_Node LM T
 is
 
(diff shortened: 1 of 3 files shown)


simonvonhackewitz (dismissed):

modules/base/src/container/Doubly_Linked_List.fz
```diff
@@ -206,12 +242,12 @@ private:public Doubly_Linked_List(
...
     LM.env.exclusive ()->
-      for e := sentinel.next.or_panic, e.next.or_panic
-      until e.is_sentinel || e.val = elem
-        !e.is_sentinel
-
-
-
+      for e := sentinel.next, e.next
+      while !e.is_sentinel
+      until id T e.payload = elem then
```
fridis:
```suggestion
      until e.payload = elem then
```

fridis: applied.

simonvonhackewitz (approved):

PR #7714, opened 2026-09-07
base: cleanup Map creation

michaellilltokiwa:
Main points:
- all map constructors private
- maps must implement this: `public type.of(kvs Sequence (tuple K V)) Map.this => abstract`
- removed public type.empty Map.this => abstract
- removed `container.map_of`


maxteufel (commented):
This API looks much nicer than the `map_of` API. Neat! I fail to see though where the different ordering comes from. Can you explain that?

michaellilltokiwa:
@maxteufel This is a difference in `as_string`, I removed `as_string` from `ordered_map` which behaves differently from `Map.as_string`

maxteufel (approved):

PR #7717, opened 2026-09-07
fe: add workaround in `AbstractType.compareTo`

michaellilltokiwa:
...to consider cotype typepar and origin typepar to be the same feature

needed for #7690

### DIFF
--- a/modules/base/src/container/Mutable_Map.fz
+++ b/modules/base/src/container/Mutable_Map.fz
@@ -132,9 +132,5 @@ public Mutable_Map(K type : property.equatable, V type) ref is
   #
   public type.from_entries(kvs Sequence (tuple K V)) Mutable_Map.this =>
     m := empty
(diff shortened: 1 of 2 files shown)


maxteufel:
for what exactly from #7690 do you need this? do you have a small example?

michaellilltokiwa:
@maxteufel 
E.g. here the subject type choices are resolved to `switch.A`/`switch.B` while the case types are resolved to `switch.type.A`/`switch.type.B`. This would lead to incompatible type errors.
```

  public redef type.equality(x, y switch.this) bool
  =>
    if A : property.equatable
      if B : property.equatable
        match x
          ax A =>
            match y
              ay A => ax = ay
              B    => false
          bx B =>
            match y
              A    => false
              by B => bx = by
      else
        compile_time_panic
    else
      compile_time_panic
``

maxteufel (approved):

PR #7718, opened 2026-09-07
base: implement orderable, hashable for `nil`

michaellilltokiwa:
This is necessary for #7690

### DIFF
--- a/modules/base/src/nil.fz
+++ b/modules/base/src/nil.fz
@@ -29,9 +29,23 @@
 # not present, e.g, in an option.
 #
 #
-public nil is
+public nil : property.hashable, property.orderable is
 
 
   # convert this to a string.
   #
   public redef as_string String => "--nil--"
+
+
+  # trivial implementation for lteq of unit value nil
+  #
+  public redef fixed type.lteq(_, _ nil) bool
+  =>
+    true
+
+
+  # create hash code for nil
+  #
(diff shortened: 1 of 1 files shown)


modules/base/src/nil.fz
```diff
@@ -29,9 +29,23 @@
...
+  public redef fixed type.lteq(_, _ nil) bool
+  =>
+    true
+
+
+  # create hash code for nil
+  #
+  public redef fixed type.hash_code(_ nil) u64
+  =>
+    0xfe6f6476106a56b6  # generated using secure random generator at https://cryptotools.dev/
```
maxteufel, line 51: what is the purpose of this random value compared to just using `0`?

michaellilltokiwa: ~~@maxteufel Good question, I pondered this as well. I think there is not much benefit apart from if there is a compiler bug we may be able to find it faster with distinct numbers.~~


~~Wait what if we have `option String` and `option.hash_code` just uses the the hash_code of `A` and `B`. Then it should better be a random number like it is here.~~

But that would be a very bad implementation for `option.hash_code`

maxteufel: `option.hash_code` needs to be able to handle same hash code from two different types in any case.

maxteufel (approved):

PR #7719, opened 2026-09-07
base: remove NYI comment

michaellilltokiwa:
I don't think it is sensible to "search for . and : in parallel"


maxteufel:
why not?

michaellilltokiwa:
@maxteufel Because the overhead of doing this in parallel (starting threads then syncing the results again) very likely exceeds the benefits

maxteufel:
This is not meant here; what I mean is combining the `.contains "."` and `.contains ":"` steps into one single pass through the string.

PR #7724, opened 2026-09-08
be/int be/jvm: fix boxing

### LINKED ISSUE #7720: JVM/interpreter: subtle difference in boxing behaviour compared to C backend
Take this samml example creating a list `FAIL <-> OK -> OK`
```
s := FAIL
e := OK
e.p <- s
e.n <- s
s.p <- e
s.n <- e
s.p.n <- OK
say s.n.n

FAIL ref is
  p := mut FAIL FAIL.this
  n := mut FAIL FAIL.this

OK : FAIL is
```
It works for the interpreter and jvm backends, but fails for the C backend:
```
 > ./build/bin/fz -interpreter test_box.fz 
instance[OK]
 > ./build/bin/fz -jvm test_box.fz 
instance[OK]
 > ./build/bin/fz -c test_box.fz && ./universe 
instance[FAIL]
```
Making `OK` a ref
```
s := FAIL
e := OK
e.p <- s
e.n <- s
s.p <- e
s.n <- e
s.p.n <- OK
say s.n.n

FAIL ref is
  p := mut FAIL FAIL.this
  n := mut FAIL FAIL.this

OK ref : FAIL is
```
fixes this:
```
 > ./build/bin/fz -interpreter test_box2.fz 
instance[OK]
 > ./build/bin/fz -jvm test_box2.fz 
instance[OK]
 > ./build/bin/fz -c test_box2.fz && ./universe 
instance[OK]
```
The most likely reason is that the two boxed values for `e` created by ` s.p <- e` and `s.n <- e` are the same instance of JVM/interpreter backend, while the C backend correctly creates two distinct instances (which is actually correct here!)

Here is a small example with OK/FAIL the right way around: 
```
e := FAIL
e.n <- OK
s OK := e      # box `e`, so s.n is OK
e.n <- FAIL    # but e.n is FAIL
say s.n        # should print OK

OK ref is
  n := mut OK OK.this

FAIL : OK is
```
we get this:
```
 > ./build/bin/fz -interpreter test_box6.fz 
instance[FAIL]
 > ./build/bin/fz -jvm test_box6.fz 
instance[FAIL]
 > ./build/bin/fz -c test_box6.fz && ./universe 
instance[OK]
```
and here is an example not using `mutate`: 
```
e := v
s R := e      # box `e`, so s.m is "OK"
e.f           # set e.m to "FAIL"
say s.m       # s.m should still be "OK"

R ref is
  m := "OK"
  f => set m := "FAIL"

v : R is
```
This must be run using `-XenableSetKeyword`:
```
 > ./build/bin/fz -XenableSetKeyword -jvm test_box7.fz 
FAIL
 > ./build/bin/fz -XenableSetKeyword -c test_box7.fz && ./universe 
OK
```

michaellilltokiwa:
fixes #7720

### DIFF
--- a/tests/reg_issue7720/reg_issue7720.fz
+++ b/tests/reg_issue7720/reg_issue7720.fz
@@ -1,0 +1,37 @@
+# This file is part of the Fuzion language implementation.
+#
+# The Fuzion language implementation is free software: you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as published
+# by the Free Software Foundation, version 3 of the License.
+#
+# The Fuzion language implementation is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License along with The
+# Fuzion language implementation.  If not, see <https://www.gnu.org/licenses/>.
+
+
+# -----------------------------------------------------------------------
+#
+#  Tokiwa Software GmbH, Germany
+#
+#  Source code of Fuzion test reg_issue7720
+#
+# -----------------------------------------------------------------------
+
+s := FAIL
+e := OK
+e.p <- s
+e.n <- s
+s.p <- e
+s.n <- e
+s.p.n <- OK
+say s.n.n
+
+FAIL ref is
+  p := mut FAIL FAIL.this
+  n := mut FAIL FAIL.this
+
+OK : FAIL is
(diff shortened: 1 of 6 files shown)


maxteufel (approved):

fridis (approved):

tests/reg_issue7720/reg_issue7720.fz
```diff
@@ -0,0 +1,37 @@
...
+s.p <- e
+s.n <- e
+s.p.n <- OK
+say s.n.n
+
+FAIL ref is
+  p := mut FAIL FAIL.this
+  n := mut FAIL FAIL.this
+
+OK : FAIL is
```
fridis, line 37: would IMHO be good to add the other two examples from the issue here as well.

PR #7726, opened 2026-09-08
mail: workaround for call to compile_time_panic via switch

maxteufel:
Previously, switch required only the first type parameter to be equatable, this recently changed and therefore now this call to equals fails. As a workaround, we only compare if there is a String.


michaellilltokiwa (approved):
Good but not a workaround IMHO ;-)

PR #7737, opened 2026-09-09
ir: fix boxing if ref to be boxed to is in nested choice

michaellilltokiwa:
We have the case box, tag, tag again, the PR where this problem shows up #7714

the value type: `container.ps_map String nom.parsers.this.json_value`
the type we want to assign to is: `outcome nom.parsers.this.json_value`

the json value is a choice defined as:

`json_value : choice String f64 bool nil (Sequence json_value) (container.Map String json_value)`

so what needs to happen is:

box ps_map to become a Map, than tag this to become a json_value, then tag it to become an outcome.

For the boxing case needsBoxing wrongly returned `json_value` instead of `Map`.

### DIFF
--- a/src/dev/flang/ast/Expr.java
+++ b/src/dev/flang/ast/Expr.java
@@ -845,21 +845,40 @@ public abstract class Expr extends ANY implements HasSourcePosition
         if (frmlT.isChoice() &&
             frmlT.isAssignableFromWithoutBoxing(t).no() &&
             frmlT.isAssignableFrom(t).yes())
-          { // we do both, box and then tag:
-            for (var cg : frmlT.choiceGenerics())
-              {
-                if (cg.isAssignableFrom(t).yes())
-                  {
-                    return cg;
-                  }
-              }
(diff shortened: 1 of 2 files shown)


src/dev/flang/ast/Expr.java
```diff
@@ -863,6 +857,31 @@ else if (frmlT.isRef() || frmlT.isParametricType() || frmlT.isThisType() && !frm
...
+   * @param t
+   * @return
+   */
+  private AbstractType findAssignableChoiceGeneric(AbstractType ct, AbstractType t)
+  {
+    for (var cg : ct.choiceGenerics())
+      {
+        if (cg.isChoice() && cg.isAssignableFrom(t).yes())
+          {
+            return findAssignableChoiceGeneric(cg, t);
```
maxteufel, line 874: i assume this recursion could lead to problems if a recursive choice type ever slipped in here. but that seems unlikely.

maxteufel (approved):

src/dev/flang/ast/Expr.java
michaellilltokiwa: @maxteufel Good point! But I think these should be caught in frontend already and boxing happens only later.

PR #7740, opened 2026-09-09
base: better implementation for `Set.type.equality`

michaellilltokiwa:


modules/base/src/container/Set.fz
```diff
@@ -181,8 +181,11 @@ is
   #
   public fixed redef type.equality(a, b Set.this) bool
   =>
-    # NYI: a bit expensive
-    (a ∪ b).size_option.or_panic = a.size_option.or_panic
+    if !a.finite.is_yes || !b.finite.is_yes
+      panic "Set.equality called on Set not known to be finite"
```
simonvonhackewitz, line 185: Could there be infinite sets where comparison is still possible? Or unknown sets that turn out to be finite?

maxteufel: yes but only in more specific implementations (more information is necessary). provided only the information we have here, we cannot safely assume this without running the risk of this never terminating.

maxteufel (approved):
I think this is as efficient as we can get here. Nice.

PR #7742, opened 2026-09-09
java code, fix formatting of `if` expressions

michaellilltokiwa:


src/dev/flang/lsp/shared/SourcePositionTool.java
```diff
@@ -79,7 +79,7 @@ public static SourcePosition byLineColumn(SourceFile sf, int line, int column)
   public static SourcePosition byLine(SourceFile sf, int line)
   {
     // lineStartPos throws in case of empty file
-    if(line == 1)
+    if (line == 1)
     {
```
simonvonhackewitz, line 83: this needs to be indented

src/dev/flang/lsp/shared/MarkdownTool.java
```diff
@@ -45,7 +45,7 @@ private static String escape(String str)
 
   public static String italic(String str)
   {
-    if(str.isBlank()){
+    if (str.isBlank()){
```
simonvonhackewitz, line 48: opening brace should be in new line and indented

src/dev/flang/lsp/shared/SourcePositionTool.java
michaellilltokiwa: I don't want to change this in this PR since this is wrong in many more places in this file.

src/dev/flang/lsp/shared/MarkdownTool.java
michaellilltokiwa: I don't want to change this in this PR since this is wrong in many more places in the LSP code base.

maxteufel (approved):

simonvonhackewitz (approved):

PR #7752, opened 2026-09-10
base: support `infix` ∈ on Sequence instead of just Set

simonvonhackewitz:


maxteufel:
We talked about this in the office; @fridis and I have reservations about this due to the mathematical origin of the notation specifically for sets, as well as the O(n) nature of `contains` on Sequences.

PR #7753, opened 2026-09-10
base/f64/float/data_size: add `f64.as_u128` and remove NYI hack in `units.data_size.scale_approx`

arturdmi:
Removes `units.data_size.scale_approx` carried a NYI hack:
```
# NYI: HACK: `f64.as_u128` missing, result therefore limited to 7.99 EiB because of `.as_i64`
data_size (bytes.as_f64 * f).as_i64.as_u128
```
Before:
```
fz -e 'say ((units.data_size.eib 8).bytes, ((units.data_size.eib 8).scale_approx 2.0).bytes)'
fz -e 'say (((units.data_size.eib 8).scale_approx 1000.0).bytes)'

error 1: FATAL FAULT `precondition`: safety: fits_in_i64
```
After:
```
fz -e 'say ((units.data_size.eib 8).bytes, ((units.data_size.eib 8).scale_approx 2.0).bytes)'
# (2**63 and 2**64)
(9223372036854775808, 18446744073709551616)

fz -e 'say (((units.data_size.eib 8).scale_approx 1000.0).bytes)'
# 9223372036854775808000 
9223372036854775808000
```
**Changes**

- `float` — declare `fits_in_u128` / `as_u128`, 
   same shape as the existing `as_i64` / `fits_in_i64`
- `f64` — implement via `mantissa` / `exponent`, which already normalize both
  normals and subnormals.
- `f32` — delegate to `f64`, lossless widening
- `scale_approx` — use `as_u128`, guard on the feature itself, drop the hack


modules/base/src/f64.fz
```diff
@@ -67,6 +67,26 @@ public f64(public val f64) : float, java_primitive is
...
+  #
+  public redef as_u128 u128 =>
+    if val < 1.0
+      u128.zero
+    else
+      shift := exponent - f64.mantissa_bits
+      if shift ≥ 0
+        mantissa.as_u128 << shift.as_u128
+      else
+        mantissa.as_u128 >> (-shift).as_u128
```
maxteufel: use `shift.abs` here?

maxteufel (changes requested):

maxteufel (approved):

michaellilltokiwa (changes requested):
@arturdmi Can you please add a few tests, specifically of the corner cases that show that this works as intended?

arturdmi:
Calculations from the test (with `bc` tool)

image

maxteufel:
> Calculations from the test (with bc tool)

@arturdmi what is this for?

arturdmi:
> > Calculations from the test (with bc tool)
> 
> @arturdmi what is this for?

to demonstrate the correctness of calculation (see expected_out of a test)

maxteufel (approved):

tests/f64_as_u128/f64_as_u128.fz
```diff
@@ -0,0 +1,73 @@
...
+#
+# -----------------------------------------------------------------------
+
+f64_as_u128 =>
+
+  # 2^n exactly, built from integers (no float literal rounding)
+  p2(n i64) u128 => u128.one << n.as_u128
+
+  # format: "ok <bc-expr> <as_u128>"  or  "no <bc-expr>"
+  chk(e String, x f64) unit =>
```
michaellilltokiwa: `chk` and `chk32` should be deduplicated:

```suggestion
  chk(F type : float, e String, x F) unit =>
```

michaellilltokiwa: Also I don't understand the comment: 

> format: "ok <bc-expr> <as_u128>"  or  "no <bc-expr>"

michaellilltokiwa:
I'm thinking maybe `as_u128` should take an argument `rounding_mode` that specifies how rounding should be done?

fridis:
> I'm thinking maybe `as_u128` should take an argument `rounding_mode` that specifies how rounding should be done?

What use case do you have for this? I would expect that documenting the rounding mode, my preference is to return to `floor` in case there is a fractional part, should be sufficient.

michaellilltokiwa (approved):

Issue #4964, opened 2025-03-22
reference point of indentation in anonymous features


michaellilltokiwa:
currently this is allowed:

https://github.com/tokiwa-software/fuzion/blob/42b6b4acaff2a54b8f2889cf5e715f052e58613d/tests/reg_issue293_confusion_of_type_and_value_params/issue293.fz#L34-L38

Would be better if the implementation of anonymous feature had to be more indented than the ref keyword.

PR #7756, opened 2026-09-11
fe/SourceModule: added two conditions to the filter

### LINKED ISSUE #7754: To-solve part in error `Feature declared using modifier 'redef' does not redefine another feature` wrong
Take this code 
```
a is
  f (a, b, c i32) =>

b : a is
  redef f(x, y i32) =>
```
that produces a justified error:
```
 > FUZION_DISABLE_ANSI_ESCAPES=true ./build/bin/fz test_redef.fz

/home/fridi/fuzion/fuzion/test_redef.fz:5:9: error 1: Feature declared using modifier 'redef' does not redefine another feature
  redef f(x, y i32) =>
--------^
Redefining feature: 'b.f'
To solve this, check spelling and argument count against the feature you want to redefine or remove 'redef' modifier in the declaration of 'b.f'.
Also make sure that the feature to be redefined is visible where it is redefined. There is the feature 'b.f' that could be made public to allow redefinition in 'b'.

one error.
```
Only problem is the last sentence: `There is the feature 'b.f' that could be made public to allow redefinition in 'b'.` This does not make sense since `b.f` is the redefining feature. This hint should only list `a.f` if that was not visible, but not here since it is visible to `b`.

arturdmi:
[fixes #7754 ]
There were 2 problems:
1. The feature itself was included in the list.
2. Visibility was not checked.
So two conditions fixes it.

I also added new test for 
#7754 
and updated exp_err for 
#4964.

### DIFF
--- a/src/dev/flang/fe/SourceModule.java
+++ b/src/dev/flang/fe/SourceModule.java
@@ -957,7 +957,7 @@ A feature that does not redefine an inherited feature must not use the `redef` m
             List<FeatureAndOuter> hiddenFeaturesSameSignature = lookup(outer, f.baseName(), null, true, true)
               .stream()
               .filter(fo -> fo._feature != f // excluding the feature itself
-                          && !visibleFor(fo._feature, f) //only truly invisible features
+                          && !visibleFor(fo._feature, f) // only truly invisible features
                           && fo._feature.featureName().equals(f.featureName()))
               .collect(List.collector());
             AstErrors.redefineModifierDoesNotRedefine(f, hiddenFeaturesSameSignature);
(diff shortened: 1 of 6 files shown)


src/dev/flang/fe/SourceModule.java
```diff
@@ -955,7 +955,11 @@ else if (c._hasPost != null && c._hasPostThen == null)
     // end::fuzion_rule_PARS_NO_REDEF[]
             */
             List<FeatureAndOuter> hiddenFeaturesSameSignature = lookup(outer, f.baseName(), null, true, true)
-              .stream().filter(fo->fo._feature.featureName().equals(f.featureName())).collect(List.collector());
+              .stream()
+              .filter(fo -> fo._feature != f // excluding the feature itself
+                          && !visibleFor(fo._feature, f) //only truly invisible features
```
maxteufel:
```suggestion
                          && !visibleFor(fo._feature, f) // only truly invisible features
```
space after comment

maxteufel (approved):

michaellilltokiwa (approved):
Nice!

PR #7761, opened 2026-09-11
fe: try remove explicit `resolveTypesOfActuals` in `loadCalledFeatureUnlessTargetVoid`

michaellilltokiwa:
We should not do  `resolveTypesOfActuals` in `loadCalledFeature`  unless there is a good reason, `resolveTypesOfActuals` should be done only in `resolveTypes` phase.


maxteufel (approved):

fridis:
Very nice! I am surprised that this does not cause any problems, but apparently it is fine!

PR #7779, opened 2026-09-16
tests: add test for timeout in record_process

### LINKED ISSUE #7379: Timeout of `tokiwa.record_process` does not work in some cases
In some cases `tokiwa.record_process` does seem to never return after timeout has been reached. When the timeout is reached the process should be killed. Maybe not all child processes get killed or the reading of out/err blocks even though the process gets killed.

Because of this the timeout of 10 minutes for individual flang_dev examples in the Jenkins pipeline does currently not work, resulting in the whole pipeline getting stuck when a single example does not terminate.

michaellilltokiwa:
I think this might be meanwhile fixed, hence added some test that might trigger the problem.

fixes #7379


maxteufel (approved):

tests/reg_issue7379/reg_issue7379.fz
```diff
@@ -0,0 +1,183 @@
...
+            e := ferr.get
+            p.close_out.or_panic
+            p.close_err.or_panic
+            process_result $o $e ec)
+      .or_panic
+
+
+reg_issue7379 =>
+  for res in ["cat".execute (container.ps_map String String).empty nil (time.duration.s 3) .or_panic,
+              "cat".execute (container.ps_map String String).empty (io.path.of "reg_issue7379.fz.stdin") (time.duration.s 3) .or_panic] do
```
simonvonhackewitz, line 180: you could just use `sleep 10m` here and avoid copying the code from tokiwa module. 

```suggestion
  for res in ["sleep 10m".execute (container.ps_map String String).empty nil (time.duration.s 3) .or_panic,
              "sleep 10m".execute (container.ps_map String String).empty (io.path.of "reg_issue7379.fz.stdin") (time.duration.s 3) .or_panic] do
```

simonvonhackewitz (commented):
I think this already worked in the past. The problem was that it is not working with the testrunner in flandev.

tests/reg_issue7379/reg_issue7379.fz
maxteufel: how does this reduce code copied?

simonvonhackewitz: Everything above `reg_issue7379 =>` is code copied from tokiwa module, because this modification is needed
```
# do not close in
# close_in.or_panic
```
in order for `cat` to not return.

PR #7781, opened 2026-09-16
base: use total order defined by IEEE for f64 with exception of -0.0=0.0

simonvonhackewitz:


modules/base/src/f64.fz
```diff
@@ -310,31 +310,36 @@ public f64(public val f64) : float, java_primitive is
...
+    b_ := if b.is_zero
+            0x8000000000000000
+          else if b.is_sign_bit_set
+            # flip all bits
+            ~b.cast_to_u64
+          else
+            # flip the first bit
+            b.cast_to_u64 ^ 0x8000000000000000
+
+    a_ <= b_
```
maxteufel, line 341: This is too inefficient for such a basic operation. We ideally have code here that in the best path is a single CPU instruction. You need to check here what the C standard says and what the JVM does, and we need them to match.

Apparently the C default floating point comparison is ordered as by IEEE 754 <https://stackoverflow.com/questions/37305326/unordered-floating-point-comparisons-in-c>, so just removing our `a.is_NaN` might be enough for the C side. You will also need to check how the JVM handles these edge cases and make sure they are the same (again hopefully such that there is as little as possible custom code).

maxteufel (changes requested):

modules/base/src/f64.fz
```diff
@@ -310,31 +310,36 @@ public f64(public val f64) : float, java_primitive is
...
+    # NYI: OPTIMIZATION: avoid `if` to make this branchless
+    a_ := if a.is_zero
+            0x8000000000000000
+          else if a.is_sign_bit_set
+            # flip all bits
+            ~a.cast_to_u64
+          else
+            # flip the first bit
+            a.cast_to_u64 ^ 0x8000000000000000
+
```
fridis, line 331: Better say highest bit, first is the lowest on little endian machines.

This does not flip, but it sets the sign bit since you have just tested `a.is_sign_bit_set` , so better do


```suggestion
            # set the highest (sign) bit
            a.cast_to_u64 | 0x8000000000000000
```

modules/base/src/f64.fz
```diff
@@ -310,31 +310,36 @@ public f64(public val f64) : float, java_primitive is
...
+    b_ := if b.is_zero
+            0x8000000000000000
+          else if b.is_sign_bit_set
+            # flip all bits
+            ~b.cast_to_u64
+          else
+            # flip the first bit
+            b.cast_to_u64 ^ 0x8000000000000000
```
fridis, line 339: dto.

modules/base/src/f64.fz
fridis: I think this code should go into the `f64.type.equal`/`lower_than_or_equal` intrinsics and be implemented the most efficient way for each backend. 

Typically, non-branching code is better, i.e., instead of
```
  if b & 0x8000000000000000 != 0 then 
    ~b
  else 
    b | 0x8000000000000000
```
One would do something like
```
  b ^ 0x8000000000000000 ^ ((b & 0x8000000000000000)-1)
```
But these kinds of transformations are very tricky...

fridis (commented):
I am not that worried about the performance right now, but the right place for this logic is in the backend intrinsic implementation.

PR #7785, opened 2026-09-16
base: add `memoize.keep` to memoize the result of nullary functions

### LINKED ISSUE #7527: `once` as part of `memoize`?
`once` is currently used as follows
```
get_value := once ()->
  say "calculating value!"
  42

say get_value.unwrap
say get_value.unwrap
```
which works:
```
 > ./build/bin/fz test_once0.fz 
calculating value!
42
42
```
but which requires the use of a field, it does not work for a function `get_value => once ...`, and it requires explicit calls to `unwrap`. 

We can achieve the same using `memoize`, but with a bit of boilerplate using `unit` as the type of the keys.
```
get_value : memoize => keep unit _->
  say "calculating value!"
  42

say get_value
say get_value
```
this works for functions and does not require `.unwrap`:
```
 > ./build/bin/fz test_once.fz 
calculating value!
42
42
```
we could add `once` as an inner feature of `memoize` as follows (using `einmal` to avoid name clash, should eventually be `once`):
```
memoize.einmal(T type, F type : ()->T, f F) =>
  keep unit (_-> f())

get_value : memoize => einmal ()->
  say "calculating value!"
  42

say get_value
say get_value
```
same result:
```
 > ./build/bin/fz test_once1.fz 
calculating value!
42
42
```
What do you think? Inheriting from `memoize` is something one must learn, but once that is understood, it might be nicer.

fridis:
If you have a nullary function that calculates a value that does not change, e.g.,

     pi
     =>
       calculate_pi

you can use `memoize` as follows to avoid repeated calls to `calculate_pi`:

     pi : memoize
     =>
       const ()->
         calculate_pi

Applied this to `codepoint.type.latin_alphabet`.

As for most effects, the behavior for multi-threaded use is still quirky.  The effect will be stored thread-locally but transferred to newly spawned threads, but not to threads spawned earlier.

Removed keyword `const` for this. First thought about using `once` instead of `const`, but this name is already taken another feature. 

Added test cases for this to `tests/lib_memoize`.

This implements and fixes #7527.


maxteufel:
please fix the `lib_memoize` test

arturdmi (approved):

maxteufel (approved):

modules/base/src/memoize.fz
```diff
@@ -183,7 +220,7 @@ is
 
 
   # use this memoized effect to retrieve the value for a given key or,
-  # if that value was not calculated yet, to calculate it
+  # if that value was not calculated yet, do calculate it
```
michaellilltokiwa:
```suggestion
  # if that value was not calculated yet, calculate it
```

maxteufel:
effects for ryu also needs a rerecording, it seems

PR #7793, opened 2026-09-17
ci/release: build src and binaries for linux, macOS and windows on amd64/arm64

### LINKED ISSUE #5797: release action should be changed to emit src and linux/macOS/windows/bsd/other? binaries

arturdmi:
Fixes #5797

I found no `github runners` for BSD so it only creates binaries for macOS, windows and linux. 
The structure is the same like in 'packaging.yml' in 'feeze'. 

Outside of workflow changes (needed for windows/arm64):
1. `Makefile`: target triple for `fuzion_rt.dll` from `clang -dumpmachine` instead of hardcoding
2. `bin/windows_install_boehm_gc.sh`: use `$MINGW_PREFIX`/`$MINGW_CHOST` instead of `/ucrt64`


.github/workflows/release.yml
```diff
@@ -34,37 +40,178 @@ env:
...
+        with:
+          pattern: packages-*
+          merge-multiple: true
+          path: artifacts
+
+      - name: Create checksums
+        run: |
+          cd artifacts
+          sha256sum ./*.tar.gz > SHA256SUMS
+          cat SHA256SUMS
```
maxteufel, line 204: i think you do not need to redownload this, you can just store the checksum as a fact after generating the file and then assemble a checksums file later

redownloading involves potential of someone replacing the file and this is exactly what we want to avoid for checksums

maxteufel (changes requested):

michaellilltokiwa:
Other than the comment of @maxteufel this looks good to me. In a second step we will need some basic test that ensures that the package created actually can run a _Hello World_ or similar.

arturdmi:
@michaellilltokiwa 
When this pull request is merged, i will add this basic test in new PR

maxteufel (approved):
nice

PR #7794, opened 2026-09-17
base: remane `memoized_*effect` as `memoized_*`

fridis:
A little less bloat in the effects listing..

waiting for #7785 to be merged first...


fridis:
closing this for now, need to think of how to fix the naming conflict `memoize.memoized` and `memoized_effect` (that I tried to rename as `memoized`, resulting in a conflict.).

PR #7797, opened 2026-09-18
base: from_u32 tail recursive as suggested by @fridis

### LINKED ISSUE #7771: `type.from_u32` not TCO
current implementation is not tail call optimised:
```
  public type.from_u32(v u32) numeric.this =>
    if      v = 0 then zero
    else if v = 1 then one
    else
      h := from_u32 v/2
      l := from_u32 v%2
      h+h+l
```
This is not necessarily a problem since we see that it will only ever need a max of ~32 stackframes.
But if we wanted to forbid (self-)recursive features that are not tail call optimisable we may want a different implementation, e.g. like this:
```
  public type.from_u32(v u32) numeric.this =>
    for accum := zero, accum + accum + (if x%2 = 1 then one else zero)
        x in (v :: (/2)).take_while (>0) .reverse
    else
      accum
```

michaellilltokiwa:
fixes #7771


tests/reg_issue118/issue118.fz.expected_err
```diff
@@ -1,21 +1,15 @@
 
---CURDIR--/issue118.fz:55:3: error 1: Used abstract features 'numeric.type.zero', 'numeric.type.one' are not implemented by 'issue118.a.type'
+--CURDIR--/issue118.fz:55:3: error 1: Used abstract feature 'numeric.type.zero' is not implemented by 'issue118.a.type'
```
simonvonhackewitz, line 2: Why does this change?

michaellilltokiwa: I think this is due to the nature the DFA works.
Before it was:
`if something then zero else one`  => The DFA considers boths paths.
now it is:
`from_u32_rec v zero one` => The DFA evalutes the args in order but `zero` is an abstract that is not implemented, hence this results in  `void` for the DFA making it stop evaluating any further.

simonvonhackewitz (approved):

PR #7801, opened 2026-09-18
fe: uncomment NYI where name of loaded module is checked that now mig…

michaellilltokiwa:
…ht work


src/dev/flang/fe/FrontEnd.java
```diff
@@ -273,12 +273,11 @@ private LibraryModule module(String m, Path p, AbstractFeature universe)
       {
         var data = ch.map(FileChannel.MapMode.READ_ONLY, 0, ch.size());
         result = libModule(data, universe);
-        // NYI: BUG: does not work anymore
-        // if (!m.equals(result.name()))
-        //   {
-        //     Errors.error("Module name mismatch for module file '" + p + "' expected name '" +
-        //                  m + "' but found '" + result.name() + "'");
-        //   }
+        if (!m.equals(result.name()) && !"main".equals(result.name()))
```
fridis, line 276: This way, any module can just return `main` as its name and it will be loaded happily. I think we should maybe do this the other way around, e.g., 

```suggestion
        if (!m.equals(result.name()) && 
            !("whatever m is for main module".equals(m) && "main".equals(result.name())))
```

PR #7803, opened 2026-09-18
base: move Function.of to Any.of, rename other `type.of` as `type.from`

michaellilltokiwa:


modules/base/src/concur/Sending_Channel.fz
```diff
@@ -40,10 +40,15 @@ public Sending_Channel(T type) ref is
     abstract
 
 
+  # close this Channel which has the
```
simonvonhackewitz: comment seems incomplete

modules/base/src/concur/Sending_Channel.fz
```diff
@@ -40,10 +40,15 @@ public Sending_Channel(T type) ref is
...
+  #
   public close unit
   post
     debug: is_closed
   =>
     abstract
 
+
+  #
```
simonvonhackewitz: comment is missing

modules/base/src/Any.fz
```diff
@@ -93,3 +93,16 @@ public Any ref is
...
+
+
+  # alternative to using `id` to force an
+  # expression into a desired type
+  #
+  # i.e. `option i32 .of 42` instead of `id (option i32) 42`
+  #
+  # usage example:
+  #
+  #     Unary u64 i32 .of (.as_u64)
```
simonvonhackewitz: "usage example" is a bit redundant with "i.e.", and `Unary` could be a bit confusing here

modules/base/src/option.fz
```diff
@@ -198,4 +198,5 @@ public option(T type) : switch T nil is
 #
 #     o option TypeOfX := x
 #
+# NYI: CLEANUP: remove, use type.of instead
```
simonvonhackewitz, line 201: maybe create an issue for this?

modules/base/src/outcome.fz
```diff
@@ -120,4 +120,5 @@ public outcome(T type) : switch T error is
 #
 #     o outcome TypeOfX := x
 #
+# NYI: CLEANUP: remove, use type.of instead
```
simonvonhackewitz, line 123: maybe create an issue for this?

modules/base/src/option.fz
fridis, line 201: This is here because it works well with type inference: I often use 
```
x := condition ? option bla : nil
```
I do not want to write 
```
x := condition ? (option ((outer_feature TYPE_PAR1 TYPE_PAR2).inner_feeature TYPE PAR 3)).of bla : nil
```

modules/base/src/outcome.fz
fridis, line 123: dito.

fridis (commented):
I am not so sure about whether I like this. `Function.type.of x` was hardly ever used and can easily be replaced by `id Function x`, while all the other uses of `xyz.of` are IMHO very convenient and often nice to use. 

`from` is just two more letters, but why rename `of` when we have `id`?

michaellilltokiwa:
closed in favour of #7840

PR #7805, opened 2026-09-18
base: remove some is_* feature, replace by `=` or `!=`

michaellilltokiwa:
Inspired idea of @fridis in recent office discussion


michaellilltokiwa:
#7810 should be merged first

modules/base/src/numeric.fz
```diff
@@ -28,7 +28,6 @@
 public numeric : property.hashable, property.orderable is
 
   public is_zero bool => numeric.this = numeric.this.zero
-  public is_one  bool => numeric.this = numeric.this.one
```
simonvonhackewitz, line 31: What's the reason to remove this? And to keep `is_zero`?

modules/base/src/order.fz
```diff
@@ -51,50 +51,6 @@ public order : choice less greater equal, property.equatable is
       equal => 0
 
 
-  # the value of this order is `less`
-  #
-  public is_less bool =>
```
simonvonhackewitz, line 56: Maybe `order` should document that it can be compared using `=`?

Also, `.is_greater_than_or_equal` will turn into `!= less`.

simonvonhackewitz (approved):

modules/base/src/numeric.fz
michaellilltokiwa: I removed `is_zero` as well now.

modules/base/src/order.fz
michaellilltokiwa: Hmm, okay but do we want to document every `choice` that implements `property.equatable`?

fridis (commented):
This is ok for now, but I suggest not to remove `is_XYZ` in general from choice types. Think of something like this

```
  tree := container.red_and_black_tree data
  n := tree.get_node "abc"
  if n.color.is_red then 
    ...
```
vs. 
```
  if n.color = container.read_and_black_tree_choices.red then 
    ...
```

PR #7806, opened 2026-09-18
base: memoize globally, not only thread-locally

fridis:
The tricky part here is that I do not want to enter an exclusive section all the time just to read a memoized value.  So, I keep a thread-local copy that can be accessed without synchronization.

Only before the thread-local copy is set for the first time we need to access a global mutable variable (or map) exclusively once.

The uses a single global `blocking_mutate` for memoization of all values, so this could become a bottleneck.  But the exclusive sections are kept short, in particular, the calculation of the memoized value is not performed exclusively.


modules/base/src/memoize.fz
```diff
@@ -196,29 +221,40 @@ public memoize is
...
+
+                # the store used for memoization
+                #
+                # NYI: ENHANCEMENT: would be nice to also support property.hashable keys
+                # using a tree map here
+                 #
+                store container.ps_map K V,
+
+                # ref containing the mutatble global value, accessed while
+                # `mmemoize_mutate.env.exclusive`.
```
michaellilltokiwa:
```suggestion
                # `memoize_mutate.env.exclusive`.
```

modules/base/src/memoize.fz
```diff
@@ -231,37 +267,77 @@ is
...
-                                       V type,
+                                        # the stored value, accessed by current thread only, so no
+                                        # need for exclusivity
+                                        store option V,
 
-                                       # the stored value
-                                       store option V
-                                       )
+                                        # ref containing the mutatble global value, accessed while
+                                        # `mmemoize_mutate.env.exclusive`.
```
michaellilltokiwa:
```suggestion
                                        # `memoize_mutate.env.exclusive`.
```

modules/base/src/memoize.fz
```diff
@@ -196,29 +221,40 @@ public memoize is
...
-                              #
-                              K type : property.orderable,
-
-                              # the value type, i.e., result type of memoized function
-                              #
-                              V type)
+# This effect uses a global lock to ensure consistency. However, this global
+# lock is used at most once for each key/value pair used by each thread.  On
+# the first access to a key, a local copy of the global store is made such that
+# future accesses usign the same key will fin the value on the local store.
```
michaellilltokiwa:
```suggestion
# future accesses using the same key will find the value on the local store.
```

michaellilltokiwa (approved):

